PR #1, opened 2020-01-07
Add storybook configs and app tile component/story from ui-lib

ash3rz:
Note - The directory structure can be changed, I just picked something


sri2k1us:
So I haven't reviewed this PR but I have some thoughts about what we should look for in Sonora when we are doing reviews. I am gonna dump that info here. If everyone is on board, we can document this somewhere in gitlab.
- PR (UI code) should include 2 screen shots at the minimum. One for desktop, One for mobile

From Slack design systems blog (mostly applicable for ui-lib but might be applicable in Sonora as well). Some of this are little subjective but I want to at least think about these things in our code review
- **Robust:** Components should look great, representing the high standards we set for our product design.
- **Accessible**: Components should work for all of our users, regardless of how they navigate and interact with our software.
- **Flexible**: Components should support a variety of use cases within our product, while supporting future extensibility.
- **Reliable**: Components should be well-tested, dependable, and immediately responsive to user interaction.

- No explicit font size, width and height setting using `px` in `css`. I am not sure if we can do this 100% of the time but let try to do this. I think this will go a long way towards responsive design. Please see -
- https://material-ui.com/customization/breakpoints/ 
- https://material-ui.com/guides/responsive-ui/
- https://material-ui.com/system/sizing/

- Use MUI Typography component for all the text. This will enforce font size consistency across the app. For example a table column name, we can always use variant `h6`. Also see https://material-ui.com/system/typography/ on how we can use Typography for text alignments and styling.
- JSDoc as much as possible
- Use Google's Roboto font - https://fonts.google.com/specimen/Roboto
- Storybook and tests

ash3rz:
@sri2k1us I agree with all of your bullet points, except I think that the document should live here with the code in a README. I started to put together a standards/convention doc in the original UI repo for our students to be able to follow, and I think doing the same thing here will benefit us.  And then if we have any external contributors, they'll have an easily findable document on what we expect in code and PRs too.

sri2k1us:
> @sri2k1us I agree with all of your bullet points, except I think that the document should live here with the code in a README. I started to put together a standards/convention doc in the original UI repo for our students to be able to follow, and I think doing the same thing here will benefit us. And then if we have any external contributors, they'll have an easily findable document on what we expect in code and PRs too.

Sounds like a plan!

ash3rz:
Thanks for the reviews!

PR #3, opened 2020-01-08
Add Apollo and JSDoc

johnworth:
Configures Apollo as Express middleware in src/server/index.js. This means you can now access the graphql query UI at `/graphql` in your browser.

Adds initial, unauthenticated support for querying a couple of fields provided by terrain. The query UI at `/graphql` should list the available schema and queries.

Adds JSDoc as a dev dependency and adds npm scripts for generating documentation.

Modifies the nodemon configuration to generate the documentation with every change.


johnworth:
Thanks for the "prettier" suggestion, missed the fact that it was actually an option. I've got my editor set up to run it on every save now, too.

PR #4, opened 2020-01-08
Add eslint with a config

johnworth:
The important bit here is the eslint config in package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,11 +33,9 @@
     "exec": "babel-node ./src/server/"
   },
   "eslintConfig": {
-    "eslintConfig": {
-      "extends": [
-        "react-app"
-      ]
-    }
+    "extends": [
+      "react-app"
+    ]
   },
   "dependencies": {
     "@cyverse-de/ui-lib": "^1.0.2",
--- a/package.json
+++ b/package.json
@@ -33,23 +33,12 @@
     "exec": "babel-node ./src/server/"
   },
   "eslintConfig": {
-    "extends": [
-      "eslint:recommended",
-      "plugin:react/recommended"
-    ],
-    "rules": {    
-      "react/react-in-jsx-scope": "off",
-      "semi" : ["warn", "always", { "omitLastInOneLineBlock" : true }]
-    },
-    "globals": {    
-      "React": "writable"
-    },
-    "parserOptions": { 
-      "ecmaVersion": 6,
-      "sourceType" : "module",
-      "ecmaFeatures": {
-        "jsx": true,
-        "modules": true
+    "eslintConfig": {
+      "extends": [
+        "react-app"
+      ],
+      "rules": {    
+        "semi" : ["warn", "always"]
       }
     }
   },
--- a/package.json
+++ b/package.json
@@ -32,6 +32,27 @@
     "ext": "js",
     "exec": "babel-node ./src/server/"
   },
+  "eslintConfig": {
+    "extends": [
+      "eslint:recommended",
+      "plugin:react/recommended"
+    ],
+    "rules": {    
+      "react/react-in-jsx-scope": "off",
+      "semi" : ["warn", "always", { "omitLastInOneLineBlock" : true }]
+    },
+    "globals": {    
+      "React": "writable"
+    },
+    "parserOptions": { 
+      "ecmaVersion": 6,
+      "sourceType" : "module",
+      "ecmaFeatures": {
+        "jsx": true,
+        "modules": true
+      }
+    }
+  },
   "dependencies": {
     "@cyverse-de/ui-lib": "^1.0.2",
     "@material-ui/core": "^4.8.3",
@@ -71,6 +92,8 @@
     "babel-loader": "^8.0.6",
     "babel-preset-react-app": "^9.1.0",
     "cross-env": "^6.0.3",
+    "eslint": "^6.8.0",
+    "eslint-plugin-react": "^7.17.0",
     "jest": "^24.9.0",
     "nodemon": "^2.0.2",
     "npm-run-all": "^4.1.5",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -32,6 +32,27 @@
     "ext": "js",
     "exec": "babel-node ./src/server/"
   },
+  "eslintConfig": {
+    "extends": [
+      "eslint:recommended",
+      "plugin:react/recommended"
+    ],
+    "rules": {    
+      "react/react-in-jsx-scope": "off",
```
johnworth: This is take from https://medium.com/@joelmasters/setting-up-eslint-for-nextjs-37163d4cabaa.

package.json
```diff
@@ -32,6 +32,27 @@
...
+      "semi" : ["warn", "always", { "omitLastInOneLineBlock" : true }]
+    },
+    "globals": {    
+      "React": "writable"
```
johnworth: Also taken from https://medium.com/@joelmasters/setting-up-eslint-for-nextjs-37163d4cabaa

package.json
```diff
@@ -32,6 +32,27 @@
     "ext": "js",
     "exec": "babel-node ./src/server/"
   },
+  "eslintConfig": {
+    "extends": [
```
johnworth: The extends block is taken from https://medium.com/@joelmasters/setting-up-eslint-for-nextjs-37163d4cabaa.

package.json
johnworth: Also found here https://medium.com/@joelmasters/setting-up-eslint-for-nextjs-37163d4cabaa

package.json
```diff
@@ -32,6 +32,27 @@
...
+      "eslint:recommended",
+      "plugin:react/recommended"
+    ],
+    "rules": {    
+      "react/react-in-jsx-scope": "off",
+      "semi" : ["warn", "always", { "omitLastInOneLineBlock" : true }]
```
johnworth: Turns on warnings (not errors) about missing semicolons. Also allows us to skip semicolons on single-line blocks.

package.json
johnworth: Found here https://medium.com/@joelmasters/setting-up-eslint-for-nextjs-37163d4cabaa

package.json
```diff
@@ -32,6 +32,27 @@
...
+    },
+    "globals": {    
+      "React": "writable"
+    },
+    "parserOptions": { 
+      "ecmaVersion": 6,
```
johnworth: Enabled es6 checks.

package.json
```diff
@@ -32,6 +32,27 @@
...
+      "sourceType" : "module",
```
johnworth: Gets eslint to stop barking about import statements.

package.json
psarando: I'm not sure I agree with this setting (or the `"globals": { "React": "writable"}` setting).

If we are only ever going to use Next.js, then this looks convenient, but if we ever want to switch to something else, or if we want to port our React components somewhere else, I think this could this break compatibility.

package.json
psarando: Not sure about this one (`react-app` also does not enable this).
Also, I'm not sure our prettier settings allow one-line `{ ... }` blocks like this? I think they will get formatted onto new lines anyway.

package.json
psarando: If we extend `react-app`, we automatically get `ecmaVersion: 2018` and these other `parserOptions`.

package.json
johnworth: I'll remove the omitLastInOneLineBlock thing.

We'll need to have a discussion about semicolon usage. I personally prefer to have them but I know that some people don't, which is why I made it a warning and not an error.

package.json
johnworth: Next.js doesn't require us to import React in pages (which we don't have many of yet), and ESLint will start barking about that unless we turn off the react-in-jsx-scope setting. If we migrate away from Next.js then we can remove it, but I think we're going to have annoyances otherwise.

I'll try removing the globals stuff and see what happens.

johnworth:
I've extended react-app (I was thinking that it would have to be a dep in the project, but doesn't appear that way). I went ahead and nuked most of the other options that were indicated by @psarando. I left in the semicolon warning, though, but removed the option that conflicted with the prettier settings.

package.json
johnworth: This has been removed now.

package.json
johnworth: Removed as per @psarando's suggestion.

package.json
johnworth: Fixed in the new version of the file.

package.json
johnworth: Also removed in the new version of the file.

psarando:
> I've extended react-app (I was thinking that it would have to be a dep in the project, but doesn't appear that way).

I think we still need to add these deps (from https://github.com/facebook/create-react-app/tree/master/packages/eslint-config-react-app#usage-outside-of-create-react-app):
```
npm install --save-dev eslint-config-react-app @typescript-eslint/eslint-plugin@2.x @typescript-eslint/parser@2.x babel-eslint@10.x eslint@6.x eslint-plugin-flowtype@3.x eslint-plugin-import@2.x eslint-plugin-jsx-a11y@6.x eslint-plugin-react@7.x eslint-plugin-react-hooks@1.x
```

johnworth:
Adding those in now, thanks for pointing that out.

package.json
psarando: 👌 I also prefer to always have semicolons, otherwise there's a risk of hard-to-trace errors.

I vote we enforce them with ESLint. Especially if our prettier settings will add them anyway (and that's also the default in `react-app` ESLint settings).

PR #5, opened 2020-01-08
Add app tile test with snapshot

ash3rz:
I followed the example from jest's docs: https://jestjs.io/docs/en/tutorial-react

Jest complains if you try to import an image though (beta symbol from app tile), so I followed this suggestion: https://stackoverflow.com/questions/46898638/importing-images-breaks-jest-test

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,5 +89,10 @@
     "hooks": {
       "pre-commit": "pretty-quick --staged"
     }
+  },
+  "jest": {
+    "moduleNameMapper": {
+      "\\.(jpg|png|svg)$": "<rootDir>/src/__mocks__/imageMock.js"
+    }
   }
 }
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -88,5 +89,10 @@
     "hooks": {
       "pre-commit": "pretty-quick --staged"
     }
+  },
+  "jest": {
+    "moduleNameMapper": {
+      "\\.(jpg|png|svg)$": "<rootDir>/src/__mocks__/imageMock.js"
```
psarando: 👍

src/__tests__/apps.js
```diff
@@ -0,0 +1,12 @@
...
+import renderer from "react-test-renderer";
+
+import { AppTiles } from "../stories/AppTile.stories";
+
+test("App Tile renders", () => {
+    const component = renderer.create(
+        <AppTiles />,
+    );
+    let tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
```
psarando: 👍

PR #7, opened 2020-01-09
Switch to configuration files (and a few other fixes)

johnworth:
The main purpose of this pull request is to switch over to reading the configuration from a file rather than entirely from environment variables. We're now using the `config` and `js-yaml` packages.

YAML was used for familiarity and consistency with other deployables in our stack. A new file is being used because it's an opportunity to trim down the complexity of the UI configuration dramatically.

Configuration files must be placed in the top-level `config` directory. This is a convention introduced by the `config` package.

Local development is supported by placing override files in the top-level `config/` directory. See https://github.com/lorenwest/node-config/wiki/Configuration-Files for more info.

In the server code, configuration values are available through the `configuration` module at `src/server/configuration.js`.

JSDoc comments have been added for the `configuration` module.

I also ran into a couple of NPM project configuration issues while developing this and went ahead and fixed them:
- nodemon was restarting the entire stack when a Next.js page was changed, preventing the browser from automatically refreshing.
- ESLint was complaining about the lack of a React import in page files, which is allowable with Next.js. I disabled the rule that triggered the error report.


package.json
```diff
@@ -29,17 +29,19 @@
...
     "ext": "js",
     "exec": "run-p dev-server build-docs"
   },
   "eslintConfig": {
     "extends": [
       "react-app"
-    ]
+    ],
+    "rules": {
+       "react/react-in-jsx-scope": "off"
```
psarando: I guess this is the standard in Next.js React files (I noticed the lack of React imports in their tutorials).
It still feels a bit weird to me, but I suppose I'll get used to it 😅

src/server/configuration.js
```diff
@@ -0,0 +1,46 @@
...
+export const terrainURL = config.get("terrain_url");
+
+/**
+ * Is set to true if the application is running in
+ * development mode. Checks to make sure the NODE_ENV environment variable
+ * is not set to 'production'.
+ *
+ * @type {boolean}
+ */
+export const isDevelopment = process.env.NODE_ENV !== "production";
```
psarando: 👍

src/server/graphql/dataSources/TerrainDataSource.js
```diff
@@ -1,10 +1,10 @@
-import { RestDataSource, RESTDataSource } from "apollo-datasource-rest";
-import _ from "lodash";
+import { RESTDataSource } from "apollo-datasource-rest";
+import { terrainURL } from "../../configuration";
```
psarando: ✨

src/server/configuration.js
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * The port the application listens on for requests. Takes its value from the
+ * 'listen_port' setting in the configuration file.
+ * @type {number}
+ */
+export const listenPort = parseInt(config.get("listen_port"));
```
psarando: I would have thought that eslint would complain about not passing the 2nd `radix` param to `parseInt` here:
```suggestion
export const listenPort = parseInt(config.get("listen_port"), 10);
```
Maybe eslint isn't running on server-side code in dev mode?

package.json
johnworth: Feels weird to me, too.

johnworth:
The ESLint thing is a little weird, I'll look into that after the merge.

PR #9, opened 2020-01-10
WIP: initial authentication implementation

slr71:
I'm not satisfied with the way this is working yet. Login works correctly, but logging out doesn't yet. Also, the service ticket isn't stored in the user's session anywhere. I was hoping to use the session to authenticate to terrain, so we'll either need to add it to the user session or find an alternative to use.

Reviews are welcome at any time. If anyone has suggestions, please feel free to add them to this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,6 +106,9 @@ dist
 # Idea config
 *.iml
 
+# VSCode settings
+.vscode/
+
 #JSDoc output
 out/
 generated-docs/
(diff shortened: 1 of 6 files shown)


slr71:
It appears that `expressjs-cas-authentication` isn't going to work in its current state. We could probably modify it so that it will work, but I'd like to explore using Passport.js for a while instead. I'm going to close this PR for now.

PR #11, opened 2020-01-10
Add node.js workflow for CI

johnworth:
The file was created through Github Actions. Looks like it runs tests for us. Want to give it a shot?


johnworth:
And Jenkins started working. No need for this, now.

PR #16, opened 2020-01-14
Initial auth implementation using OAuth2

slr71:
I wanted to get something working quickly, so I'm going to go with OAuth2 for the time being. The passport strategy for OAuth2 seems to be better documented than the one for OpenID Connect. This isn't quite where I want it to end up yet, but I think it's at a point where it can be merged so that other people can start working on this.

The default configuration settings are almost sufficient for local testing, but you will need a `local.yaml` file for some additional configuration settings:

```yaml
terrain_url: https://example.org/terrain

db:
    host: database.example.org
    user: dbuser
    password: dbpassword

oauth2:
    client_id: oauthclientid
    client_secret: oauthclientsecret
```

Once you have these settings configured, you can start up your local server and open up a browser connection to it. To log in, navigate to `http://localhost:3000/login`. You should be redirected to the OAuth provider's login page. After entering your credentials you can navigate to `http://localhost:3000/graphql` to get to the GraphQL playground. While you're logged in, the authorization header should automatically be added to GraphQL requests. To log out, point your browser to `http://localhost:3000/logout`.

The next step is to add ways for users to be redirected to the `/login` and `/logout` routes when they click on the appropriate widgets in the UI.

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -2,9 +2,17 @@ base_url: http://localhost:3000
 terrain_url: http://terrain
 listen_port: 3000
 
+db:
+    host: localhost
+    port: 5432
+    user: null
+    password: null
+    database: de
+
 sessions:
     secret: "keyboard cat"
     secure_cookie: false
+    ttl: 8
 
 oauth2:
     authorization_url: "https://olson.cyverse.org/cas/oauth2.0/authorize"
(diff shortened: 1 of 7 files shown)


slr71:
Thanks for the review!

slr71:
Thanks for the review!

PR #17, opened 2020-01-16
Sanitize the filesystem data model and query.

johnworth:
## Scope
Move the folder listing return value into the Folder type and rename the top-level query.

## Changes

- Moves listing into the Folder type and renames it 'contents'.
- Removes the top-level listFolder query.
- Modifies the resolvers to handle the new layout.
- Modifies TerrainDataSource to remove stat information from the listFolder return value, which is now an iterable.

## Problems
Nesting the folder listings gets expensive quickly. 

## Sample query containing nested listings
The return value is too large to reproduce here.

```
{
  filesystem(paths: [
    "/iplant/home/ipcdev/"
  ]) {
    id
    name: label
    path
    type
    
    ... on File {
      fileSize
    }
    
    ... on Folder {
      contents {
        id
        name: label
        path
        type
        
        ... on File {
          fileSize
        }
        
        ... on Folder {
          contents {
            id
            name: label
            path
            type
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/src/server/graphql/dataSources/TerrainDataSource.js
+++ b/src/server/graphql/dataSources/TerrainDataSource.js
@@ -7,6 +7,7 @@
 import { RESTDataSource } from "apollo-datasource-rest";
 import { terrainURL } from "../../configuration";
 import _ from "lodash";
+import DataLoader from "dataloader";
 
 export const FILE = "file";
 export const FOLDER = "folder";
@@ -41,12 +42,16 @@ export default class TerrainDataSource extends RESTDataSource {
      * @param {string} fullPath - The full path to the file in the data store.
      * @returns {(File | Folder)}
      */
-    async filesystemStat(fullPaths) {
-        let retval = await this.post("/secured/filesystem/stat", {
+    fsStatDataLoader = new DataLoader(async (fullPaths) => {
+        const resp = await this.post("/secured/filesystem/stat", {
             paths: fullPaths,
         });
-        return _.values(retval.paths);
-    }
+
+        // Guarantees ordering, hopefully
+        return _.map(fullPaths, (path) => resp.paths[path]);
+    });
+
+    filesystemStat = async (path) => this.fsStatDataLoader.load(path);
 
     /**
      * Returns a folder listing.
--- a/src/server/graphql/dataSources/TerrainDataSource.js
+++ b/src/server/graphql/dataSources/TerrainDataSource.js
@@ -39,7 +39,9 @@ export default class TerrainDataSource extends RESTDataSource {
     /**
      * Returns info about a file or folder from terrain.
      * @param {string} fullPath - The full path to the file in the data store.
-     * @returns {(File | Folder)}
+     *
+     * Each object in the returned array is either a File or a Folder.
+     * @returns {Array<Object>}
      */
     fsStatDataLoader = new DataLoader(async (fullPaths) => {
         const resp = await this.post("/secured/filesystem/stat", {
@@ -60,7 +62,7 @@ export default class TerrainDataSource extends RESTDataSource {
      * @param {string} entityType - The type of items included in the results. Should be one of FILE, FOLDER, or ANY.
      * @param {string} sortColumn - The column to sort on. One of NAME, ID, LASTMODIFIED, DATECREATED, SIZE, or PATH.
      * @param {string} sortDirection - One of "ASC" (for ascending order) or "DESC" (for descending order).
-     * @returns {Array<{(File | Folder)}>}
+     * @returns {FolderContents}
      */
     async listFolder(
         path,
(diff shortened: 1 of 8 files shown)


src/server/graphql/resolvers/Filesystem.js
```diff
@@ -44,17 +44,30 @@ export default {
...
+    },
+
+    Folder: {
+        fileCount: async (folder, _args, { dataSources }) => {
+            const stat = await dataSources.terrain.filesystemStat(folder.path);
+            return stat["file-count"];
+        },
+        shareCount: async (folder, _args, { dataSources }) => {
+            const stat = await dataSources.terrain.filesystemStat(folder.path);
+            return stat["share-count"];
```
slr71: Does GraphQL manage caching the result of the call to the `stat` endpoint for us?

johnworth: The caching provided by Apollo doesn't provide batching as well, AFAIK. We're both batching requests and caching it with dataloader. Not sure if the in-process LRU cache would be okay in our environment because of the multiple replicas and I'd really rather not add Redis or memcached to the stack until we have multiple places that absolutely require one of them.

ianmcorvidae: Doing my thing and reading about it, I guess Dataloader caches within the context of a single request, but not cross-request. Still a nice improvement, probably, and it's really hard for us to provide nice cache semantics outside of a single request. The caching provided by Apollo might be more useful in non-filesystem contexts where we can provide more useful cache headers in Terrain if we want (of course, it'll also be less useful there, but anyway).

src/server/graphql/typesDefs/Filesystem.js
```diff
@@ -113,6 +118,11 @@ export default gql`
         This is an expensive field, especially when nested inside the contents.
         """
         fileCount: Int
+
+        """
+        This an expensive field to include inside content listings.
```
psarando: 👍 good comments

src/server/graphql/typesDefs/Filesystem.js
```diff
@@ -98,16 +105,25 @@ export default gql`
...
 
-    # Allows us to fetch everything at once without having to cache results or
-    # hit stat a bunch of times.
-    type FolderListing {
-        stat: Folder
-        listing: [FilesystemObject]
+        """
+        This an expensive field to include inside content listings.
+        """
+        folderCount: Int
```
psarando: I almost forgot: the file and folder counts are slow because they come from the stat endpoint, but the listing endpoint does return a `total` field (the combined count for the listing), and that will be required for paging listings in the UI.

johnworth:
folderCount and contentType are now available. 

I'll go ahead and convert over to ES6 syntax now, I'd rather standardize on that now and I don't want to forget about it later.

src/server/graphql/dataSources/TerrainDataSource.js
```diff
@@ -41,12 +42,16 @@ export default class TerrainDataSource extends RESTDataSource {
...
-        let retval = await this.post("/secured/filesystem/stat", {
+    fsStatDataLoader = new DataLoader(async (fullPaths) => {
+        const resp = await this.post("/secured/filesystem/stat", {
             paths: fullPaths,
         });
-        return _.values(retval.paths);
-    }
+
+        // Guarantees ordering, hopefully
+        return _.map(fullPaths, (path) => resp.paths[path]);
```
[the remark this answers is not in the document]
johnworth: Should be fixed now, thanks.

johnworth:
I added the `FileContents` type and changed `File.contents` to it. It contains `total`, `totalBad`, and `listing` fields. A client query to grab that information could look like this:

```
{
  filesystem(path: "/iplant/home/ipcdev/testing-vice") {
    ... on Folder {      
      contents {
        total
        totalBad
        
        listing {
          path
          type
        }
      }
    }
  }
}
```

src/server/graphql/dataSources/TerrainDataSource.js
```diff
@@ -77,34 +81,24 @@ export default class TerrainDataSource extends RESTDataSource {
...
+        });
 
-        return retval;
+        const listingFiles = files.map((f) => {
+            f.type = FILE;
+            return f;
+        });
+
+        return {
+            total,
```
psarando: Don't forget to change the `@returns` JSDoc for this function, too.

sri2k1us:
Yes! this `_` syntax is driving me nuts 😩.
> I'd also still like to explore getting rid of `lodash` in favor of ES6 syntax, but that can be saved for the next PR.

src/server/graphql/dataSources/TerrainDataSource.js
```diff
@@ -82,24 +81,20 @@ export default class TerrainDataSource extends RESTDataSource {
...
+
+        const listingFiles = files.map((f) => {
+            f.type = FILE;
+            return f;
+        });
 
-        return listing;
+        return [...listingFolders, ...listingFiles];
     }
 }
```
sri2k1us: 👍

PR #18, opened 2020-01-16
CORE-868: allow the user to click the "user" button in the navigation…

slr71:
… bar in order to log in

This is a small change to allow users who aren't logged in yet to log in by clicking the user button in the apps bar. It's an incremental step; we still have some details to flesh out. 😁


slr71:
Thanks for the review! (I was actually hoping to get this merged in, so the review is much appreciated.) 🙂

PR #19, opened 2020-01-21
No lodash

johnworth:
Removes our dependency on lodash. Use ES6 syntax where possible, otherwise add small dependencies on the `camelcase` and `is-plain-object` packages.


src/server/graphql/resolvers/index.js
```diff
@@ -1,13 +1,11 @@
-import { merge } from "lodash";
-
 import Query from "./Query";
 import BigInt from "graphql-bigint";
 import Filesystem from "./Filesystem";
 
-export default merge(
-    {
+export default {
+    ...{
```
psarando: This works, but `BigInt: BigInt` could be pulled up to the top-level instead of being nested in a spreaded-object.
Also, ES6 let's you omit the key if the value is stored in a var with the same name (think also of ES6 destructuring).
So this could also be written as:
```javascript
export default {
    BigInt,
    ...Filesystem,
    ...Query,
};
```

johnworth: Thanks! Yeah, that was carryover from docs...somewhere. I'll fix it.

PR #22, opened 2020-01-22
Create and use a default MUI theme

johnworth:
This adds a default theme that lives in `src/components/theme/default` that uses theme information from ui-lib.

The custom app in `src/pages/_app.js` has been wrapped in a `ThemeProvider` that sets the theme to the new one mentioned above.

A decorator has been added to the Storybook config that wraps the stories in a `ThemeProvider` that is passed the new default theme mentioned above.

Explicit references to the ui-lib palette have been removed from the existing components (`appBar`, and `apps/tile`) and replaced with references to the new default theme as returned by MUI's `useTheme()` or passed into `withStyles()`.

### DIFF
--- a/src/components/appBar/CyVerseAppBar.js
+++ b/src/components/appBar/CyVerseAppBar.js
@@ -44,7 +44,7 @@ const useStyles = makeStyles((theme) => ({
         flexGrow: 1,
     },
     appBar: {
-        backgroundColor: theme.palette.blue,
+        backgroundColor: theme.palette.primary,
     },
     title: {
         flexGrow: 1,
--- a/src/components/appBar/CyVerseAppBar.js
+++ b/src/components/appBar/CyVerseAppBar.js
@@ -172,7 +172,10 @@ function CyverseAppBar(props) {
                             >
                                 <AccountCircleIcon
                                     fontSize="large"
-                                    style={{ color: theme.palette.white }}
+                                    style={{
+                                        color:
+                                            theme.palette.primary.contrastText,
+                                    }}
                                 />
                             </IconButton>
                         </div>
--- a/src/components/appBar/CyVerseAppBar.js
+++ b/src/components/appBar/CyVerseAppBar.js
@@ -12,7 +12,6 @@ import constants from "../../constants";
 
 import {
     build,
-    palette,
     formatHTMLMessage,
     formatMessage,
     withI18N,
@@ -35,7 +34,7 @@ import {
     Toolbar,
     Typography,
 } from "@material-ui/core";
-import { fade, makeStyles } from "@material-ui/core/styles";
+import { fade, makeStyles, useTheme } from "@material-ui/core/styles";
 import SearchIcon from "@material-ui/icons/Search";
 import ArrowDropDownIcon from "@material-ui/icons/ArrowDropDown";
 import AccountCircleIcon from "@material-ui/icons/AccountCircle";
@@ -45,7 +44,7 @@ const useStyles = makeStyles((theme) => ({
         flexGrow: 1,
     },
     appBar: {
-        backgroundColor: palette.blue,
+        backgroundColor: theme.palette.blue,
     },
     title: {
         flexGrow: 1,
@@ -98,6 +97,7 @@ const useStyles = makeStyles((theme) => ({
 const searchOptions = ["All", "Data", "Apps", "Analyses"];
 
 function CyverseAppBar(props) {
+    const theme = useTheme();
     const classes = useStyles();
     const router = useRouter();
     const { intl, children } = props;
@@ -172,7 +172,7 @@ function CyverseAppBar(props) {
                             >
                                 <AccountCircleIcon
                                     fontSize="large"
-                                    style={{ color: palette.white }}
+                                    style={{ color: theme.palette.white }}
                                 />
                             </IconButton>
                         </div>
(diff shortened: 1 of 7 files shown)


src/components/theme/default/index.js
```diff
@@ -0,0 +1,15 @@
...
+        primary: {
+            main: palette.blue,
+        },
+        secondary: {
+            main: palette.lightBlue,
+        },
+    },
+});
+
+export default theme;
```
sri2k1us: We might want to define overrides all the intentions defined here - https://material-ui.com/customization/palette/

src/pages/_app.js
```diff
@@ -30,9 +32,11 @@ export default class MyApp extends App {
...
-            <CyverseAppBar>
-                <Component {...props} />
-            </CyverseAppBar>
+            <ThemeProvider theme={theme}>
+                <CyverseAppBar>
+                    <Component {...props} />
+                </CyverseAppBar>
+            </ThemeProvider>
         );
     }
```
sri2k1us: 👍

src/components/appBar/CyVerseAppBar.js
```diff
@@ -186,6 +186,7 @@ function CyverseAppBar(props) {
 
 function SearchOptions(props) {
     const { intl } = props;
+    const theme = useTheme();
```
sri2k1us: 👍

src/components/theme/default/index.js
sri2k1us: This seems very useful - https://material-ui.com/customization/color/#color-tool

johnworth: Agreed on both counts. In the UX meeting yesterday, Mariah mentioned getting us access to the updated set of official CyVerse color values. I was planning on one of us going through and adding all of them to the theme once we have them.

src/components/appBar/CyVerseAppBar.js
```diff
@@ -45,7 +44,7 @@ const useStyles = makeStyles((theme) => ({
         flexGrow: 1,
     },
     appBar: {
-        backgroundColor: palette.blue,
+        backgroundColor: theme.palette.blue,
```
psarando: I don't think this will work, because I don't think the theme's `palette` object has `blue` defined (or `darkBlue`, or `white`, etc.) by default.

Can you verify if the theme colors are displaying correctly in storybook? Or are there any errors in the console?

We could probably add these fields to the `createMuiTheme` call in the index file, but I think it's better to stick with the theme's intentions where possible.
So instead of `theme.palette.blue`, I think this should be `theme.palette.primary`.
And instead of `theme.palette.red`, try `theme.palette.error`.
See https://material-ui.com/customization/default-theme/?expand-path=$.palette for the default fields in the `palette`.

src/components/appBar/CyVerseAppBar.js
```diff
@@ -172,7 +172,7 @@ function CyverseAppBar(props) {
                             >
                                 <AccountCircleIcon
                                     fontSize="large"
-                                    style={{ color: palette.white }}
+                                    style={{ color: theme.palette.white }}
```
psarando: I'm not sure what to do about this color, but I wonder if `theme.palette.primary.contrastText` is meant for a situation like this. 🤔

src/components/appBar/CyVerseAppBar.js
johnworth: I made sure that there weren't any rendering or console errors in Storybook before I filed the PR. 

If I'm understanding things correctly, the ui-lib settings override or add to default MUI theme, so if something isn't overridden the MUI default will be used. I'll double/triple-check that that's the case, though.

src/components/appBar/CyVerseAppBar.js
ash3rz: Seems like the default for 'theme.palette.primary.contrastText` is white, but not sure if that gets recalculated once we modify `theme.palette.primary` with CyVerse colors.  It sounds like the right one to use anyway: https://material-ui.com/customization/default-theme/

src/components/appBar/CyVerseAppBar.js
psarando: I ask, because in the index file, I only see a reference to ui-lib's `palette.blue` and `palette.lightBlue` in the `createMuiTheme` call, so I'm not sure how the MUI theme would be able to access all the other colors from the `ui-lib` palette.

johnworth: Okay, I've changed:
- `theme.palette.blue` references to `theme.palette.primary`
- `theme.palette.red` references to `theme.palette.error`
- ui-lib's palette is now merged into the default palette

That last one should hopefully pick up any missing color references.

src/components/appBar/CyVerseAppBar.js
johnworth: Yup, it gets recalulated. It's showing up gray instead of white. Working on a fix now.

johnworth: Wait, no. I'm confusing the heck out of myself.

johnworth: Okay, I changed it to `theme.palette.primary.contrastText` and it's showing up as white now. Sorry for my dumbness.

johnworth:
Thanks for the reviews!

PR #23, opened 2020-01-24
 Navigation

sri2k1us:
- Added responsive navigation bar
- Created pages to match app navigation points
- Mapped root to "Dashboard"
- Note - I can update the colors and Icons to match Mariah's prototype in another PR. This needs to get in asap so that people have some way to navigate and don't have to deal with ugly style breaks.

image

image

### DIFF
--- a/src/components/navigation/Navigation.js
+++ b/src/components/navigation/Navigation.js
@@ -4,20 +4,25 @@
  * A component that displays app level navigation
  */
 import React from "react";
+import intlData from "./messages";
+import ids from "./ids";
+import { build, getMessage, withI18N } from "@cyverse-de/ui-lib";
 import CloudUploadIcon from "@material-ui/icons/CloudUpload";
 import AppsIcon from "@material-ui/icons/Apps";
 import DashboardIcon from "@material-ui/icons/Dashboard";
 import AssessmentIcon from "@material-ui/icons/Assessment";
 import MoreHorizIcon from "@material-ui/icons/MoreHoriz";
-import { Button, Grid, Hidden, NoSsr, Paper } from "@material-ui/core";
+import { Button, Grid, Hidden, Paper } from "@material-ui/core";
 import { useRouter } from "next/router";
 import { makeStyles, useTheme } from "@material-ui/core/styles";
+import { injectIntl } from "react-intl";
 
 const useStyles = makeStyles((theme) => ({
     button: {
         margin: theme.spacing(1),
         "&:hover": {
             backgroundColor: theme.palette.blue,
+            color: theme.palette.white,
         },
     },
 
@@ -25,10 +30,6 @@ const useStyles = makeStyles((theme) => ({
         margin: theme.spacing(1),
         backgroundColor: theme.palette.blue,
         color: theme.palette.white,
-    },
-
-    paper: {
-        margin: theme.spacing(1),
     },
 }));
 
@@ -42,12 +43,14 @@ function Navigation(props) {
     const classes = useStyles();
     const router = useRouter();
     const { navigate } = props;
+    const theme = useTheme();
     console.log("navigate->" + navigate);
     return (
-        <Paper elevation={10} className={classes.paper}>
+        <Paper elevation={10} style={{ backgroundColor: theme.palette.lightGray }}>
             <Grid container>
                 <Grid item xs>
                     <Button
+                        id={build(ids.NAVIGATION, ids.DASHBOARD_BTN)}
                         className={
                             navigate === DASHBOARD
                                 ? classes.selectedButton
@@ -57,7 +60,7 @@ function Navigation(props) {
                             router.push("/dashboard");
                         }}
                     >
-                        <DashboardIcon /> <Hidden xsDown>Dashboard</Hidden>
+                        <DashboardIcon/> <Hidden xsDown>{getMessage("dashboard")}</Hidden>
                     </Button>
                 </Grid>
                 <Grid item xs>
@@ -71,7 +74,7 @@ function Navigation(props) {
                             router.push("/data");
                         }}
                     >
-                        <CloudUploadIcon /> <Hidden xsDown>Data</Hidden>
+                        <CloudUploadIcon/> <Hidden xsDown>{getMessage("data")}</Hidden>
                     </Button>
                 </Grid>
                 <Grid item xs>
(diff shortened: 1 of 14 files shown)


src/components/navigation/Navigation.js
```diff
@@ -0,0 +1,126 @@
...
+import { build, getMessage, withI18N } from "@cyverse-de/ui-lib";
+import CloudUploadIcon from "@material-ui/icons/CloudUpload";
+import AppsIcon from "@material-ui/icons/Apps";
+import DashboardIcon from "@material-ui/icons/Dashboard";
+import AssessmentIcon from "@material-ui/icons/Assessment";
+import MoreHorizIcon from "@material-ui/icons/MoreHoriz";
+import { Button, Grid, Hidden, Paper } from "@material-ui/core";
+import { useRouter } from "next/router";
+import { makeStyles, useTheme } from "@material-ui/core/styles";
+import { injectIntl } from "react-intl";
```
ash3rz: I didn't see this actually used anywhere - we can probably remove it

src/components/navigation/Navigation.js
```diff
@@ -0,0 +1,126 @@
...
+const DASHBOARD = "dashboard";
+const DATA = "data";
+const APPS = "apps";
+const ANALYSES = "analyses";
+const MORE = "more";
+
+function Navigation(props) {
+    const classes = useStyles();
+    const router = useRouter();
+    const { navigate } = props;
```
ash3rz: Minor, but I was a little confused what this was for at first. Maybe `activeView` or even just `selected` might be a bit more clear?

src/pages/_app.js
```diff
@@ -4,6 +4,12 @@ import theme from "../components/theme/default";
 import { ThemeProvider } from "@material-ui/core/styles";
 
 function MyApp({ Component, pageProps }) {
+    React.useEffect(() => {
+        const jssStyles = document.querySelector("#jss-server-side");
+        if (jssStyles) {
+            jssStyles.parentElement.removeChild(jssStyles);
+        }
+    }, []);
```
ash3rz: 👍

src/pages/_document.js
```diff
@@ -0,0 +1,72 @@
...
+    // 1. document.getInitialProps
+    // 2. app.render
+    // 3. page.render
+    // 4. document.render
+    //
+    // On the client
+    // 1. app.getInitialProps
+    // 2. page.getInitialProps
+    // 3. app.render
+    // 4. page.render
```
ash3rz: 👍 ✨

stories/Navigation.stories.js
```diff
@@ -0,0 +1,13 @@
...
+import Navigation from "../src/components/navigation/Navigation";
+import { storiesOf } from "@storybook/react";
+
+class NavigationTest extends Component {
+    render() {
+        return <Navigation />;
+    }
+}
+
+storiesOf("Navigation", module).add("with navigation", () => (
```
ash3rz: Could save it for another PR, but there's actually a new format that Storybook is recommending on newer versions.   https://storybook.js.org/docs/formats/component-story-format/

src/components/navigation/Navigation.js
```diff
@@ -0,0 +1,126 @@
...
+        backgroundColor: theme.palette.blue,
+        color: theme.palette.white,
+    },
+}));
+
+const DASHBOARD = "dashboard";
+const DATA = "data";
+const APPS = "apps";
+const ANALYSES = "analyses";
+const MORE = "more";
```
ash3rz: Since the other pages that use this component also have to reference these strings, maybe we should move this to a separate file that this component and the pages reference?

src/components/navigation/Navigation.js
```diff
@@ -0,0 +1,126 @@
...
+                <Grid item xs>
+                    <Button
+                        id={build(ids.NAVIGATION, ids.DASHBOARD_BTN)}
+                        className={
+                            navigate === DASHBOARD
+                                ? classes.selectedButton
+                                : classes.button
+                        }
+                        onClick={() => {
+                            router.push("/dashboard");
```
ash3rz: I'm not sure if this could end up being more trouble than it's worth, but with the data listing, I had the page control the routing instead of the component.  My thinking was if we need to change the routing for whatever reason (like maybe `analyses` needs to get changed to `jobs` because people don't know the difference between that and `apps`, just as an example), we don't have to go digging through the components.  Again, not sure if that's worth all the potential passing around or not.  Not something we have to decide now though.

src/components/navigation/Navigation.js
```diff
@@ -0,0 +1,126 @@
...
+                    </Button>
+                </Grid>
+                <Grid item xs>
+                    <Button
+                        className={
+                            navigate === DATA
+                                ? classes.selectedButton
+                                : classes.button
+                        }
+                        onClick={() => {
```
ash3rz: Just noting for another PR that we'll need to add accessibility stuff

sri2k1us: What accessibility stuff is needed for buttons ? I couldn't find aria stuff for buttons in Material UI page. I do see that these need tools tips since the button text  wont show in small screens.

src/components/navigation/Navigation.js
sri2k1us: Not sure if I follow you here. There is no corresponding page for Navigation component 🤔 . My thinking was this Navigation component will handle all the logic for navigating from one view to the other.

src/components/navigation/Navigation.js
ash3rz: Ah, never mind, buttons already have keyboard accessibility automatically.  That's what I was thinking of.  Tool tips is a good point though 👍

src/components/navigation/Navigation.js
ash3rz: Oh, right 😊 Hm... although that makes me wonder, is there not a reason to put the navigation only in the `_app` page like the `CyVerseAppBar`?

sri2k1us: Thats possible. I had it that way at one point but changed it for some reason. I can give that a shot again.

src/pages/_app.js
```diff
@@ -4,6 +4,12 @@ import theme from "../components/theme/default";
 import { ThemeProvider } from "@material-ui/core/styles";
 
 function MyApp({ Component, pageProps }) {
+    React.useEffect(() => {
+        const jssStyles = document.querySelector("#jss-server-side");
+        if (jssStyles) {
+            jssStyles.parentElement.removeChild(jssStyles);
```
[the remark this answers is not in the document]
sri2k1us: Also mentioned in my commit message!

sri2k1us:
Thanks for reviews.

PR #25, opened 2020-01-24
Add support for file uploads.

johnworth:
Files can be POSTed to the authenticated /api/upload path using a form.

A test form is available on /upload, which will be updated later for
more functionality.

A snapshot was updated due to updates to dependencies that occurred when
the requests library was added.

To try things out with your local config:
* Pull the PR.
* npm ci
* npm run dev
* Access http://localhost:3000/login in a browser and login
* Access http://localhost:3000/testing/upload in a browser and upload a file
* If successful, check for the file in the DE.

The page at `/testing/upload` will go away once a more functional upload dialog is available.

### DIFF
--- a/src/__tests__/fileio.js
+++ b/src/__tests__/fileio.js
@@ -9,11 +9,10 @@ const mockRequest = (destination, token) => {
         query: {
             destination: destination || "",
         },
+        pipe: jest.fn(),
     };
 
-    let pipe = jest.fn();
-    pipe.mockImplementation(() => req);
-    req.pipe = pipe;
+    req.pipe.mockImplementation(() => req);
     return req;
 };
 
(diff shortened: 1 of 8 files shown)


src/server/index.js
```diff
@@ -6,6 +6,7 @@ import pgsimple from "connect-pg-simple";
 
 import * as config from "./configuration";
 import * as authStrategy from "./authStrategy";
+import { uploadHandler, testUploadForm } from "./fileio";
```
psarando: I don't see this `testUploadForm` import being used anywhere in this file.
In fact, I don't see it exported from `fileio` either 🤔

src/__tests__/fileio.js
```diff
@@ -0,0 +1,68 @@
+import { configurableUploadHandler } from "../server/fileio";
+import nock from "nock";
+
+const mockRequest = (destination, token) => {
+    let req = {
```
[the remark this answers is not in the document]
johnworth: Should be fixed now.

src/server/index.js
johnworth: Should be fixed now.

PR #26, opened 2020-01-28
Add Download Support

johnworth:
Add server support for file downloads.

Downloads are available through `GET /api/download?source=/path/to/file/in/data/store`.  The `Content-Type` header is set to `application/octet-stream` and the `Content-Disposition` is set to `attachment; filename=<basename of the source>`.

The nock tests were removed because I was having trouble adding a second nock-based test and having the tests themselves stay stable across runs. They might get re-added after more experimentation, but that will be in a separate PR.

No test page was needed for this API endpoint. To test it out:
* Pull PR.
* Run `npm run dev` (assuming you have a `config/local.yaml` file).
* Hit `http://localhost:3000/api/download?source=/iplant/home/<username>/<filename>` in a browser. A download should be initiated.

### DIFF
--- a/src/server/fileio.js
+++ b/src/server/fileio.js
@@ -35,9 +35,53 @@ export const configurableUploadHandler = (req, res, apiBaseURL) => {
         })
     ).pipe(res);
 };
+
 /**
  * Same as configurableUploadHandler, but defaults to using the configured
  * terrain handler.
  */
 export const uploadHandler = (req, res) =>
     configurableUploadHandler(req, res, terrainURL);
+
+/**
+ * Handles proxying file download requests to the Terrain API.
+ * @param {object} req - An Express Request object.
+ * @param {object} res - An Express Response object.
+ * @param {string} apiBaseURL - The base URL to the backing API. Usually an instance of the terrain API.
+ * @returns null
+ */
+export const configurableDownloadHandler = (req, res, apiBaseURL) => {
+    const token = req?.user?.accessToken;
+
+    if (!token) {
+        res.status(401).send("Authorization required.");
+        return;
+    }
+
+    const { source } = req.query;
+    if (!source || source === "") {
+        res.status(500).send("source query parameter must be set.");
+        return;
+    }
+
+    const apiURL = new URL(apiBaseURL);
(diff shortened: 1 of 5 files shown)


src/server/fileio.js
```diff
@@ -35,9 +37,59 @@ export const configurableUploadHandler = (req, res, apiBaseURL) => {
...
+        })
+    ).pipe(res);
+};
+
+/**
+ * Same as configurableDownloadHandler, but defaults to using the configured
+ * terrain handler.
+ */
+export const downloadHandler = (req, res) =>
+    configurableDownloadHandler(req, res, terrainURL);
```
ash3rz: 👍

src/server/fileio.js
```diff
@@ -35,9 +37,59 @@ export const configurableUploadHandler = (req, res, apiBaseURL) => {
...
+    const token = req?.user?.accessToken;
+
+    if (!token) {
+        res.status(401).send("Authorization required.");
+        return;
+    }
+
+    const { source } = req.query;
+    if (!source || source === "") {
+        res.status(500).send("source query parameter must be set.");
```
ianmcorvidae: A super tiny thing that doesn't really matter much, but this should probably be a 400 (client error) rather than 500 (server error)

johnworth: Good catch, thanks. I'll push a fix soon.

src/server/fileio.js
```diff
@@ -35,9 +37,59 @@ export const configurableUploadHandler = (req, res, apiBaseURL) => {
...
+    if (!source || source === "") {
+        res.status(400).send("source query parameter must be set.");
+        return;
+    }
+
+    const apiURL = new URL(apiBaseURL);
+    apiURL.pathname = path.join(apiURL.pathname, "/secured/fileio/download");
+    apiURL.search = `path=${source}`;
+
+    console.log(apiURL.href);
```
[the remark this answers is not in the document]
johnworth: Someone (I think Ian) mentioned that he liked the fact that backend request URLs were logged by Apollo, so I added the logging here since this code bypasses Apollo.

PR #29, opened 2020-01-31
Wrap app with apollo client

ash3rz:
Since we've decided we no longer need `getInitialProps` in `_app.js`, I followed [this example](https://nextjs.org/docs/advanced-features/custom-app) to simplify that code into just a function.


johnworth:
I'm going to go ahead and merge this so that I can rebase my changes on it

src/withApolloClient.js
```diff
@@ -0,0 +1,42 @@
...
+
+const withApolloClient = (App) => {
+    return class extends Component {
+        render() {
+            const link = createHttpLink({
+                uri: "/graphql",
+                credentials: "same-origin",
+                fetch: fetch,
+            });
+            const apolloClient = new ApolloClient({
```
psarando: It seems that this will create a new `ApolloClient` every time this `withApolloClient` component re-renders.
I'm not sure if this is required (maybe it will only render once per request or per user session?), or if we want to define 1 static `apolloClient` and `link` outside this wrapper component?

ianmcorvidae: I admit, this was my question too. The [apollo tutorial](https://www.apollographql.com/docs/react/get-started/#create-a-client) seems to create a `const` client directly in the namespace, then [passes that in](https://www.apollographql.com/docs/react/get-started/#connect-your-client-to-react) to the `ApolloProvider`. We do need to handle auth, of course. I notice that there's a function option you can configure the ApolloClient with, called [`request`](https://www.apollographql.com/docs/react/get-started/#configuration-options) that it mentions is useful for authentication purposes, that's called on every request and can be used to set `fetchOptions`. I didn't get much of a chance to dig into it beyond looking at docs, admittedly, so there may be a reason, but I admit that the approach with one client rather than a new one each render seems a lot cleaner to me.

ash3rz: That's a good point.  I could move the `link` and `apolloClient` creation to outside of the wrapper.  As for auth, I think we have that settled now by using `credentials: 'same-origin'`, but that was Sarah's finding so she probably knows more about it than I do.

PR #30, opened 2020-02-03
Move accessToken checking into middleware, route /api calls through Express Middleware

johnworth:
This pulls the logic for checking the accessToken out of the handlers and into a very basic Express middleware function.

Creates a new Express router and uses the new middleware on it. Change the /api/upload and /api/download endpoints to be routed with the new router.

This is all to show how we can proxy requests to terrain.

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -102,9 +102,12 @@ app.prepare()
         });
 
         const api = express.Router();
-        api.use(authStrategy.authnTokenMiddleware);
-        api.post("/upload", uploadHandler);
-        api.get("/download", downloadHandler);
+        api.post("/upload", authStrategy.authnTokenMiddleware, uploadHandler);
+        api.get(
+            "/download",
+            authStrategy.authnTokenMiddleware,
+            downloadHandler
+        );
 
         server.use("/api", api, (req, res) => {
             res.sendStatus(401);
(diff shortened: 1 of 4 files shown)


src/server/index.js
```diff
@@ -101,8 +101,14 @@ app.prepare()
...
-        server.post("/api/upload", uploadHandler);
-        server.get("/api/download", downloadHandler);
+        const api = express.Router();
+        api.use(authStrategy.authnTokenMiddleware);
+        api.post("/upload", uploadHandler);
+        api.get("/download", downloadHandler);
+
+        server.use("/api", api, (req, res) => {
+            res.sendStatus(401);
+        });
```
slr71: I was thinking we'd probably let terrain handle returning the 401, and if terrain is managing the decision about whether or not auth is required then we can probably make due without the middleware. At the very least, I think we'd want the middleware to be applied to specific endpoints where we know it'll be required. Otherwise, it becomes harder to add API endpoints that allow anonymous API access later.

johnworth: I modified the middleware to add the token to the request so it will get passed along to terrain automatically when the req.pipe() function is used.

I've also modified it so that the middleware is applied on a per-router basis, not to the entire router.

johnworth: I've just modified the middleware to only pass along the authorization header if the token is present in the user object.

PR #35, opened 2020-02-05
Configure jsx-a11y plugin for eslint. 

johnworth:
This also adds a lint script.

I think this is looking over now and commenting on even though it's a WIP. We'll have to fix up the existing components before merging, though, since they have existing a11y issues that are causing the CI actions to fail with the new eslint config.

### DIFF
--- a/src/__tests__/__snapshots__/apps.js.snap
+++ b/src/__tests__/__snapshots__/apps.js.snap
@@ -20,11 +20,15 @@ exports[`App Tile renders 1`] = `
       <div
         className="_class-avatar-106"
       >
-        <div>
+        <div
+          onClick={[Function]}
+          onKeyPress={[Function]}
+          role="button"
+          tabIndex="0"
+        >
           <img
             alt="avatar"
             id="appListing.676846d4-854a-11e4-980d-7f0fcca75dbb.card"
-            onClick={[Function]}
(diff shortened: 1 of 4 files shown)


psarando:
Yeah, let's give it a shot. Maybe all the warnings should be fixed and included in this PR (as you mentioned), then we can an idea of everything this is catching.

johnworth:
> Yeah, let's give it a shot. Maybe all the warnings should be fixed and included in this PR (as you mentioned), then we can an idea of everything this is catching.

I've updated the pull request to fix the reported a11y issues. It's no longer in the WIP state.

src/__tests__/__snapshots__/apps.js.snap
```diff
@@ -1162,11 +1176,15 @@ exports[`App Tile renders 1`] = `
       <div
         className="_class-avatar-106"
       >
-        <div>
+        <div
+          onClick={[Function]}
+          onKeyPress={[Function]}
```
ianmcorvidae: I hate to be the bearer of bad news, but I think `onKeyPress` is the non-recommended one here D:

https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onkeypress mentions it's deprecated and recommends using onkeydown instead.

slr71: Ah, good catch.

PR #41, opened 2020-02-11
Global styles

sri2k1us:
- Add Roboto font family
- Remove Body scroll. AppBar, NavBar and Table header will remain sticky.
- Add CyVerse Fav icon 
This is on top #23. So it is sufficient to review only last 4 commits.

image

image


src/components/data/listing/TableView.js
```diff
@@ -87,7 +87,7 @@ function TableView(props) {
     const tableColumns = getTableColumns(isMedium, isLarge);
 
     return (
-        <TableContainer component={Paper}>
+        <TableContainer component={Paper} style={{minHeight: 480, maxHeight: isMedium ? 1024 : 600}}>
```
ash3rz: 🤔 Do we want a max height?

sri2k1us: I don't think scrollbars will show up without maxHeight

src/server/index.js
```diff
@@ -131,9 +131,15 @@ app.prepare()
         logger.info("adding the api router to the express server");
         server.use("/api", api);
 
+        //map root to Dashboard
+        server.get("/", (req, res) => {
+            app.render(req, res, "/dashboard", undefined);
+        });
```
johnworth: This feels like routing logic that should be in the UI instead of the server. Maybe doing something like this somewhere? 

https://stackoverflow.com/a/58182678

src/components/data/listing/TableView.js
johnworth: I don't know if it's related to the maxHeight, but I can't see the footer on the table when the full listing is > 1 page.

src/server/index.js
sri2k1us: Good catch 👍 . With express, the actual doesn't get updated so I wasn't really happy it anyway.

src/components/data/listing/TableView.js
```diff
@@ -87,7 +87,13 @@ function TableView(props) {
     const tableColumns = getTableColumns(isMedium, isLarge);
 
     return (
-        <TableContainer component={Paper}>
+        <TableContainer
+            component={Paper}
+            style={{
+                minHeight: 320,
+                maxHeight: isMedium || isLarge ? 800 : 320,
```
ash3rz: I _think_ (haven't tried this) we can get away with doing something like `100vh` (see Relative Lengths [here](https://www.w3schools.com/cssref/css_units.asp)).  The only thing is there might have to be a calculation to subtract the height of the app bar and navigation for it to work properly.  That could get complicated, but I think it might serve us better than having a set height.  (Again, don't know if it'll actually work, but I think it will)

src/components/navigation/Navigation.js
```diff
@@ -19,18 +19,139 @@ import { makeStyles, useTheme } from "@material-ui/core/styles";
...
+        borderRadius: 0,
+        width: 200,
+        height: 50,
+        backgroundColor: theme.palette.primary.main,
+        color: theme.palette.white,
+    },
+
+    dashboardButton: {
+        backgroundImage: `url("/dashboard.png")`,
+        backgroundPosition: "15px 10px",
```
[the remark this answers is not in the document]
sri2k1us: I can give that a shot!

src/components/appBar/CyVerseAppBar.js
```diff
@@ -112,71 +105,55 @@ function CyverseAppBar(props) {
...
-                            id={build(ids.APP_BAR_BASE_ID, ids.SEARCH)}
-                        >
+                        <div className={classes.search}>
                             <div className={classes.searchIcon}>
-                                <SearchIcon />
+                                <SearchIcon color="primary" />
                             </div>
                             <InputBase
-                                placeholder={formatMessage(intl, "search")}
+                                placeholder="Search…"
```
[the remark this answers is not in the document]
sri2k1us: Actually I rewrote most of the code in this page to get the new Look and Feel. This is more of a WIP now since, I still have to get it working with small screens.

src/components/appBar/CyVerseAppBar.js
```diff
@@ -225,7 +202,10 @@ function SearchOptions(props) {
                 <Button
                     id={build(ids.APP_BAR_BASE_ID, ids.SEARCH_FILTER_BTN)}
                     onClick={handleClick}
-                    style={{ backgroundColor: theme.palette.white }}
+                    style={{
+                        backgroundColor: theme.palette.primary.main,
+                        color: theme.palette.white,
```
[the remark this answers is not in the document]
sri2k1us: > LGTM 👍
> 
> I made a few minor `theme` suggestions, and it seems like we need i18n messages in the app bar.
> I also agree with @ash3rz that we should try out a relative height for the table contents.

Thanks for reviews. I am gonna close this PR for now since I still have to get it working with small screens and do some cleans ups.

sri2k1us:
> LGTM 👍
> 
> I made a few minor `theme` suggestions, and it seems like we need i18n messages in the app bar.
> I also agree with @ash3rz that we should try out a relative height for the table contents.

PR #43, opened 2020-02-11
Move terrain route handling to separate file.

johnworth:
Moves the terrain-related API routes into a file separate from src/server/index.js.


src/server/terrainRouter.js
```diff
@@ -0,0 +1,89 @@
+import request from "request";
+import path from "path";
+import querystring from "querystring";
+import { terrainURL } from "./configuration";
+import * as authStrategy from "./authStrategy";
+import logger from "./logging";
+import express from "express";
+
+const terrain = ({ method, url, pathname, headers }) => {
```
sri2k1us: Some JSDoc for each method here will be helpful.

PR #44, opened 2020-02-12
Data header bar

ash3rz:
This will add the header to the listing. Still need to update the colors for the table column header so it's white, but I'll do that in a separate PR.

Any largish size:
image

Upload Menu:
image

Mobile view:
image

Temporary view for the list/thumbnail view with the table icon shown:
image


ash3rz:
Thanks for the review!

PR #46, opened 2020-02-13
CORE-868: provide a reasonably easy way for the user to be returned t…

slr71:
…o the current URL after logging in

This approach does require the UI to redirect the user to a different URL depending on where the user is currently. Suppose, for example that the current URL path is `/data?path=/foo/bar/baz`. If the user chooses to log in, the UI will redirect the user to `/login/data?path=/foo/bar/baz`. Once authentication is complete, the user will be redirected back to `/data?path=/foo/bar/baz` with the user's identification in the request.

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -3,6 +3,7 @@ import next from "next";
 import passport from "passport";
 import session from "express-session";
 import pgsimple from "connect-pg-simple";
+import { ensureLoggedIn } from "connect-ensure-login";
 
 import * as config from "./configuration";
 import * as authStrategy from "./authStrategy";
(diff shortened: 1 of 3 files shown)


src/server/index.js
```diff
@@ -66,6 +67,11 @@ app.prepare()
             })
         );
 
+        logger.info("adding the /login/* handler");
+        server.get("/login/*", ensureLoggedIn("/login"), (req, res) => {
```
ash3rz: From what I can tell, this should be doable.  Looks like `next/router` has an `asPath` method that should return the current page.  So from the UI side, we'd be navigating to `/login/${router.asPath}`.

slr71:
Thanks for the review!

PR #47, opened 2020-02-14
Bump ui-lib version, update table row component, update theme

ash3rz:
Not 100% sure if this is what we intended to do with the theme in sonora or not.


ash3rz:
Thanks for the review!  I'll do a quick rebase before merging.

PR #51, opened 2020-02-18
Drag-n-drop file uploads

johnworth:
This PR depends on https://github.com/cyverse-de/sonora/pull/48.

Adds the ability to drop files into the data listing and have them be uploaded to the directory being listed. The listing should auto-refresh once the uploads complete.

This uses the UploadTrackingProvider and associated hooks to add to the global uploads state. This will eventually be used by an Upload Queue component to list the in-progress uploads.

A story called `Uploads/UploadDropTarget/Upload Drop Target Test` with a mocked up API call can be used to see the how the number of tracked and completed uploads is changed when you drop files onto a listing.

Most likely, additional changes will be made to this code in the PR that adds the Upload Queue.

Screenshot without drag-n-drop:

Screen Shot 2020-02-24 at 8 53 16 AM (3)

With drag-n-drop:

Screen Shot 2020-02-24 at 8 53 30 AM


src/components/data/listing/UploadDrop.js
```diff
@@ -0,0 +1,144 @@
...
+
+    // Get all of the files split out into their own list.
+    let fileItems = allItems
+        .filter((i) => i.kind === "file")
+        .map((i) => i.getAsFile())
+        .filter((f) => f.size > 0) // filter out 0-byte files and directories.
+        .map((i) => ({ kind: KindFile, value: i }));
+
+    // Clear out any nulls from the list.
+    fileItems = fileItems.filter((i) => i !== null && i !== "undefined");
```
slr71: I don't quite understand this line. It would seem to me that the last `map` in the previous line would mean that each of the items in `fileItems` would be an object and this line would have no effect. Am I missing something here?

src/common/callApi.js
```diff
@@ -1,17 +1,24 @@
...
+ * @module callApi
+ */
+
+/**
  * A function which should help standardize how fetch calls are handled.
  * The promise returned from fetch is returned to the caller so more
  * can be chained to it.
  *
- * Sample usages:
+ * @example <caption>Basic GET request</caption>
```
psarando: 👍

src/common/callApi.js
```diff
@@ -48,24 +54,114 @@ const callApi = (props) => {
...
+ * @param {object} props - An object containing method, endpoint, and headers fields.
+ * @returns {object}
+ */
+export const checkForError = async (resp, { method, endpoint, headers }) => {
+    if (!resp.ok) {
+        throw await getAPIErrorFromResponse(resp, method, endpoint, headers);
+    }
+
+    return resp;
+};
```
psarando: 👍

src/components/data/listing/UploadDropTarget.js
```diff
@@ -0,0 +1,104 @@
...
+ * Performs file uploads to the `path` directory in the data store when
+ * files are dropped onto the child components.
+ *
+ * @param {UploadDropTargetProps} props - The props for the component.
+ * @returns {Object}
+ */
+const UploadDropTarget = (props) => {
+    const uploadDispatch = useUploadTrackingDispatch();
+    const [dragCounter, setDragCounter] = useState(0);
+    const { children, path, uploadCompletedCB } = props;
```
psarando: It took me a second to parse `CB`.
I think I'm used to seeing callbacks named something like `onUploadCompleted`, but it's not a problem.

src/components/data/listing/UploadDropTarget.js
```diff
@@ -0,0 +1,104 @@
...
+    const handleDragOver = (event) => {
+        setupEvent(event);
+    };
+
+    const handleDragEnter = (event) => {
+        setupEvent(event);
+
+        // This event fires for all sub-elements that the items are dragged over,
+        // so we track the number of elements and check it below to prevent the
+        // border color from flickering for every element that's passed over.
```
psarando: 👍

src/components/data/listing/UploadDropTarget.js
```diff
@@ -0,0 +1,104 @@
...
+        if (event.dataTransfer.items && event.dataTransfer.items.length > 0) {
+            setDragCounter(dragCounter + 1);
+        }
+    };
+
+    const handleDragLeave = (event) => {
+        setupEvent(event);
+
+        // See the comment in handleDragIn() for why we're doing this.
```
psarando: 👍 
but I think you mean `handleDragEnter`

src/components/data/listing/UploadDrop.js
psarando: I'm thinking the same thing, but if there ever is a chance of a null in this list, then I would think this `filter` should be called first on `allItems`.

Also, I think this could be simplified as `allItems.filter(i => i)`, since `null` and `undefined` are already falsey (and I don't think we want to keep other falsey values like `0`).

psarando: Or the very first `filter` can guard against nulls:
```javascript
    let fileItems = allItems
        .filter((i) => (i && i.kind === "file"))
```

johnworth: This should be fixed now.

PR #52, opened 2020-02-18
Add listing component story with mocked API, extract fetch usage

ash3rz:
The big thing here is extracting `fetch` so that we can standardize logging and potentially error handling during API calls.  
There are some considerations there.  What I really wanted was a HOC that passed down this `callAPI` function plus `loading` and `error` props because I foresee us doing `const [loading, setLoading] = useState(false); const [error, setError] = useState(null);` everywhere, but I couldn't get it to work without encountering lots of lint warnings or taking away flexibility from the components using it.  For now, I've gone with this to keep it simple-ish.

### DIFF
--- a/src/components/data/listing/Listing.js
+++ b/src/components/data/listing/Listing.js
@@ -12,13 +12,14 @@ import { TablePagination, useMediaQuery, useTheme } from "@material-ui/core";
 import Header from "../Header";
 import messages from "../messages";
 import TableView from "./TableView";
+import callApi from "../../../common/callApi";
 
 import { camelcaseit } from "../../../common/functions";
 
 function Listing(props) {
     const theme = useTheme();
-    let isMedium = useMediaQuery(theme.breakpoints.up("sm"));
-    let isLarge = useMediaQuery(theme.breakpoints.up("lg"));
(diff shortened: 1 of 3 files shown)


stories/data/Listing.stories.js
```diff
@@ -0,0 +1,157 @@
...
+        },
+    ],
+};
+
+export const DataListingTest = () => {
+    fetchMock
+        .restore()
+        .get(/\/api\/filesystem\/paged-directory.*/, successResp);
+    return <ListingTest />;
+};
```
sri2k1us: 🎉 👍

ash3rz:
Thanks for the reviews!

src/components/data/listing/Listing.js
```diff
@@ -37,53 +38,23 @@ function Listing(props) {
...
-            // We need to figure out a consistent error handler.
-            .catch((e) => {
-                console.log(`error ${e.message}`)
-                setLoading(false);
-            });
+            setLoading,
+        }).then((respData) => {
+            respData &&
+                setData({
+                    total: respData?.total,
```
psarando: Since there's already a `respData && setData` check above, does that make these `respData?.` checks redundant?

PR #53, opened 2020-02-19
Intercom

sri2k1us:


src/pages/_app.js
```diff
@@ -1,14 +1,66 @@
...
+        : NavigationConstants.DASHBOARD;
 
-function MyApp({ Component, pageProps }) {
     React.useEffect(() => {
         const jssStyles = document.querySelector("#jss-server-side");
         if (jssStyles) {
             jssStyles.parentElement.removeChild(jssStyles);
         }
+        const APP_ID = intercomAppId;
+        const enabled = intercomEnabled;
```
ash3rz: Minor, but could we not just use `intercomAppId` and `intercomEnabled` directly?

sri2k1us:
> LGTM :shipit:
> My only question is if it's going to still be bottom-right? I was thinking since the app bar is going to be sticky no matter what page you're on, maybe it should be within the app bar (if the widget can be made small enough to look ok). Although, I'm not sure how that would look on mobile.
Yes this will render the intercom widget in the same location as in the current DE. Am not sure if we will able to adjust icon size or position.  I had a hard time adjusting its position last time I integrated it. I can take a stab at it in a separate PR.

PR #55, opened 2020-02-19
CORE-868: allow the client-side code to obtain user profile information.

slr71:
The motivation for this change is to allow the client-side code to alter its appearance depending on whether or not the user is currently authenticated, and the groups that the user is in. The profile that is returned will be `null` if the user isn't authenticated. Otherwise, it will be an object containing the following members:

```json
{
    "attributes": {
        "email": "username@domain.org",
        "entitlement": [ "group1", "group2", "group3" ],
        "firstName": "User's First Name",
        "lastName": "User's Last Name",
        "name": "User's Full Name",
    },
    "id": "username"
}
```

The `entitlement` attribute contains the list of LDAP groups that the user belongs to. The client-side code may choose to alter its appearance based on the groups that the user belongs to. For example, if the user belongs to one of the administrative groups, additional UI components might be displayed.

The `CyVerseAppBar` component is currently responsible for retrieving the user profile information and storing it in the user profile context. Other components can then use the `useUserProfile` hook to retrieve the user profile information:

```javascript
import { useUserProfile } from "/path/to/context/userProfile.js";

function SomeReactComponent(props) {
    // ...
    console.log(userProfile);
    // ...
}
```

### DIFF
--- a/src/components/appBar/CyVerseAppBar.js
+++ b/src/components/appBar/CyVerseAppBar.js
@@ -9,6 +9,7 @@ import React from "react";
 import intlData from "./messages";
 import ids from "./ids";
 import constants from "../../constants";
+import callApi from "../../common/callApi";
 
 import {
     build,
@@ -108,18 +109,18 @@ function CyverseAppBar(props) {
 
     React.useEffect(() => {
         const fetchUserProfile = async function() {
-            const profile = await fetch("/api/profile", {
+            const profile = await callApi({
+                endpoint: "/api/profile",
                 method: "GET",
                 credentials: "include",
-            })
-                .then((resp) => resp.json())
-                .catch((e) => console.log(`error ${e.message}`));
+            });
             setUserProfile(profile);
         };
         fetchUserProfile(setUserProfile);
     }, [setUserProfile]);
 
     const handleUserButtonClick = (event) => {
+        console.log(userProfile);
         if (!userProfile) {
             router.push(`/login${router.asPath}`);
         }
--- a/src/components/appBar/CyVerseAppBar.js
+++ b/src/components/appBar/CyVerseAppBar.js
@@ -116,7 +116,7 @@ function CyverseAppBar(props) {
             });
             setUserProfile(profile);
         };
-        fetchUserProfile(setUserProfile);
+        fetchUserProfile();
     }, [setUserProfile]);
 
     const handleUserButtonClick = (event) => {
--- a/src/components/appBar/CyVerseAppBar.js
+++ b/src/components/appBar/CyVerseAppBar.js
@@ -105,7 +105,7 @@ function CyverseAppBar(props) {
     const handleUserButtonClick = (event) => {
         const { user } = props.children.props;
         if (!user) {
-            router.push("/login");
+            router.push(`/login${router.asPath}`);
         }
     };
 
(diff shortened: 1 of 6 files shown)


src/components/appBar/CyVerseAppBar.js
```diff
@@ -102,10 +104,24 @@ function CyverseAppBar(props) {
...
+        };
+        fetchUserProfile(setUserProfile);
+    }, [setUserProfile]);
+
     const handleUserButtonClick = (event) => {
-        const { user } = props.children.props;
-        if (!user) {
-            router.push("/login");
+        if (!userProfile) {
+            router.push(`/login${router.asPath}`);
```
ash3rz: 👍

src/components/appBar/CyVerseAppBar.js
```diff
@@ -102,10 +104,24 @@ function CyverseAppBar(props) {
...
+            const profile = await fetch("/api/profile", {
+                method: "GET",
+                credentials: "include",
+            })
+                .then((resp) => resp.json())
+                .catch((e) => console.log(`error ${e.message}`));
+            setUserProfile(profile);
+        };
+        fetchUserProfile(setUserProfile);
+    }, [setUserProfile]);
```
ash3rz: I'll have to read up on contexts more because `setUserProfile` is a useEffect dependency I did not expect that I find interesting 🤔 
There's also the `callApi` function in the `common` folder now if you wanted to use that for `fetch`, but since your usage is so small, it doesn't really matter.

slr71: It turned out to be easiest to make the API call from within a React component (because `fetch` wasn't available otherwise), so I ended up not putting the call inside the context. `callApi` sounds really useful. I'll probably update the code to use it instead.

slr71: I was getting warnings when I didn't pass in `setUserProfile` as a dependency. My understanding is that because it's available outside of the `useEffect` hook, it has to be listed as a dependency. Otherwise, if `setUserProfile` changed somehow, the user profile wouldn't be retrieved again.

`setUserProfile` doesn't need to be passed to setUserProfile. I was originally planning on moving `fetchUserProfile` to another file but then decided against it. I removed the parameter from the function definition, but forgot to remove it from the function call. 😖 Thanks for catching that.

slr71:
Thanks for the reviews!

src/pages/_app.js
```diff
@@ -13,11 +14,13 @@ function MyApp({ Component, pageProps }) {
     }, []);
     return (
         <ThemeProvider theme={theme}>
-            <UploadTrackingProvider>
-                <CyverseAppBar>
-                    <Component {...pageProps} />
-                </CyverseAppBar>
-            </UploadTrackingProvider>
+            <UserProfileProvider>
```
psarando: It looks like we need this Provider wrapper added as a Storybook decorator, similar to this:
https://github.com/cyverse-de/sonora/blob/e5d2ea1f3030395abb41a47ac787763c3a2b32bd/.storybook/config.js#L7-L9

Currently the AppBar story is throwing an error: `useUserProfile must be used within a UserProfileProvider`

slr71: Ah, yes we do. Thanks for catching this. I'll get that fixed sometime today.

PR #59, opened 2020-02-24
CORE-868: trust reverse proxies

slr71:


slr71:
I'm going to merge this one immediately because I want to do some testing in QA. Reviews are still welcome.

PR #61, opened 2020-02-27
Upload Queue Dialog

johnworth:
Most of the upload-related components and code is contained in the `src/components/uploads` directory now. 

The `UploadManager` in `components/uploads/manager` is a non-visible component that accesses the upload tracking context and triggers the actual API calls. The `UploadManager` tag was added to `src/pages/_app.js` because we want the uploads to progress regardless of the page the user navigates to, as long as they keep the tab/window open.

The `UploadQueue` in `components/uploads/queue` renders a list of tracked uploads contained in the upload tracking context.

The `UploadDialog` in `components/uploads/dialog` renders a dialog containing an `UploadQueue`. It may eventually become an drawer of some sort, but that's not happening now. The `UploadDialog` is accessible through the Data Listing's Uploads Menu. A story was added to render the `UploadDialog`.

IDs and i18n should be in place.

The "dirty bit" to trigger a warning to the user has not been added. That will come in a later PR.

The errors indicated in the `Upload Dialog` do not have messages displayed as of yet. I'm guessing we're going to want that info in the `Alerts` page and popping up in snackbars, but AFAIK neither of those is available yet.

Screen Shot 2020-02-27 at 2 56 41 PM

Screen Shot 2020-02-27 at 2 57 22 PM

Screen Shot 2020-02-27 at 3 20 35 PM

Mobile, from storybook:

Screen Shot 2020-02-27 at 3 22 21 PM

### DIFF
--- a/src/components/uploads/dialog/index.js
+++ b/src/components/uploads/dialog/index.js
@@ -1,3 +1,10 @@
+/**
+ * @author johnworth
+ *
+ * A dialog displaying the tracked uploads.
+ *
+ * @module uploads/dialog
+ */
 import React from "react";
 
 import {
--- a/src/components/uploads/dialog/index.js
+++ b/src/components/uploads/dialog/index.js
@@ -78,6 +78,7 @@ const UploadDialog = ({ open, handleClose = () => {} }) => {
                     id={buildID(ids.BASE, ids.CLOSE)}
                     onClick={handleClose}
                     color="primary"
+                    aria-label="close"
                 >
                     {getMessage("close")}
                 </Button>
--- a/src/components/uploads/dialog/index.js
+++ b/src/components/uploads/dialog/index.js
@@ -61,7 +61,7 @@ const UploadDialog = ({ open, handleClose = () => {} }) => {
                 <DialogContentText>
                     {getMessage("help")}
                     <IconButton
-                        aria-label="close"
+                        aria-label={getMessage("closeAria")}
                         id={buildID(ids.BASE, ids.CLOSE_X)}
                         className={classes.closeDialog}
                         onClick={handleClose}
@@ -78,7 +78,7 @@ const UploadDialog = ({ open, handleClose = () => {} }) => {
                     id={buildID(ids.BASE, ids.CLOSE)}
                     onClick={handleClose}
                     color="primary"
-                    aria-label="close"
+                    aria-label={getMessage("closeAria")}
                 >
                     {getMessage("close")}
                 </Button>
--- a/src/components/uploads/dialog/index.js
+++ b/src/components/uploads/dialog/index.js
@@ -1,0 +1,82 @@
+import React from "react";
+
+import {
+    Button,
+    Dialog,
+    DialogActions,
+    DialogContent,
+    DialogContentText,
+    DialogTitle,
+    IconButton,
+    useMediaQuery,
+} from "@material-ui/core";
+
+import { Close as CloseIcon } from "@material-ui/icons";
+
+import { useTheme, makeStyles } from "@material-ui/styles";
+
+import { getMessage, withI18N, build as buildID } from "@cyverse-de/ui-lib";
+
+import UploadQueue from "../queue";
+
+import messages from "./messages";
+import ids from "./ids";
+
+const useStyles = makeStyles((theme) => ({
+    closeDialog: {
+        position: "absolute",
+        right: theme.spacing(1),
+        top: theme.spacing(1),
+        color: theme.palette.blueGrey,
+    },
+}));
+
+/**
+ * The upload dialog component that displays the list of uploads being tracked.
+ *
+ * @params {Object} props
+ * @params {boolean} props.open - Whether or not the dialog is open.
+ * @params {Object} props.handleClose - Callback that is executed when the dialog is closed.
+ * @returns {Object}
+ */
+const UploadDialog = ({ open, handleClose = () => {} }) => {
+    const classes = useStyles();
+    const theme = useTheme();
+    const fullScreen = useMediaQuery(theme.breakpoints.down("sm"));
+
+    return (
+        <Dialog fullScreen={fullScreen} open={open} onClose={handleClose}>
+            <DialogTitle id={buildID(ids.BASE, ids.TITLE)}>
+                {getMessage("title")}
+            </DialogTitle>
+
+            <DialogContent>
+                <DialogContentText>
+                    {getMessage("help")}
+                    <IconButton
+                        aria-label="close"
+                        id={buildID(ids.BASE, ids.CLOSE_X)}
+                        className={classes.closeDialog}
+                        onClick={handleClose}
+                    >
+                        <CloseIcon />
+                    </IconButton>
+                </DialogContentText>
+
+                <UploadQueue id={buildID(ids.BASE, ids.LIST)} />
+            </DialogContent>
+
+            <DialogActions>
+                <Button
+                    id={buildID(ids.BASE, ids.CLOSE)}
+                    onClick={handleClose}
+                    color="primary"
+                >
+                    {getMessage("close")}
+                </Button>
+            </DialogActions>
+        </Dialog>
+    );
+};
+
+export default withI18N(UploadDialog, messages);
--- a/src/components/uploads/dialog/messages.js
+++ b/src/components/uploads/dialog/messages.js
@@ -3,6 +3,7 @@ export default {
     messages: {
         title: "Upload Queue",
         help: "The list of uploads running in this browser tab or window.",
+        closeAria: "close",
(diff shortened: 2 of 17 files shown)


src/pages/_app.js
```diff
@@ -1,15 +1,59 @@
...
+            x.parentNode.insertBefore(s, x);
+        }
+
+        if (window.attachEvent) {
+            window.attachEvent("onload", loadWidget);
+        } else {
+            window.addEventListener("load", loadWidget, false);
+        }
+    }
+};
```
sri2k1us: 👍

src/components/uploads/queue/index.js
```diff
@@ -0,0 +1,170 @@
...
+
+import { useTheme, makeStyles } from "@material-ui/core/styles";
+
+const useStyles = makeStyles((theme) => ({
+    ellipsis: {
+        overflow: "hidden",
+        textOverflow: "ellipsis",
+        paddingRight: theme.spacing(1),
+    },
+}));
```
sri2k1us: does this ellipsis style work ? I only ask bcos I had to follow this https://stackoverflow.com/questions/17779293/css-text-overflow-ellipsis-not-working to get it working.

src/components/uploads/manager/index.js
```diff
@@ -0,0 +1,103 @@
...
+    const [waiting, running] = partitionUploads(tracker.uploads);
+
+    // Controls starting the uploads.
+    useEffect(() => {
+        // Browser can technically do more than 3 uploads at once now, but
+        // we're limiting it here to prevent hammering our API too much.
+        if (
+            tracker.uploads.length > 0 &&
+            waiting.length > 0 &&
+            running.length < 3
```
sri2k1us: This number can be moved to a constants file to make it easier to update later.

src/contexts/uploadTracking.js
```diff
@@ -6,12 +6,15 @@
...
- * @module uploadTracking
+ * @module contexts/uploadTracking
  */
 
 import React from "react";
 import UUID from "uuid/v4";
 
+export const KindFile = "file";
+export const KindURL = "url";
+
```
sri2k1us: may just `FILE` and `URL` but I like being `kind` 😄

sri2k1us:
LGTM. I had some minor / trivial comments. Also one question for the reviewers - So far components are named after their use, but I see @johnworth going with `index.js` convention. Which one are we going follow ? I don't mind either way but it will be good to be conistent.

src/components/uploads/queue/index.js
johnworth: Hmm, odd. It worked fine for me.

src/components/uploads/manager/index.js
johnworth: This should be fixed now.

src/components/uploads/dialog/index.js
```diff
@@ -0,0 +1,90 @@
...
+        <Dialog fullScreen={fullScreen} open={open} onClose={handleClose}>
+            <DialogTitle id={buildID(ids.BASE, ids.TITLE)}>
+                {getMessage("title")}
+            </DialogTitle>
+
+            <DialogContent>
+                <DialogContentText>
+                    {getMessage("help")}
+                    <IconButton
+                        aria-label="close"
```
ash3rz: Only commenting once, but I think any aria text should still be internationalized

src/components/uploads/manager/index.js
```diff
@@ -0,0 +1,105 @@
...
+        ) {
+            waiting.forEach((upload, idx) => {
+                if (idx <= 2) {
+                    startUpload(upload, upload.parentPath, dispatch);
+                }
+            });
+        }
+    }, [tracker, waiting, running, dispatch]);
+
+    return <div id="upload-manager" />;
```
ash3rz: Minor, but this should be in an IDs file

src/components/uploads/manager/index.js
```diff
@@ -0,0 +1,105 @@
...
+    useEffect(() => {
+        // Browser can technically do more than 3 uploads at once now, but
+        // we're limiting it here to prevent hammering our API too much.
+        if (
+            tracker.uploads.length > 0 &&
+            waiting.length > 0 &&
+            running.length < MAX_RUNNING_UPLOADS
+        ) {
+            waiting.forEach((upload, idx) => {
+                if (idx <= 2) {
```
ash3rz: This is completely trivial and feel free to ignore, but every time I see this I try to figure out what `idx` stands for 😄  If it were `index` or even just `i` I wouldn't pause.

src/components/uploads/queue/index.js
```diff
@@ -0,0 +1,170 @@
...
+            </ListItemAvatar>
+
+            {isSmall ? (
+                <ListItemText
+                    primary={<EllipsisField>{upload.filename}</EllipsisField>}
+                />
+            ) : (
+                <ListItemText
+                    primary={<EllipsisField>{upload.filename}</EllipsisField>}
+                    secondary={upload.parentPath}
```
ash3rz: 👍

src/components/uploads/dialog/index.js
```diff
@@ -0,0 +1,90 @@
+/**
+ * @author johnworth
```
ash3rz: Picking a random spot to respond to Sriram's question - I'm not personally used to seeing index files, so I would prefer file names that indicate the purpose.  That's just my opinion though.

psarando:
> Also one question for the reviewers - So far components are named after their use, but I see @johnworth going with `index.js` convention. Which one are we going follow ? I don't mind either way but it will be good to be conistent.

I also have no strong opinion either way. I think I initially went without the `index.js` files in the current UI just to try and mimic the :java: namespaces in the GWT code, but that's obviously not needed in this code.

The `index.js` convention does remind me of apps in OSX, where something like `Slack.app` is really a folder with a bunch of other files and folders under it.

src/components/uploads/dialog/messages.js
```diff
@@ -0,0 +1,9 @@
+export default {
+    locales: "en-us",
+    messages: {
+        title: "Upload Queue",
+        help: "The list of uploads running in this browser tab or window.",
+        closeAria: "close",
+        close: "Close",
```
psarando: I don't think these have to be different strings. As far as I know, the aria label is going to be read by a screen reader, so these could be the same: https://www.w3.org/TR/wai-aria/#aria-label

Actually, after reading that link, it appears that an aria label isn't required in this case, since the button already has the visible text "Close" that a screen reader should use.

src/components/uploads/queue/messages.js
```diff
@@ -0,0 +1,6 @@
+export default {
+    locales: "en-us",
+    messages: {
+        cancelUploadAria: "cancel-upload",
```
psarando: This could simply be the string "Cancel Upload".

src/components/uploads/manager/index.js
psarando: Should this be `if (idx < MAX_RUNNING_UPLOADS)`?

psarando: idk 🤔

src/components/uploads/manager/index.js
```diff
@@ -0,0 +1,105 @@
...
+                case running:
+                    acc[running].push(upload);
+                    break;
+                default:
+                    acc[other].push(upload);
+            }
+
+            return acc;
+        },
+        [[], [], []]
```
psarando: Personally I would have used an object:
```javascript
                case waiting:
                    acc.waiting.push(upload);
                    break;
                case running:
                    acc.running.push(upload);
                    break;
                default:
                    acc.other.push(upload);
            }

            return acc;
        },
        {waiting: [], running: [], other: []}
```

Then below:
```javascript
const {waiting, running} = partitionUploads(tracker.uploads);
```

But there's also nothing wrong with using an array here.

src/components/uploads/queue/index.js
```diff
@@ -0,0 +1,176 @@
...
+import {
+    CheckCircle as CheckCircleIcon,
+    Cancel as CancelIcon,
+    Error as ErrorIcon,
+    Description as DescriptionIcon,
+    Http as HttpIcon,
+} from "@material-ui/icons";
+
+import {
+    KindFile,
```
psarando: A `File` that will rewind 📼

src/contexts/uploadTracking.js
```diff
@@ -169,7 +208,7 @@ const uploadReducer = (state, action) => {
         // state.uploads array.
         case "add": {
             const newState = { ...state };
-            newState.uploads.unshift(action.upload);
+            newState.uploads = [...newState.uploads, action.upload];
```
psarando: 👍

PR #62, opened 2020-02-28
Data nav

sri2k1us:
- Root folders
image
- Max 2 expanded paths. Rest is `...`
image

- iPhone SE
image
- On Small screens long folder names are `...`
image

- 401 Error. Once we have un-secure end-points, we can put a login button here 
image

- Invalid path
image


src/components/data/listing/TableView.js
```diff
@@ -85,7 +86,7 @@ function TableView(props) {
 
     const tableId = build(baseId, ids.listingTable);
     const tableColumns = getTableColumns(isMedium, isLarge);
-
+    console.log("Error=>" + error + "<==");
```
ash3rz: Can probably remove this message?

src/constants.js
```diff
@@ -3,4 +3,6 @@
 export default {
     // The CyVerse page for DE. Not the DE base URL.
     CYVERSE_URL: "https://cyverse.org/discovery-environment",
+    COMMUNITY_DATA_PATH: "/iplant/home/shared",
+    SHARED_WITH_ME_PATH: "/iplant/home",
```
ash3rz: We can save this for another PR, but these values might be better suited coming from the external config file

src/components/data/listing/Listing.js
```diff
@@ -30,34 +38,81 @@ function Listing(props) {
...
-                        })),
-                        ...respData?.files.map((f) => ({ ...f, type: "FILE" })),
-                    ].map((i) => camelcaseit(i)), // camelcase the fields for each object, for consistency.
+            if (respData) {
+                const respRoots = respData.roots;
+                const home = respRoots.find(
+                    (root) =>
+                        root.label !== formatMessage(intl, "trash") &&
+                        root.label !== formatMessage(intl, "sharedWithMe") &&
+                        root.label !== formatMessage(intl, "communityData")
```
ash3rz: I don't think we need these `formatMessage` bits?  This is being checked against the response we're getting from the service which, I assume, will always be in English.  I may be missing something though.

src/components/data/listing/Listing.js
```diff
@@ -30,34 +38,81 @@ function Listing(props) {
...
-                    listing: [
-                        ...respData?.folders.map((f) => ({
-                            ...f,
-                            type: "FOLDER",
-                        })),
-                        ...respData?.files.map((f) => ({ ...f, type: "FILE" })),
-                    ].map((i) => camelcaseit(i)), // camelcase the fields for each object, for consistency.
+            if (respData) {
+                const respRoots = respData.roots;
+                const home = respRoots.find(
```
ash3rz: Couldn't we also grab the user's CyVerse ID from the context and find their home directory by finding the root whose label is their ID?  I'm just thinking if we ever add more roots for whatever reason, then this check would have to be updated to exclude those.

sri2k1us: Makes sense. I will refactor this logic.

src/components/data/listing/Listing.js
sri2k1us: Right. They will in english. May be move them to constants.

src/constants.js
sri2k1us: Yes. I have to file some issues against data-info. One of this is modifying response to the `roots` endpoint. I think lot hard coded checks in this PR can be fixed with some changes to `roots` endpoint

sri2k1us:
> LGTM :shipit:
> I just had some minor comments/suggestions.
> 
> Also, from the screenshots, it looks like the first part of the breadcrumb after the root dropdown is a bit far, like there's too much spacing/padding in between. It's the same for desktop and mobile.

The dropdown for roots is wide as the largest label. May be I can try to add a border to see if it makes it look better.

sri2k1us:
> > LGTM :shipit:
> > I just had some minor comments/suggestions.
> > Also, from the screenshots, it looks like the first part of the breadcrumb after the root dropdown is a bit far, like there's too much spacing/padding in between. It's the same for desktop and mobile.
> 
> The dropdown for roots is wide as the largest label. May be I can try to add a border to see if it makes it look better.

Welp...I couldn't do anything about the spacing 😩

src/components/data/DataNavigation.js
```diff
@@ -0,0 +1,460 @@
...
+    };
+    const routerPath = relativePathItems.reduce(reducer);
+    console.log(
+        "reduced path=>" +
+            routerPath +
+            " path index=>" +
+            selectedPathItemIndex +
+            " relative path=>" +
+            relativePathItems.length
+    );
```
psarando: Looks like this can be removed now?

src/components/data/DataNavigation.js
```diff
@@ -0,0 +1,460 @@
...
+ */
+function pathToRoute(root, relativePath, selectedPathItemIndex) {
+    const relativePathItems = getPathItems(relativePath);
+    const reducer = (accumulator, currentVal, idx) => {
+        if (idx <= selectedPathItemIndex) {
+            return accumulator + "/" + currentVal;
+        }
+        return accumulator;
+    };
+    const routerPath = relativePathItems.reduce(reducer);
```
psarando: This works, but I bet a combination of array `slice` and `join` might make it a little more concise.
```javascript
     const routerPath = [root, ...relativePath.slice(0, selectedPathItemIndex + 1)].join("/");
     return `/${NavigationConstants.DATA}?path=${routerPath}`;
```

src/components/data/DataNavigation.js
```diff
@@ -0,0 +1,460 @@
...
+    let relativePath = "";
+    if (path.indexOf(userHomePath) !== -1) {
+        relativePath = path.replace(userHomePath, "");
+    } else if (path.indexOf(userTrashPath) !== -1) {
+        relativePath = path.replace(userTrashPath, "");
+    } else if (path.indexOf(constants.COMMUNITY_DATA_PATH) !== -1) {
+        relativePath = path.replace(constants.COMMUNITY_DATA_PATH, "");
+    } else if (path.indexOf(constants.SHARED_WITH_ME_PATH) !== -1) {
+        relativePath = path.replace(constants.SHARED_WITH_ME_PATH, "");
+    }
```
psarando: It might be safer to use `path.startsWith` instead of `indexOf`.
And the string passed to `startsWith` should probably end with a `/` to be extra safe.

src/constants.js
psarando: These paths are already coming from the roots endpoint now?

In the `Listing` API call, these roots are detected based on their labels, but they're only being put into the `dataRoots` list. Could these paths also be cached the same as `userHomePath` and `userTrashPath`?

src/components/data/DataNavigation.js
```diff
@@ -0,0 +1,460 @@
...
+    const router = useRouter();
+    const [anchorEl, setAnchorEl] = useState(null);
+    const relativePath = getRelativePath(path, userHomePath, userTrashPath);
+    const pathItems = getPathItems(relativePath);
+    const [selectedIndex, setSelectedIndex] = useState(0);
+
+    useEffect(() => {
+        console.log(
+            "selected index=>" + pathItems.length > 0 ? pathItems.length - 1 : 0
+        );
```
psarando: More debug logging that can be removed?

src/components/data/DataNavigation.js
```diff
@@ -0,0 +1,460 @@
...
+                        aria-haspopup="true"
+                        aria-controls={formatMessage(
+                            intl,
+                            "selectedFolderAriaMenuItemControl"
+                        )}
+                        aria-label={pathItems[selectedIndex]}
+                        onClick={handleClickListItem}
+                        className={classes.selectedListItem}
+                    >
+                        <>
```
psarando: I think `ListItem` can already have multiple children, so this fragment wrapper may not be needed.

src/components/data/DataNavigation.js
```diff
@@ -0,0 +1,460 @@
...
+
+    const handleClose = () => {
+        setAnchorEl(null);
+    };
+
+    if (dataRoots.length === 0) {
+        return null;
+    }
+    return (
+        dataRoots.length > 0 && (
```
psarando: Since there's already a check above that returns `null`, this one seems unnecessary.

src/components/data/DataNavigation.js
```diff
@@ -0,0 +1,460 @@
...
+                        aria-haspopup="true"
+                        aria-controls={formatMessage(
+                            intl,
+                            "selectedDataRootAriaMenuItemControl"
+                        )}
+                        aria-label={dataRoots[selectedIndex].label}
+                        onClick={handleClickListItem}
+                        className={classes.selectedListItem}
+                    >
+                        <>
```
psarando: This fragment wrapper might not be needed.

src/components/data/listing/Listing.js
```diff
@@ -168,21 +225,34 @@ function Listing(props) {
         setPage(0);
     };
 
+    //route to default path
+    if (dataRoots.length > 0 && (!error || error.length === 0) && !path) {
+        router.push(
+            "/" + NavigationConstants.DATA + `?path=${dataRoots[0].path}`
```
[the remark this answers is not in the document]
sri2k1us: I had to look it but fragments are never sent to the server. See  https://github.com/zeit/next.js/issues/779

src/constants.js
sri2k1us: I think I  understand what you saying but let me try it out and see :P

sri2k1us:
> > > LGTM :shipit:
> > > I just had some minor comments/suggestions.
> > > Also, from the screenshots, it looks like the first part of the breadcrumb after the root dropdown is a bit far, like there's too much spacing/padding in between. It's the same for desktop and mobile.
> > 
> > 
> > The dropdown for roots is wide as the largest label. May be I can try to add a border to see if it makes it look better.
> 
> Welp...I couldn't do anything about the spacing 😩

I was able to fix it 😓

sri2k1us:
Thanks for reviews.

PR #63, opened 2020-02-28
CORE-868: switch to keycloak

slr71:
This PR switches Keycloak from CAS to Keycloak. The features that we needed are all readily available in the official `keycloak-connect` authentication module, so `passport` and its auth strategies have been removed as dependencies.


slr71:
I just did a rebase to ensure that these changes can be merged without any issues.

src/server/api/router.js
```diff
@@ -49,21 +49,12 @@ export default function apiRouter() {
...
-        if (req.user) {
-            res.json({
-                id: req.user.profile.id,
-                attributes: req.user.profile.attributes,
-            });
-        } else {
-            res.json(null);
-        }
-    });
+    api.get("/profile", (req, res) => res.json(auth.getUserProfile(req)));
```
psarando: 👍

src/server/index.js
```diff
@@ -55,46 +61,30 @@ app.prepare()
...
+        server.get("/login", keycloakClient.protect());
 
         logger.info("adding the /login/* handler");
-        server.get("/login/*", ensureLoggedIn("/login"), (req, res) => {
+        server.get("/login/*", keycloakClient.protect(), (req, res) => {
             res.redirect(req.url.replace(/^\/login/, ""));
         });
 
-        logger.info("adding the /logout handler");
-        server.get("/logout", (req, res) => {
```
psarando: So keycloak doesn't support logging out?

src/server/logging.js
```diff
@@ -15,23 +16,24 @@ const logger = createLogger({
     transports: [new transports.Console()],
 });
 
+const getLoggableUserID = (req) => getUserID(req) || "logged-out-user";
```
psarando: 👍

src/server/index.js
slr71: It does; the logout endpoint is built into the middleware, so we don't have to define it explicitly anymore.

slr71:
I'm running one more test against terrain in QA just to make sure it has a good chance of working. 🤞 If it works, I'll merge this PR.

PR #66, opened 2020-03-09
Dashboard

johnworth:
**Summary**

Initial version of a user-facing dashboard that works for both logged-in and logged-out users. Depends on the new dashboard-aggregator service in the backend, along with changes to Terrain to allow access to the new service.

The dashboard needs to show CyVerse events, news items, and public apps for logged-out users, while adding recent analyses and recently added apps for logged-in users. 

By default, all sections only display 8 items, though modifying the dashboard to display more is easy and making it user-configurable should be possible in the future.

The elevation of the app/nav bar was adjusted upwards so that the cards in the dashboard would actually appear to pass underneath it, rather than merge with it in some sort of blobby, eye-hurting, Lovecraftian horror of a rendering error. I believe this is the first page that scrolls, which is why no one else ran into this issue before.

Additionally, a new `PageSpacer` element was added to push all pages below the navigation bar without needing a top margin on every page. 

This pull request also contains some clean up in the server-side of Sonora, namely removing the large switch statements in the callApi() and terrain request handling code, along with moving the authentication code into a new file/module and making the objects configured there into singletons.

**Known Issues:**

The client-side of the dashboard uses the fetch() function provided by browsers (and polyfilled where not supported, IIRC) rather than axios because the new axios code broke the dashboard calls somehow and this was the easiest way to get it working again in short order. Someone with more experience with the axios code should probably take a look at it.

The `Open` link for analyses and apps doesn't work yet. The intention is to have them link to specific analyses and apps within Sonora, but I don't believe that functionality is in yet.

The sections have no user-facting configuration. The mockups show a button/link to display options, but I couldn't figure out what each section would need as configuration.

The limit on the number of items applies to all sections since they're grabbed at the same time and should have similar performance characteristics as a result.

**Screenshots**

Screen Shot 2020-04-10 at 9 18 49 AM

Screen Shot 2020-04-10 at 9 19 15 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,6 @@
     "is-plain-object": "^3.0.0",
     "js-yaml": "^3.13.1",
     "keycloak-connect": "^7.0.1",
-    "moment": "^2.24.0",
     "next": "^9.3.1",
     "numeral": "^2.0.6",
     "postman-request": "^2.88.1-postman.18",
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "is-plain-object": "^3.0.0",
     "js-yaml": "^3.13.1",
     "keycloak-connect": "^7.0.1",
+    "moment": "^2.24.0",
     "next": "^9.3.1",
     "numeral": "^2.0.6",
     "postman-request": "^2.88.1-postman.18",
--- a/src/common/callApi.js
+++ b/src/common/callApi.js
@@ -44,7 +44,9 @@ const callApi = (props) => {
         setError,
     } = props;
 
-    const requestOptions = {
+    let requestOptions = {
+        method: method,
+        url: endpoint,
         headers: headers
             ? headers
             : {
@@ -52,78 +54,24 @@ const callApi = (props) => {
               },
         credentials: "include",
     };
+
+    if (body) {
+        requestOptions.body = body;
+    }
+
     //TODO: Remove loading and error handling from this code once everything is ported to react-query
     setLoading && setLoading(true);
-    switch (method) {
-        case "GET":
-            return axiosInstance
-                .get(endpoint, requestOptions)
-                .then((resp) => checkForError(resp, props))
-                .then((apiResponse) => apiResponse.data)
-                .catch((error) => {
-                    setError && setError(error);
-                })
-                .finally(() => {
-                    setLoading && setLoading(false);
-                });
-        case "POST":
-            return axiosInstance
-                .post(endpoint, body, requestOptions)
-                .then((resp) => checkForError(resp, props))
-                .then((apiResponse) => apiResponse.data)
-                .catch((error) => {
-                    setError && setError(error);
-                })
-                .finally(() => {
-                    setLoading && setLoading(false);
-                });
-        case "PATCH":
-            return axiosInstance
-                .patch(endpoint, body, requestOptions)
-                .then((resp) => checkForError(resp, props))
-                .then((apiResponse) => apiResponse.data)
-                .catch((error) => {
-                    setError && setError(error);
-                })
-                .finally(() => {
-                    setLoading && setLoading(false);
-                });
-        case "PUT":
-            return axiosInstance
-                .put(endpoint, body, requestOptions)
-                .then((resp) => checkForError(resp, props))
-                .then((apiResponse) => apiResponse.data)
-                .catch((error) => {
-                    setError && setError(error);
-                })
-                .finally(() => {
-                    setLoading && setLoading(false);
-                });
-        case "DELETE":
-            return axiosInstance
-                .delete(endpoint, requestOptions)
-                .then((resp) => checkForError(resp, props))
-                .then((apiResponse) => apiResponse.data)
-                .catch((error) => {
-                    setError && setError(error);
-                })
-                .finally(() => {
-                    setLoading && setLoading(false);
-                });
-        case "HEAD":
-            return axiosInstance
-                .head(endpoint, requestOptions)
-                .then((resp) => checkForError(resp, props))
-                .then((apiResponse) => apiResponse.data)
-                .catch((error) => {
-                    setError && setError(error);
-                })
-                .finally(() => {
-                    setLoading && setLoading(false);
-                });
-        default:
-            throw Error("Unsupported method " + method);
-    }
+
+    return axiosInstance
+        .request(requestOptions)
+        .then((resp) => checkForError(resp, props))
+        .then((apiResponse) => apiResponse.data)
+        .catch((error) => {
+            setError && setError(error);
+        })
+        .finally(() => {
+            setLoading && setLoading(false);
+        });
 };
 
 export default callApi;
--- a/src/components/dashboard/index.js
+++ b/src/components/dashboard/index.js
@@ -68,24 +68,24 @@ const getOrigination = (kind, content) => {
 
     switch (kind) {
         case constants.KIND_ANALYSES:
-            origination = "Started by";
+            origination = getMessage("startedBy");
             date = content.start_date;
             break;
         case constants.KIND_APPS:
             if (content.integration_date) {
-                origination = "Integrated by";
+                origination = getMessage("integratedBy");
                 date = content.integration_date;
             } else {
-                origination = "Edited by";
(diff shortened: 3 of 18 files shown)


package.json
```diff
@@ -65,6 +65,7 @@
     "is-plain-object": "^3.0.0",
     "js-yaml": "^3.13.1",
     "keycloak-connect": "^7.0.1",
+    "moment": "^2.24.0",
```
sri2k1us, line 68: @johnworth - We already have date formatting lib in `ui-lib`. https://github.com/cyverse-de/ui-lib/blob/master/src/util/DateFormatter.js. We are using `date-fns` since `moment` mutation making it buggy.

src/components/dashboard/index.js
```diff
@@ -0,0 +1,329 @@
...
+        width: "100%",
+        height: 128, // This is needed to get the vertical scrolling to stop cutting off the bottom of the content.
+
+        [theme.breakpoints.down("sm")]: {
+            height: 32,
+        },
+    },
+    section: {
+        marginTop: 25,
+        marginBottom: 50,
```
sri2k1us, line 45: I will post just here but all `margin` and `spacing` should take value from `theme.spacing`.

src/components/dashboard/index.js
```diff
@@ -0,0 +1,329 @@
...
+            origination = getMessage("publishedBy");
+            date = content.date_added;
+            break;
+        default:
+            origination = getMessage("by");
+            date = new Date();
+    }
+
+    date = parse(date);
+    date = format(date, "MMMM Do YYYY, h:mm:ss a");
```
sri2k1us, line 115: This can done using the `formatDate` in `ui-lib`. But that can wait until https://github.com/cyverse-de/ui-lib/pull/33 is merged.

src/components/dashboard/index.js
```diff
@@ -0,0 +1,329 @@
...
+                });
+
+            if (!data) {
+                data = {};
+            }
+
+            setData(data);
+        };
+        fetcher();
+    }, []);
```
sri2k1us, line 271: hoping this will be a straightforward `react-query` + axios change. I can look into this after this PR is merged.

src/server/api/terrain.js
```diff
@@ -39,10 +39,14 @@ export const handler = ({ method, pathname, headers }) => {
             req
         )
             .then((apiResponse) => {
+                res.set(apiResponse.headers);
                 res.status(apiResponse.status);
                 res.send(apiResponse.data);
             })
-            .catch((e) => {
+            .catch(async (err) => {
+                console.log(err);
```
sri2k1us, line 47: this should `logger.error`.

src/server/api/terrain.js
```diff
@@ -95,40 +99,24 @@ export const call = (
...
-        case "DELETE":
-            return axiosInstance.delete(axiosUrl, requestOptions);
-
-        case "HEAD":
-            return axiosInstance.head(axiosUrl, requestOptions);
-
-        default:
-            throw Error("Unsupported method " + method);
-    }
+    return axiosInstance.request(requestOptions);
```
sri2k1us, line 121: 👍

src/server/index.js
```diff
@@ -1,12 +1,10 @@
 import express from "express";
-import keycloak from "keycloak-connect";
 import next from "next";
-import pgsimple from "connect-pg-simple";
-import session from "express-session";
 
 import * as config from "./configuration";
 import apiRouter from "./api/router";
 import NavigationConstants from "../common/NavigationConstants";
+import * as authn from "./auth";
```
sri2k1us, line 7: 👍

src/pages/_app.js
```diff
@@ -118,6 +119,7 @@ function MyApp({
                                 </Head>
                                 <ReactQueryDevtools initialIsOpen={false} />
                                 <Navigation activeView={pathname} />
+                                <PageSpacer />
```
sri2k1us, line 122: Since `PageSpacer` is not needed for smaller screens, we can use <Hidden smDown> <PageSpacer/><Hidden> here instead of setting height and width to 0 in `pageSpacer/index.js`.

src/components/dashboard/index.js
johnworth: This should be fixed now.

src/pages/_app.js
johnworth: I'm not sure what change you'd like me to make here.

src/server/api/terrain.js
johnworth: This should be fixed, thanks for the heads up.

src/pages/_app.js
sri2k1us: It was just a suggestion. The comment didn't get formatted properly. Comment updated now.

johnworth: Thanks for the correction.

src/common/callApi.js
```diff
@@ -44,86 +44,34 @@ const callApi = (props) => {
...
-                })
-                .finally(() => {
-                    setLoading && setLoading(false);
-                });
-        default:
-            throw Error("Unsupported method " + method);
-    }
+
+    return axiosInstance
+        .request(requestOptions)
```
ash3rz, line 66: 👍

src/components/uploads/UploadDropTarget.js
```diff
@@ -94,7 +94,7 @@ const UploadDropTarget = (props) => {
 };
 
 UploadDropTarget.propTypes = {
-    children: PropTypes.object.isRequired,
+    children: PropTypes.array.isRequired,
```
ash3rz, line 97: 👍

src/components/dashboard/index.js
```diff
@@ -0,0 +1,335 @@
...
+    );
+};
+
+const Dashboard = () => {
+    const classes = useStyles();
+    const [data, setData] = useState({});
+
+    useEffect(() => {
+        const fetcher = async () => {
+            let data = await fetch(
```
ash3rz, line 246: We've switched over to using react-query, not sure if you wanna tackle that in this PR or not.

src/components/dashboard/index.js
```diff
@@ -0,0 +1,335 @@
...
+        [constants.KIND_FEEDS, constants.SECTION_NEWS, getMessage("newsFeed")],
+        [
+            constants.KIND_FEEDS,
+            constants.SECTION_EVENTS,
+            getMessage("eventsFeed"),
+        ],
+    ];
+
+    return (
+        <div className={classes.gridRoot}>
```
ash3rz, line 309: Could we add a static ID here so that if nothing gets returned from the service, the tests still have some known ID to check that the dashboard came up?

Also, a [loading skeleton](https://material-ui.com/components/skeleton/) might be good since I believe this is the default page for Sonora.

src/components/dashboard/index.js
johnworth: I'm not going to tackle that now.

src/components/dashboard/index.js
johnworth: I've added a loading skeleton, along with switching over to react-query.

src/components/dashboard/index.js
johnworth: Sorry about saying I wasn't going to tackle that then, I'm feeling better today. Went ahead and tackled that, along with adding a loading skeleton.

src/components/dashboard/index.js
```diff
@@ -0,0 +1,344 @@
...
+            </div>
+        </div>
+    );
+};
+
+const DashboardSkeleton = () => {
+    const classes = useStyles();
+    const numSkellies = [0, 1, 2];
+    const skellies = numSkellies.map((n) => (
+        <div className={classes.section}>
```
ash3rz, line 248: One last thing since I just made a PR for this too - an ID here so the students can detect when loading begins and ends to ease testing.

johnworth: Okay, I wrapped the loading skeletons in a div and gave it a static ID. Let me know if I need to do more.

ash3rz: Nope, that's perfect - thanks!

PR #68, opened 2020-03-11
CORE-753 Enable sorting on all visible data columns, replace Path with Created date

ash3rz:
I only had sorting enabled on the Name column.  Now every column should be sortable.  

Since we got feedback that the Path column wasn't useful, I replaced it with Created. Still need to enable customizing the columns, but that will take some thought and be for another PR.

image

### DIFF
--- a/src/components/data/listing/TableView.js
+++ b/src/components/data/listing/TableView.js
@@ -53,7 +53,7 @@ function getTableColumns(isMedium, isLarge) {
     }
 
     if (isLarge) {
-        columns.push({ name: "Created", align: "left", enableSorting: true, key: "datecreated" });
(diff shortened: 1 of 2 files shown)


src/components/data/listing/TableView.js
```diff
@@ -42,18 +42,18 @@ function getTableColumns(isMedium, isLarge) {
...
+                enableSorting: true,
+                key: "datemodified",
             },
-            { name: "Size", align: "left", enableSorting: false }
         );
     }
 
     if (isLarge) {
-        columns.push({ name: "Path", align: "left", enableSorting: false });
+        columns.push({ name: "Created", align: "left", enableSorting: true, key: "datecreated" });
```
psarando: I think the current UI is labeling this column `Date Submitted`?

ash3rz:
Thanks for the review!

PR #69, opened 2020-03-16
CORE-940 User can customize data columns

ash3rz:
For now, I've made it the default that only the checkbox, name, and dot menu show in the table view.  I removed the breakpoints because I figure users can set whatever columns make sense for their screen size as needed.  In another PR, I can probably update the default columns to check for a value stored in local storage so the setting persists.

Desktop view:
image

Menu open:
image

Mobile view:
image

Scrolling on mobile:
customcolumns

### DIFF
--- a/src/__tests__/__snapshots__/data.js.snap
+++ b/src/__tests__/__snapshots__/data.js.snap
@@ -86,7 +86,7 @@ exports[`Data Table View renders 1`] = `
         <th
           aria-sort={null}
           className="MuiTableCell-root MuiTableCell-head MuiTableCell-alignCenter MuiTableCell-stickyHeader"
-          id="tableView.listingTable."
+          id="tableView.listingTable.checkbox"
           scope="col"
         >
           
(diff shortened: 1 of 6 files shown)


sri2k1us:
Code LGTM. I have few concerns / questions though:
- I am not sure about removing breakpoints. I agree users have the power to choose the columns that fits the screen size. What happens when users chose to view to more than what the screen can fit ? Will there be a horizontal scroll. A screenshot of this behavior will be good

- I agree with @psarando comment on using check mark for selection. Text shift looks odd.

ash3rz:
Thanks for the reviews!

I've updated the screenshots and added a gif to show that you can scroll horizontally.  I felt on the fence about getting rid of breakpoints too, so we can revisit if needed.  I started off leaving them, but then I wasn't sure what to do if a user, for example, added all the columns and then made their browser window smaller or if their browser already couldn't fit all the columns. It seemed confusing to start removing/hiding columns in that case especially if scroll works, so I removed the breakpoints.  I'm totally open to other ideas on this though.

PR #71, opened 2020-03-23
CORE-868: use constants to build the regular expression for URL paths…

slr71:
… that might appear in the browser address bar

This PR adds a function to generate the regular expression that is used to match URLs that might appear in a browser address bar. This approach should be a lot easier to maintain than the other approach. As long as a corresponding constant is added to `NavigationConstants.js`, it will appear in this regular expression. Metaprogramming for the win!

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -6,6 +6,7 @@ import session from "express-session";
 
 import * as config from "./configuration";
 import apiRouter from "./api/router";
+import NavigationConstants from "../components/layout/NavigationConstants";
 
 import logger, { errorLogger, requestLogger } from "./logging";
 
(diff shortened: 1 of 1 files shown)


slr71:
Thanks for the reviews!

src/server/index.js
```diff
@@ -14,6 +15,31 @@ export const app = next({
...
+            if (routeAlternation === "") {
+                routeAlternation = v;
+            } else {
+                routeAlternation = `${routeAlternation}|${NavigationConstants[k]}`;
+            }
+        }
+    }
+
+    // Build and compile the full regular expression.
+    return new RegExp(`^/(${routeAlternation})`);
```
psarando: I'm not sure I follow the logic for building `routeAlternation` here.

It looks like `escapeRegExp` is only going to be called for the very first value, otherwise the value is appended as-is to `routeAlternation`?

Otherwise, if `escapeRegExp` is supposed to called on every value, then I might have built an array, and called `.join("|")` on it to make the `routeAlternation` string.

slr71: Oops! That was an editing error. I meant to change the expression in the `else` clause as well. Thanks for catching that.

PR #73, opened 2020-03-24
Automatically open upload tracking dialog when an upload is started.

sri2k1us:
Right now, there is no feedback to user when file(s) are drag-n-drop into the page for upload. This change will automatically open the Upload tracking dialog when a new upload is started.

### DIFF
--- a/src/components/data/UploadMenuBtn.js
+++ b/src/components/data/UploadMenuBtn.js
@@ -5,7 +5,7 @@
  * when clicked
  */
 
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 
 import { build, getMessage } from "@cyverse-de/ui-lib";
 import { Button, Hidden, makeStyles, Menu, MenuItem } from "@material-ui/core";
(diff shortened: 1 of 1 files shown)


src/components/data/UploadMenuBtn.js
```diff
@@ -15,16 +15,24 @@ import UploadDialog from "../uploads/dialog";
...
 function UploadMenuBtn(props) {
     const { baseId } = props;
     const classes = useStyles();
+    const tracker = useUploadTrackingState();
 
     const [uploadAnchor, setUploadAnchor] = useState(null);
     const [uploadDialogOpen, setUploadDialogOpen] = useState(false);
 
+    useEffect(() => {
+        if (tracker.uploads.length > 0) {
```
ash3rz: I haven't looked too closely at the uploads code, but I believe this will open the upload dialog with every update to the tracker. So if I drag and drop in 10 items, the dialog will open once for the submission, and then I can close it to work on something else. But I think 10 times after that it will open for each file as they reach the completed state, which could be a bit much.

sri2k1us: Hmm good catch..It does open multiple times. I think it opens every time a new batch is started (batch size is 3). So I am seeing it open around 3 - 4 times. I figured out a different way to provide feedback to user. I will close this one for now.

PR #75, opened 2020-03-25
Upload tracking

sri2k1us:
- Display announcement when files are queued for upload.
- Note - Checks won't pass until https://github.com/cyverse-de/ui-lib/pull/31 is merged and deps are updated here.


src/components/data/UploadMenuBtn.js
```diff
@@ -5,26 +5,73 @@
...
+                    {formatMessage(intl, "uploadQueue")}
+                </Typography>
+            </Button>
+        );
+    }, [intl, theme.palette.primary.contrastText]);
+
+    useEffect(() => {
+        if (tracker.uploads.length > 0) {
+            announce({
+                text: `${tracker.uploads.length} ${formatMessage(
```
ash3rz: I think the DE snackbar component will accept anything, not just a string.  So this could be updated to:
```
text: getMessage("filesQueuedForUploadMsg", {values: {total: tracker.uploads.length}})

// messages.js
filesQueuedForUploadMsg: "{total} file(s) queued for upload"
```

And then in that case, we can get rid of `injectIntl`

sri2k1us: Hmm That doesn't seems to work. I can give it another shot later.

sri2k1us:
Thanks for the reviews.

src/components/data/UploadMenuBtn.js
psarando: Even if `getMessage` won't work, `formatMessage` will take an optional `value` param.

So `filesQueuedForUploadMsg` can still be "{total} file(s) queued for upload", then this `text` can be updated to `formatMessage(intl, "filesQueuedForUploadMsg", {total: tracker.uploads.length})`.

PR #76, opened 2020-03-26
Misc

sri2k1us:
- Fix margins around appBar IconButtons across different breakpoints
- set filesystem resource icon color to primary

### DIFF
--- a/src/components/layout/Navigation.js
+++ b/src/components/layout/Navigation.js
@@ -140,7 +140,9 @@ function Navigation(props) {
                                 id={build(ids.NAVIGATION, ids.DATA_BTN)}
                                 disableElevation
                                 className={
-                                    activeView === NavigationConstants.DATA
+                                    activeView.includes(
+                                        NavigationConstants.DATA
+                                    )
                                         ? classes.selectedDataButton
                                         : classes.dataButton
                                 }
--- a/src/components/layout/Navigation.js
+++ b/src/components/layout/Navigation.js
@@ -140,9 +140,7 @@ function Navigation(props) {
                                 id={build(ids.NAVIGATION, ids.DATA_BTN)}
                                 disableElevation
                                 className={
-                                    activeView.includes(
-                                        NavigationConstants.DATA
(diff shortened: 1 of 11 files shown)


sri2k1us:
> I had a question about one change, but LGTM :shipit:

I have a question about your question. What/ Where is your question?

src/components/layout/Navigation.js
```diff
@@ -140,7 +140,9 @@ function Navigation(props) {
                                 id={build(ids.NAVIGATION, ids.DATA_BTN)}
                                 disableElevation
                                 className={
-                                    activeView === NavigationConstants.DATA
+                                    activeView.includes(
```
ash3rz: It didn't save my comment!

I don't think any of the other navigation buttons are using an `includes` check.  Was there a reason for this change?

sri2k1us: Yes. Since I introduced dynamic routing, when the page navigate to a path for example `https://localhost:3000/data/ds/iplant/home/sriram`, the activeView is  `/data/iplant/home/sriram`. Earlier it used to just `data` .

sri2k1us: Eventually, when we introduce dynamic routing for apps, analyses etc...We will have to update the corresponding checks as well.

ash3rz: Ah, I didn't realize how that change affected this code. It sounds like we might need more checks though in case someone has a folder/path with `apps` in the name or an app has `data` in the name, etc.

ash3rz: Wow, something weird going on with GitHub comments!  I only just saw your second comment which is exactly what I said 😆

sri2k1us: Ok. I updated to logic to extract active view from the router path and restored existing logic in Navigation. Hopefully this workouts for all the cases.

psarando:
LGTM :shipit:

sri2k1us:
Thanks for the reviews.

PR #77, opened 2020-03-27
App listing

sri2k1us:
- Basic app listing using react-query.
- Fetch app categories
image

Coming up in next:
- Columns customization
- More ways to browse apps (like browse by app type, all public apps etc..)
- App details
- App Toolbar actions
- App Tile View.


src/components/apps/listing/Listing.js
```diff
@@ -0,0 +1,130 @@
...
+    const [order, setOrder] = useState("asc");
+    const [orderBy, setOrderBy] = useState("name");
+    const [selected, setSelected] = useState([]);
+    const [lastSelectIndex, setLastSelectIndex] = useState(-1);
+    const [page, setPage] = useState(0);
+    const [rowsPerPage, setRowsPerPage] = useState(25);
+    const { status, resolvedData: data, error } = usePaginatedQuery(
+        ["getApps", { rowsPerPage, orderBy, order, page }],
+        getApps
+    );
```
sri2k1us: We need to make a choice here ...With `usePaginatedQuery` when user switches between pages, there won't be a loading skeleton. Instead the page will continue to show the current data until new data is available. If we want loading skeleton, then we need to switch back to `useQuery`. I feel we need the loading skeleton otherwise the users are going be confused and will start to think nothing happened.

sri2k1us: I switched to using `useQuery`.

src/components/apps/listing/TableView.js
```diff
@@ -0,0 +1,240 @@
...
+                                        key={app.id}
+                                        id={rowId}
+                                        onClick={(event) =>
+                                            handleClick(event, appId, index)
+                                        }
+                                    >
+                                        <TableCell padding="checkbox">
+                                            <DECheckbox
+                                                checked={isSelected}
+                                                tabIndex={0}
```
ash3rz: I haven't tried this, but if this tabIndex is 0, even though the `TableRow` tabIndex is -1, can the user still reach this by using the keyboard?

src/components/apps/listing/TableView.js
```diff
@@ -0,0 +1,240 @@
...
+                                                rowId,
+                                                ids.integratorName
+                                            )}
+                                        >
+                                            {app.integrator_name}
+                                        </TableCell>
+                                        <TableCell
+                                            id={build(rowId, ids.RATING)}
+                                        >
+                                            <Rate
```
ash3rz: As an aside, you had mentioned before you left that you had some new ideas on what to do with app ratings but we never discussed them.

src/components/data/listing/TableView.js
```diff
@@ -33,7 +33,7 @@ import ids from "../ids";
 import messages from "../messages";
 import ResourceIcon from "./ResourceIcon";
 import SpanLink from "./SpanLink";
-import TableLoading from "./TableLoading";
+import TableLoading from "../../utils/TableLoading";
```
ash3rz: 👍

src/constants.js
```diff
@@ -6,4 +6,8 @@ export default {
     SHARED_WITH_ME: "Shared With Me",
     COMMUNITY_DATA: "Community Data",
     PATH_SEPARATOR: "/",
+    APPS_UNDER_DEV: "Apps under development",
+    FAV_APPS: "Favorite Apps",
+    MY_PUBLIC_APPS: "My public apps",
+    HPC: "High-Performance Computing",
```
ash3rz: Not necessary for this PR, but I wonder if we should standardize on calling stuff HPC vs Agave?

src/pages/apps.js
```diff
@@ -1,5 +1,6 @@
 import React from "react";
+import Listing from "../components/apps/listing/Listing";
 
 export default function Apps() {
-    return <div>Launch an app from here</div>;
+    return <Listing baseId="apps" />;
```
ash3rz: Not necessary for this PR, but at least for Data, I'm eventually going to have a `DataView` component that has everything, so things don't get added to the `Listing` component.  Not sure if you want to do the same.  Apps might be simpler and may not need it.

stories/apps/TableView.stories.js
```diff
@@ -0,0 +1,654 @@
...
+import TableView from "../../src/components/apps/listing/TableView";
+
+export const AppsTableViewTest = () => {
+    const logger = (message) => {
+        console.log(message);
+    };
+
+    const listing = {
+        total: 643,
+        apps: [
```
ash3rz: Just a suggestion, if these are the same apps as what's in the `Listing` story, one thing you can do to save space is move these to a `AppMocks` file.  I started doing that with mock data that I needed in multiple places, or even just to keep the story files a little smaller.

src/constants.js
sri2k1us: Correct. We also need to decide between `interactive` vs `VICE`

src/pages/apps.js
sri2k1us: Once we add few more UI elements in apps, I think the code will us more :)

src/components/apps/listing/TableView.js
sri2k1us: Don't know..I think the tabIndex was a copy-pasta..I will double check.

src/components/apps/listing/TableView.js
```diff
@@ -0,0 +1,250 @@
...
+                    order={order}
+                    orderBy={orderBy}
+                    baseId={tableId}
+                    columnData={columnData}
+                    onRequestSort={handleRequestSort}
+                    onSelectAllClick={handleSelectAllClick}
+                />
+                {loading && (
+                    <TableBody>
+                        <TableLoading numColumns={6} numRows={25} />
```
psarando, line 129: Should this also use `columnData.length`?
```suggestion
                        <TableLoading numColumns={columnData.length} numRows={25} />
```

stories/apps/Listing.stories.js
```diff
@@ -0,0 +1,655 @@
...
+import fetchMock from "fetch-mock";
+import { UploadTrackingProvider } from "../../src/contexts/uploadTracking";
+import Listing from "../../src/components/apps/listing/Listing";
+
+export default {
+    title: "Apps",
+};
+
+function ListingTest(props) {
+    const logger = (message) => {
```
psarando, line 11: eslint warning here:
> 'logger' is assigned a value but never used

src/components/apps/listing/Listing.js
```diff
@@ -0,0 +1,166 @@
...
+    const rangeSelect = (start, end, targetId) => {
+        const rangeIds = [];
+        for (let i = start; i <= end; i++) {
+            rangeIds.push(data?.apps[i].id);
+        }
+
+        let isTargetSelected = selected.includes(targetId);
+
+        isTargetSelected ? deselect(rangeIds) : select(rangeIds);
+    };
```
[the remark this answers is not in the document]
sri2k1us: Good point...I copy pasted all of selection logic from data listing since it pretty much the same logic for app listing.

stories/apps/Listing.stories.js
sri2k1us: Hmm I don't get lint warning from stories folder 🤔

sri2k1us:
Thanks for reviews. I address the PR comments in upcoming PRs since I have 2 branches on top of this one. Will save me some time in rebasing.

PR #78, opened 2020-03-28
Update data path routing and data view subcomponent organization

ash3rz:
I ended up making a few key changes:
* Move the last couple of queries in the Listing file out into functions in the endpoints folder
* Move DataNavigation out of the Header component and into the Listing component
* Move the data roots query from the Listing component into the DataNavigation component directly
* Update DataNavigation to not use router directly

Eventually I want to add a new component called DataView where everything (Upload, Breadcrumb, Header, Listing, Details Drawer) goes.  Right now, everything is getting added to the Listing component which I feel will bite us later on.

All of these changes should help us to 1) make the entire data view reusable (whether in a dialog or within a split 50/50 page) and 2) get the data view ready for using react-query.

### DIFF
--- a/src/components/data/DataNavigation.js
+++ b/src/components/data/DataNavigation.js
@@ -21,6 +21,10 @@ import {
     Tooltip,
     Typography,
 } from "@material-ui/core";
+import HomeIcon from "@material-ui/icons/Home";
+import FolderSharedIcon from "@material-ui/icons/FolderShared";
+import GroupIcon from "@material-ui/icons/Group";
+import DeleteIcon from "@material-ui/icons/Delete";
 import { makeStyles } from "@material-ui/core/styles";
 import { injectIntl } from "react-intl";
 import ArrowDropDownIcon from "@material-ui/icons/ArrowDropDown";
@@ -31,6 +35,8 @@ import NavigationConstants from "../../common/NavigationConstants";
 import ids from "./ids";
 import constants from "../../constants";
 import { getEncodedPath, getStorageIdFromPath } from "./utils";
+import { getFilesystemRoots } from "../endpoints/Filesystem";
+import { useUserProfile } from "../../contexts/userProfile";
 
 const useStyles = makeStyles((theme) => ({
     selectedListItem: {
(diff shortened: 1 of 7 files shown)


src/components/data/DataNavigation.js
```diff
@@ -488,11 +489,11 @@ function DataNavigation(props) {
             </Menu>
 
             <Hidden xsDown>
-                {path && (!error || error.length === 0) ? (
```
ash3rz: This error checking will come back after the switch to react-query.

sri2k1us:
> I ended up making a few key changes:
> 
> * Move the last couple of queries in the Listing file out into functions in the endpoints folder
> * Move DataNavigation out of the Header component and into the Listing component
> * Move the data roots query from the Listing component into the DataNavigation component directly
> * Update DataNavigation to not use router directly
> 
> Eventually I want to add a new component called DataView where everything (Upload, Breadcrumb, Header, Listing, Details Drawer) goes. Right now, everything is getting added to the Listing component which I feel will bite us later on.
> 
> All of these changes should help us to 1) make the entire data view reusable (whether in a dialog or within a split 50/50 page) and 2) get the data view ready for using react-query.

I agree with code refactor mostly. But when I reviewed the code it looked suspicious in few places. So I checked out your branch and ran it locally. I found few bugs. Attached is a screenshot

image

src/components/data/DataNavigation.js
```diff
@@ -168,15 +142,14 @@ function FolderSelectorMenu({
...
-        const storageId = getStorageIdFromPath(router.pathname);
         const relativePath = getRelativePath(
             path,
             userHomePath,
             userTrashPath,
             sharedWithMePath,
             communityDataPath
         );
-        router.push(pathToRoute(storageId, root, relativePath, index));
+        handlePathChange(`${relativePath}${constants.PATH_SEPARATOR}${index}`);
```
sri2k1us: 🤔 Not sure if this will work. It will eval to something like `/analyses/1`

src/components/data/DataNavigation.js
```diff
@@ -290,8 +262,7 @@ function BreadCrumb({
         event.preventDefault();
         const relativePathItems = getPathItems(relativePath);
         const index = relativePathItems.indexOf(crumb);
-        const storageId = getStorageIdFromPath(router.pathname);
-        router.push(pathToRoute(storageId, root, relativePath, index));
+        handlePathChange(`${relativePath}${constants.PATH_SEPARATOR}${index}`);
```
sri2k1us: Same as above?

src/components/data/utils.js
```diff
@@ -21,14 +21,23 @@ function getEncodedPath(path) {
...
         return "";
     }
-    return pathname.split(constants.PATH_SEPARATOR)[2];
+
+    const paths = pathname.split(constants.PATH_SEPARATOR);
+    if (paths.length > 2) {
+        return paths[2];
+    } else {
+        // Default to data store.  Needed if user navigates to just /data.
+        return "ds";
```
sri2k1us: If user lands on `/data`, the corresponding `index.js` file will forward the request to `listing.js` component with empty path which in turn will forward the request to `/data/ds/<user home path>. 

`useEffect(() => {
        //route to default path
        if (dataRoots.length > 0 && !path) {
            handlePathChange(dataRoots[0].path);
        }
    }, [dataRoots, handlePathChange, path]);`
 

So router path will already have the storage id.

Anyway I do think it is a good idea to have `ds` as default storage id. But just want to make sure we are all on the same page when it comes to how the code works.

src/components/data/DataNavigation.js
ash3rz: Good catch!  I'm not sure how this was working on my machine before I made the PR, but I'll double check.  Either way, this logic doesn't look right 👍

src/components/data/utils.js
ash3rz: I made these changes super late on a Friday so there are probably bugs 😄 Thanks for reviewing

When I navigated to `/data` without this change, I ended going to `/data/undefined/iplant/home/aramsey` I believe.  I'll double all the handler calls to see what's needed and what's not.

And unfortunately no, this was not intended to fix the reloading, just intended to get the data view to be reusable.  I happened to notice the reloading while I was making these changes so I made the comment in the UI channel.  If needed, I can look into the reloading issue after this PR.

src/components/data/DataNavigation.js
ash3rz: These should be fixed now.  I still need to double check the changes in the `utils` about defaulting to `ds` though, will tackle tomorrow.

src/components/data/utils.js
ash3rz: OK, I just ran through the code again. The default in utils is needed because if the user goes to `/data` there is no storage ID.  

The Listing fetches the roots, then once the roots come back, Listing calls `handlePathChange` with the root home path "/iplant/home/userId". Then the page tries to get the storage ID using `router.pathname` which is still just `/data` at that point.  So we end up with `/data/undefined/iplant/home/userId`. Adding the default `ds` in the `utils` file though, updates the path to `/data/ds/iplant/home/userId`, and from then on `router.pathname` will find a storage ID.

You might be thinking of this bit of code that used to be in the Listing, but has been removed to get routing out of our components.  This would default to adding `ds` in the path: https://github.com/cyverse-de/sonora/blob/bf4649bd8930437be0b337bb462a9f845553e21f/src/components/data/listing/Listing.js#L250-L256

sri2k1us: Oh right I was purely commenting  based on how I had coded. Sorry about the confusion.

src/components/data/DataNavigation.js
```diff
@@ -91,11 +92,7 @@ function pathToRoute(storageId, root, relativePath, selectedPathItemIndex) {
         return accumulator;
     };
     const routerPath = relativePathItems.reduce(reducer);
-    return `${constants.PATH_SEPARATOR}${NavigationConstants.DATA}${
-        constants.PATH_SEPARATOR
-    }${storageId}${root}${constants.PATH_SEPARATOR}${getEncodedPath(
-        routerPath
-    )}`;
+    return `${root}${constants.PATH_SEPARATOR}${routerPath}`;
```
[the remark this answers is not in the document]
ash3rz: I think this still works because now all routing/URL updates go through the `handlePathChange` method in the data pages which will always encode the path.  So if I also encode it here and test with a path with `&` in it, I end up with a double encoding: `http://localhost:3000/data/ds/iplant/home/aramsey/te%2526st`

ash3rz:
Thanks for the reviews!

PR #80, opened 2020-04-01
App listing2

sri2k1us:
- Add Type Filter
- Add `Browse All Apps`
- Add 302 redirection logic for HPC Auth
- switch to `axios` for Terrain API calls.

Note about `axios` switch:
I had to switch to `axios` cos `node-fetch 2.6.0` wouldn't return `Location` headers for `302` redirects and `node-fetch 3.x beta` worked well for `302` redirect but returns incorrect results by ignoring query params.
      
image

- HPC auth redirect

image

- Note this PR is on top https://github.com/cyverse-de/sonora/pull/77. So reviews can start from https://github.com/cyverse-de/sonora/pull/80/commits/768cc5a9872f81591dea14625a35b843fe5e908a


src/server/api/terrain.js
```diff
@@ -106,5 +116,7 @@ export const call = (
         };
     }
 
-    return fetch(apiURL, requestOptions);
+    requestOptions.url = apiURL.toString();
+
+    return axios(requestOptions);
```
psarando, line 121: Will changing this from `fetch` to `axios` break the `fetchMock` calls in the stories?

src/components/apps/AppNavigation.js
```diff
@@ -161,7 +203,46 @@ function AppNavigation(props) {
...
+                        intl,
+                        "categoriesMenuItemAriaControl"
+                    )}
+                    aria-label={allAppsCategory.name}
+                >
+                    <ListItemIcon>{allAppsCategory.icon}</ListItemIcon>
+                    <ListItemText>{allAppsCategory.name}</ListItemText>
+                </ListItem>
             </Menu>
+            <div className={classes.divider} />
```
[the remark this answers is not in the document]
sri2k1us: The `Divider` component places vertical or horizontal line. I think this is just pushing off all the elements to the right.

src/server/api/terrain.js
sri2k1us: Good catch! Test seems to be broken but I think I found a `axios` mock lib - https://www.npmjs.com/package/axios-mock-adapter. I will update the tests.

package.json
```diff
@@ -66,7 +67,7 @@
     "js-yaml": "^3.13.1",
     "keycloak-connect": "^7.0.1",
     "next": "^9.3.1",
-    "node-fetch": "^2.6.0",
+    "node-fetch": "^3.0.0-beta.4",
```
ash3rz, line 70: Are we completely switched over at this point to axios?  Do we still need this dependency?

src/components/apps/AgaveAuthPromptDialog.js
```diff
@@ -0,0 +1,70 @@
...
+        window.location.replace(location);
+        handleClose();
+    };
+    return (
+        <Dialog
+            open={open}
+            onClose={handleClose}
+            id={build(baseId, ids.AGAVE_AUTH_PROMPT_DIALOG)}
+            fullScreen={fullScreen}
+            aria-labelledby={formatMessage(intl, "agaveRedirectTitle")}
```
ash3rz, line 35: `aria-labelledby` should contain an element ID, I believe.  This looks to be a shared message for the dialog title.  So in your case, you just have to add an ID to the `DialogTitle` component and add that same ID here in the `aria-labelledby`.

package.json
sri2k1us: I will double check uploads one last time and then I can pull the plug on `node-fetch`

sri2k1us:
Looks like we still need `node-fetch` at least until we port `fetch` call in `CyverseAppBar` which I already did in AppDetails PR #83 .

Thanks for the reviews.

PR #81, opened 2020-04-01
WIP: Add preferences page

jack-mitt:
Hello all,

Here is my work so far for the preferences page, sorry it took so long there were a few speed bumps. I am still working on IDs and internationalization, but I wanted to get the code out there to start getting some feedback on other things to improve. There currently is no functionality for the file browser and the default path is temporary. We are waiting on the functionality for the BROWSE button and I was not sure how to get the default analyses path (maybe someone can shed some light). Here are a few screenshots: 
image

image

image


src/components/preferences/PreferencesPage.js
```diff
@@ -0,0 +1,87 @@
...
+        setFieldValue("preferences.enableWaitTimeMessage", true);
+        setFieldValue("preferences.enableAnalysisEmailNotification", true);
+        setFieldValue("preferences.enableHPCPrompt", true);
+        setFieldValue("preferences.notificationKBShortcut", "N");
+        setFieldValue("preferences.dataKBShortcut", "D");
+        setFieldValue("preferences.closeKBShortcut", "Q");
+        setFieldValue("preferences.appsKBShortcut", "A");
+        setFieldValue("preferences.analysisKBShortcut", "Y");
+        // TODO find a way to get users default output folder
+        //setFieldValue("preferences.default_output_folder", )
```
ash3rz, line 35: Haven't yet reviewed the rest but wanted to quickly comment on this - I totally forgot to update you after the config file was updated 😖 Sorry

So the config file that gets mounted in has been updated, you can see the QA config file [here](https://gitlab.cyverse.org/core-sw/k8s-resources/blob/master/resources/configs/qa/sonora.yaml).  You'll see it's a yaml file and the value you need is in `irods.home_path`.

So, now you need to update the `configuration.js` file in Sonora to parse that yaml file and your new irods home path value.  You'll see we're first validating that the values exist: https://github.com/cyverse-de/sonora/blob/596193a16d0ff2ef01fd0c1683409dfd5392d216/src/server/configuration.js#L83-L87

And then, after validated, we're exporting the values from the yaml file out of this configuration file for usage: https://github.com/cyverse-de/sonora/blob/596193a16d0ff2ef01fd0c1683409dfd5392d216/src/server/configuration.js#L202-L220

You just need to do the same thing for your `irods.home_path` value.

Once you have that change, then you can use the exported value from `configuration.js` here in your preferences component to build the user's home path. Part of the user's home path is their cyverse ID, which you can get using the `useUserProfile` hook that we built into Sonora.  [DataNavigation](https://github.com/cyverse-de/sonora/blob/596193a16d0ff2ef01fd0c1683409dfd5392d216/src/components/data/DataNavigation.js#L353) has a quick example.

src/components/preferences/PreferencesPage.js
```diff
@@ -0,0 +1,87 @@
...
+        setRestoreDef(false);
+    };
+
+    return (
+        <div className={classes.root}>
+            <Formik initialValues={config} onSubmit={handleSubmit}>
```
ash3rz, line 41: This is not necessary, just something to note - since we're not putting webhooks on this page anymore, you could actually set initialValues to `config.preferences` since everything you need is in the `preferences` object.  And then for the rest of your component, everywhere you have `name="preferences.*"`, you could remove `preferences.`.  Again, not necessary, what you've got completely works fine, just thought I'd mention it.

src/components/preferences/PreferencesTab.js
```diff
@@ -0,0 +1,152 @@
...
+        <div className={classes.preferences}>
+            <Typography className={classes.sectionHeader}>General</Typography>
+            <br />
+            <Typography className={classes.section} display="inline">
+                Remember last file path for Apps
+            </Typography>
+            <Field
+                component={FormSwitch}
+                name="preferences.rememberLastPath"
+                value="rememberLastPath"
```
ash3rz, line 49: I might be remembering wrong, but I don't think you need these `value` props on your Fields?  The value for each Field should automatically be coming from your `initialValues` and mapped by the `name` prop.

src/components/preferences/ShortcutsTab.js
```diff
@@ -0,0 +1,91 @@
...
+
+export default function ShortcutsTab() {
+    const classes = useStyles();
+
+    const onTextChange = (setFieldValue, fieldName) => (event, value) => {
+        const new_value = event.target.value;
+        setFieldValue(fieldName, new_value);
+    };
+
+    const FormTextField = ({
```
ash3rz, line 18: There's already this exact component in `cyverse-de/ui-lib`.  I'd switch over to using that version of `FormTextField` instead.

src/components/preferences/styles.js
```diff
@@ -0,0 +1,70 @@
...
+    dropDown: {
+        width: "20%",
+        marginLeft: "3%",
+        marginRight: "3%",
+    },
+
+    actionButton: {
+        fontSize: 16,
+        marginTop: 5,
+        marginLeft: 5,
```
ash3rz, line 40: Just picking a random spot, instead of setting these margins to specific numbers, I would use the [theme's spacing](https://material-ui.com/customization/spacing/) instead.  So lots of `marginTop: theme.spacing(1)`, for example.

src/components/preferences/PreferencesPage.js
```diff
@@ -0,0 +1,87 @@
...
+                            onCancelBtnClick={() => setRestoreDef(false)}
+                            heading="Restore Defaults"
+                            messages="messaging"
+                            message={
+                                <LoadingMask loading={false}>
+                                    <div>
+                                        Are you sure you would like to restore
+                                        defaults?
+                                    </div>
+                                </LoadingMask>
```
ash3rz, line 79: I mentioned this on our call today in regards to something else, but we're kind of shifting away from the loading spinners to the [skeletons](https://material-ui.com/components/skeleton/). I don't think it's necessarily bad to have a spinner on a dialog though.  The alternative would be to close the dialog and apply the skeleton to the preferences page while the preferences are saving.  I don't feel strongly one way or the other, just noting it here.

src/components/preferences/ShortcutsTab.js
psarando: Yes, use the `FormTextField` from `ui-lib`. That will also mean the `onTextChange` function is no longer needed here, because `Formik` already handles this with its `onChange` function that is part of the `field`, which is spread into the `FormTextField` in `ui-lib`.

The `FormTextField` in `ui-lib` also supports a `label` prop, so these separate `Typography` labels will no longer be needed either.

However, the `inputProps={{ maxLength: 1 }}` prop will need to be added to each `Field` below (that's what that `... custom` prop is for in the `FormTextField` in `ui-lib`).

sri2k1us:
Closing this PR. Thanks @jack-mitt.

psarando:
FYI @jack-mitt your work on this PR was integrated into #122 🎉

PR #84, opened 2020-04-08
Pagination

sri2k1us:
image

image

image

After Adjusting the Grid:
image

image

image

image

### DIFF
--- a/src/components/utils/DEPagination.js
+++ b/src/components/utils/DEPagination.js
@@ -6,7 +6,6 @@ import {
     ClickAwayListener,
     Grid,
     Grow,
-    Hidden,
     MenuItem,
     MenuList,
     Paper,
@@ -23,14 +22,16 @@ import intlData from "./messages";
 const useStyles = makeStyles((theme) => ({
     paper: {
         flexGrow: 1,
-        marinTop: theme.spacing(1),
+        marginTop: theme.spacing(1),
+        marginBottom: theme.spacing(1),
     },
     paginationItems: {
         paddingTop: theme.spacing(1),
+        paddingBottom: theme.spacing(1),
         color: theme.palette.info,
     },
     buttonPadding: {
-        paddingTop: theme.spacing(1),
+        paddingTop: theme.spacing(1.5),
     },
 }));
 
@@ -136,10 +137,8 @@ function DEPagination(props) {
     const matches = useMediaQuery(theme.breakpoints.down("sm"));
     return (
         <Paper className={classes.paper}>
-            <Grid container spacing={1}>
-                <Hidden mdDown>
-                    <Grid item xs={1} sm={4}></Grid>
-                </Hidden>
+            <Grid container>
+                <Grid item sm={4} md={5}></Grid>
                 <Grid item>
                     <Pagination
                         id={build(baseId, ids.PAGINATION_TOOLBAR)}
@@ -151,7 +150,7 @@ function DEPagination(props) {
                         page={page}
                     />
                 </Grid>
-                <Grid item xs={1} sm={4}>
+                <Grid item>
                     <ItemsPerPage
                         onPageSizeChange={onPageSizeChange}
                         selectedPageSize={pageSize}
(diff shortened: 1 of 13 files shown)


src/components/utils/DEPagination.js
```diff
@@ -0,0 +1,167 @@
...
+        pageSize,
+        baseId,
+        intl,
+    } = props;
+    const matches = useMediaQuery(theme.breakpoints.down("sm"));
+    return (
+        <Paper className={classes.paper}>
+            <Grid container spacing={1}>
+                <Hidden mdDown>
+                    <Grid item xs={1} sm={4}></Grid>
```
ash3rz, line 141: Is this supposed to be an empty Grid item?

src/components/utils/DEPagination.js
```diff
@@ -0,0 +1,167 @@
...
+import ArrowDropDownIcon from "@material-ui/icons/ArrowDropDown";
+import { injectIntl } from "react-intl";
+import { build, formatMessage, withI18N } from "@cyverse-de/ui-lib";
+import ids from "./ids";
+import intlData from "./messages";
+
+const useStyles = makeStyles((theme) => ({
+    paper: {
+        flexGrow: 1,
+        marinTop: theme.spacing(1),
```
ash3rz, line 26: Tpyo?

src/components/utils/DEPagination.js
```diff
@@ -0,0 +1,167 @@
...
+} from "@material-ui/core";
+import ArrowDropDownIcon from "@material-ui/icons/ArrowDropDown";
+import { injectIntl } from "react-intl";
+import { build, formatMessage, withI18N } from "@cyverse-de/ui-lib";
+import ids from "./ids";
+import intlData from "./messages";
+
+const useStyles = makeStyles((theme) => ({
+    paper: {
+        flexGrow: 1,
```
ash3rz, line 25: Is there something else we can do to center the pagination? Maybe this is a spot where [Container](https://material-ui.com/components/container/) might be useful?  At least in the screenshots it looks like it's always in the same spot, so it looks more centered if you have a lot of pages, but looks offset if you only have a few pages. And then on mobile, it's all the way to the left with a gap on the right.

src/components/utils/DEPagination.js
sri2k1us: Yes. thats what is pushing the `Pagination` component closer to center.

src/components/utils/DEPagination.js
sri2k1us: There is not much choice when it comes mobile. We have to align left otherwise even with couple of pages, it will overflow into 2nd line. But I will try `Container` and `Box` to see if it better suited for aligning stuff to the center in large screens.

src/components/utils/DEPagination.js
```diff
@@ -0,0 +1,166 @@
...
+        baseId,
+        intl,
+    } = props;
+    const matches = useMediaQuery(theme.breakpoints.down("sm"));
+    return (
+        <Paper className={classes.paper}>
+            <Grid container>
+                <Grid item sm={4} md={5}></Grid>
+                <Grid item>
+                    <Pagination
```
psarando, line 143: Not for this PR, but eventually it would be nice to include a number field so the user can manually enter their own page number (it doesn't look like this MUI-labs component supports that out-of-the-box).

Maybe only if the number of pages exceeds a certain threshold?

sri2k1us: We can definitely can add that large screens at least.

src/components/utils/DEPagination.js
```diff
@@ -0,0 +1,166 @@
...
+                    aria-controls={open ? menuId : undefined}
+                    aria-expanded={open ? "true" : undefined}
+                    aria-label={formatMessage(intl, "selectPageSize")}
+                    aria-haspopup="menu"
+                    onClick={handleToggle}
+                >
+                    {selectedPageSize} <ArrowDropDownIcon />
+                </Button>
+            </Tooltip>
+            <Popper
```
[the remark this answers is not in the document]
sri2k1us: Inspired by https://material-ui.com/components/button-group/#split-button

sri2k1us:
Thanks for the reviews.

PR #85, opened 2020-04-10
Add patch method to callApi

ash3rz:
I just ran into this while switching stuff to react-query


ash3rz:
Thanks for the review!

PR #89, opened 2020-04-14
Add static IDs to loading skeletons

ash3rz:


ash3rz:
Thanks for the review!

PR #90, opened 2020-04-15
Data view dynamic routing fix

ash3rz:
I did two things in this PR:
* The first commit has the least amount of changes needed to get the dynamic routing to work.  The main thing was specifying both the dynamic URL or `href` and also the `as` similar to [here](https://nextjs.org/docs/api-reference/next/link#dynamic-routes) but for `next/router` instead of `next/link`. I also simplified things so that we don't have to maintain roughly the same logic in 3 different files (`index`, `/ds/index/`, and `/ds/[...pathItems]`)
* The second commit, after understanding how the routing works a little better, I simplified the logic for figuring out the entire URL path (as in `/data/ds/pathToFile`).  Instead of having logic to figure out the storage ID and then build the full URL, we're making sure no matter how the user gets to the listing, that the pathing in nextjs is complete and just using that.

### DIFF
--- a/src/components/data/utils.js
+++ b/src/components/data/utils.js
@@ -18,26 +18,4 @@ function getEncodedPath(path) {
     return encodedPath;
 }
 
-/**
- *  Get storage id from router path
- *
- *  Sample path: /data/ds/iplant/home/ipcdev
- *
- * @param {string} pathname - path from the router
- * @returns {string|*} storage id
- */
-function getStorageIdFromPath(pathname) {
-    if (!pathname) {
-        return "";
-    }
-
-    const paths = pathname.split(constants.PATH_SEPARATOR);
-    if (paths.length > 2) {
-        return paths[2];
-    } else {
-        // Default to data store.  Needed if user navigates to just /data.
-        return "ds";
-    }
-}
-
-export { getEncodedPath, getStorageIdFromPath };
+export { getEncodedPath };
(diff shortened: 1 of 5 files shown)


sri2k1us:
LGTM and the page does look stable and just refreshes which is great. But I think I may have found a bug when navigating a folder with spl. chars. See screen shot: 
image

The error I got from service was `{"error_code":"ERR_DOES_NOT_EXIST","path":"/iplant/home/sriram/+!*$)(><:\"~.#/?="}`. The folder I clicked was `/iplant/home/sriram/+!*$)(><:"~.#\?=`. So somehow, there is an extra `\`

ash3rz:
> LGTM and the page does look stable, just refreshing the table which is great 🎉. But I think I may have found a bug when navigating a folder with spl. chars. See screen shot:
> image
> 
> The error I got from service was `{"error_code":"ERR_DOES_NOT_EXIST","path":"/iplant/home/sriram/+!*$)(><:\"~.#/?="}`. The folder I clicked was `/iplant/home/sriram/+!*$)(><:"~.#\?=`. Also see how the breadcrumbs got messed up because of extra `/` before the `?`.

Good catch!  I just pushed a fix

ash3rz:
Thanks for the review!  QA informed me they're having some issues testing which they suspect is due to the hard refresh.  I'll go ahead and merge this, but reviews still welcome!

PR #92, opened 2020-04-17
Move dashboard services to axios.

sri2k1us:


src/server/api/router.js
```diff
@@ -30,6 +30,9 @@ export default function apiRouter() {
         terrainHandler({
             method: "GET",
             pathname: "/dashboard",
+            headers: {
+                "Content-Type": "application/json",
+            },
```
ash3rz, line 35: Is this actually needed for a GET request?

sri2k1us: @johnworth had it in the fetch-request. So I copied it here.

sri2k1us:
> Would it be possible to rename serviceFacade/dashboardServiceFacade.js to just serviceFacade/dashboard?

Done!

sri2k1us:
Thanks for the reviews.

src/server/api/router.js
psarando: I think our Swaggerized terrain endpoints use this header to decide what format to use in the response. So this content-type tells terrain we want a JSON response, but requesting `application/x-yaml` content-type will return a YAML response.

PR #93, opened 2020-04-17
Rename serviceFacade files to remove stutter from imports

johnworth:
Renames the serviceFacade directory to serviceFacades to match the plurality of the other directories at the same level.

Renames the individual modules containing the facades so their names no longer contain the phrase serviceFacade. It caused imports to stutter.


sri2k1us:
LGTM :shipit:

PR #96, opened 2020-04-21
Migrate data navigation roots to context

ash3rz:
This will allow the data roots query to be cached and prevent the DataNavigation component from rendering and not rendering between page changes between /data, /data/ds, and /data/ds/[...pathItems].

### DIFF
--- a/src/components/data/DataNavigation.js
+++ b/src/components/data/DataNavigation.js
@@ -41,6 +41,7 @@ import constants from "../../constants";
 import { getFilesystemRoots } from "../../serviceFacades/filesystem";
 import { useUserProfile } from "../../contexts/userProfile";
 import { useQuery } from "react-query";
+import { useDataRoots } from "../../contexts/dataRoots";
 
 const useStyles = makeStyles((theme) => ({
     selectedListItem: {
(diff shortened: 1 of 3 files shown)


sri2k1us:
I read up little bit more of on react-context before reviewing this code. I feel, while this might fix the issue, IMO contexts are not the right way to do it. My understanding is that they are meant for global state management or when certain data needs to passed down as props to several components deep. Here only Data Navigation will need these roots. 

I managed to fix this issue with caching in `react-query` with a minor change. Please see https://github.com/cyverse-de/sonora/pull/98. Hopefully, we can use the same model to cache app categories, reference genomes, info-types  and many more things that won't change often.

Let me know what you think.

ash3rz:
> I read up little bit more of on react-context before reviewing this code. I feel, while this might fix the issue, IMO contexts are not the right way to do it. My understanding is that they are meant for global state management or when certain data needs to passed down as props to several components deep. Here only Data Navigation will need these roots.
> 
> I managed to fix this issue with caching in `react-query` with a minor change. Please see #98. Hopefully, we can use the same model to cache app categories, reference genomes, info-types and many more things that won't change often.
> 
> Let me know what you think.

I agree and I was already using the `Infinity` caching on info-types, I'm not sure why I didn't think to use it here. I'll go ahead and close this PR.

PR #98, opened 2020-04-23
Data roots caching

sri2k1us:
- Fix error that occurs when loading `Data` view.
- Cache data roots using `react-query`. This will fix issue where the `roots` wont load when switching pages.
- Cache App categories.


src/components/data/DataNavigation.js
```diff
@@ -367,44 +396,19 @@ function DataNavigation(props) {
...
+            onSuccess: preProcessData,
             onError: () => {
                 //temporary workaround -> should still display community data
                 announce({
                     text: formatMessage(intl, "dataSignIn"),
                     variant: AnnouncerConstants.INFO,
                 });
             },
+            staleTime: Infinity,
+            cacheTime: Infinity,
```
sri2k1us, line 411: This setting will make `react-query` cache the roots. If you open, the react-query  dev tools. You can see this in action.

src/components/data/DataNavigation.js
```diff
@@ -461,8 +465,11 @@ function DataNavigation(props) {
     };
 
     if (dataRoots.length === 0) {
+        const cacheRoots = queryCache.getQueryData("dataFileSystemRoots");
+        preProcessData(cacheRoots);
```
sri2k1us, line 469: If `dataRoots` is empty, simply get it from cache.

src/components/data/listing/Listing.js
```diff
@@ -92,7 +85,18 @@ function Listing(props) {
 
     useEffect(() => {
         setSelected([]);
-    }, [path]);
+        if (path) {
```
sri2k1us, line 88: when landing on `/data` or `/data/ds`, the `path` is null, which resulted in the error.

sri2k1us: I also found another bug where when you logged out, and land on `/data `or `/data/ds`, the path sent to data-info get appended with key-cloak query params. I haven't fixed this issue yet.

src/components/apps/AppNavigation.js
```diff
@@ -126,15 +133,40 @@ function AppNavigation(props) {
...
+    const { isFetching } = useQuery({
         queryKey: "getPrivateCategories",
         queryFn: getPrivateCategories,
-        config: { onSuccess: preProcessData, refetchOnWindowFocus: false },
+        config: {
+            onSuccess: preProcessData,
+            onError: () => {
+                //temporary workaround -> should still display community data
+                announce({
+                    text: "Unable to fetch categories",
```
ash3rz, line 146: This should be i18n

src/components/apps/AppNavigation.js
```diff
@@ -126,15 +133,40 @@ function AppNavigation(props) {
...
+                });
+            },
+            staleTime: Infinity,
+            cacheTime: Infinity,
+        },
     });
 
+    useEffect(() => {
+        setCategoryStatus(isFetching);
+    }, [isFetching, setCategoryStatus]);
```
ash3rz, line 157: Seems like you could modify `const {isFetching: categoryStatus} = useQuery...` instead of adding this `useEffect`

sri2k1us:
Thanks for review.

PR #99, opened 2020-04-23
Upload, Delete, and Create Folder

ash3rz:
I started off having a `+ Create` button on the header which was a menu with Folder and File as menu items, but then decided to go with just creating a create folder button directly in the header for now.  I can switch back though if people think that'll be better, I still have the code.

Desktop:
image

Mobile:
image

Folder name dialog on mobile:
image


src/components/data/utils.js
```diff
@@ -18,4 +21,31 @@ function getEncodedPath(path) {
     return encodedPath;
 }
 
-export { getEncodedPath };
+const validateDiskResourceName = (name) => {
```
ash3rz, line 24: I stole this from @psarando App Launch code.  If App Launch gets merged first, I'll refactor to use that

src/components/data/listing/Listing.js
```diff
@@ -90,6 +94,15 @@ function Listing(props) {
         },
     });
 
+    const refreshListing = () => queryCache.refetchQueries(dataPagedListingKey);
+
```
sri2k1us, line 98: I think it will be good do a force refetch here to prevent cached data being returned. See https://github.com/tannerlinsley/react-query#querycacherefetchqueries

src/components/utils/errorCode.js
```diff
@@ -0,0 +1,7 @@
+export function getErrorCode(error) {
+    return error?.response?.data?.error_code;
+}
+
+export const ERROR_CODES = {
+    ERR_EXISTS: "ERR_EXISTS",
+};
```
sri2k1us, line 7: 👍

stories/data/CreateFolder.stories.js
```diff
@@ -0,0 +1,36 @@
...
+import { createFolderFailResp, createFolderSuccessResp } from "./DataMocks";
+import { mockAxios } from "../axiosMock";
+import CreateResourceDialog from "../../src/components/data/CreateFolderDialog";
+
+export const CreateFolderTest = () => {
+    const logger = (message) => {
+        console.log(message);
+    };
+
+    const successResp = boolean("Success Response", true);
```
sri2k1us, line 14: 👍

src/components/data/CreateFolderDialog.js
```diff
@@ -0,0 +1,165 @@
...
+                            </Button>
+                        </DialogActions>
+                    </Dialog>
+                </Form>
+            )}
+        </Formik>
+    );
+}
+
+export default withI18N(injectIntl(CreateFolderDialog), messages);
```
sri2k1us, line 165: I think we should add the `parent folder` path to this dialog to inform the user where the folder is being created.

sri2k1us: One other concern with Dialogs in general, if there is a service error (in this case, the folder already exists), we are displaying an error announcement. But the modal dialog makes it hard to realize that there was an announcement in the background (i tried this PR). Should we display these errors in the dialog itself just like how validation messages are displayed? Some thing to think about. Not a deal breaker for this PR.

ash3rz: Good point. I think I can refactor so that it will check for existing folder names and display an error within the dialog.

I also realized we should probably be checking permissions before displaying this button, so I'll add that as well.

ash3rz:
Thanks for the review!  I'll go ahead and merge this but any additional reviews still welcome!

src/components/data/UploadMenuBtn.js
```diff
@@ -29,17 +29,46 @@ import UploadDialog from "../uploads/dialog";
...
 import { injectIntl } from "react-intl";
 import intlData from "./messages";
+import { processSelectedFiles, trackUpload } from "../uploads/UploadDrop";
 
 const useStyles = makeStyles(styles);
 
+/**
+ * A component that remains hidden, but enables file browsing via the browser
+ * and passes back the selected files in onChange.  It's activated via the shared
+ * element id with the Browse Local menu item.
```
psarando, line 45: 👍

PR #101, opened 2020-04-24
Switch to using build-time environment variables

ash3rz:
The only issue with using `serverRuntimeConfig` in our next.config.js file, is that you have to use `getInitialProps` in order to have the client side of your application get access to those variables.  The hidden issue there is `getInitialProps` cannot be used in child components (see https://nextjs.org/docs/api-reference/data-fetching/getInitialProps#caveats). Our application is using a custom app (`_app`) in order to utilize a persistent layout between pages, so every page is actually a child component of `_app`.
nextjs currently recommends using build-time environment variables over runtime configuration as well (see https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration)

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -1,7 +1,7 @@
 const config = require("config");
 
 module.exports = {
-    serverRuntimeConfig: {
+    env: {
         INTERCOM_APP_ID: config.get("intercom.app_id"),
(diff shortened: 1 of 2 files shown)


ash3rz:
Thanks for the review!  I'll merge this now so Jack has these initial changes to work off of.  Additional reviews still welcome though

PR #102, opened 2020-04-28
Add IDs for the dashboard sections

johnworth:
Refactors the dashboard IDs as well to be more consistent and easy to
create.

### DIFF
--- a/src/components/dashboard/index.js
+++ b/src/components/dashboard/index.js
@@ -32,7 +32,7 @@ import ids from "./ids";
 import * as constants from "./constants";
 import { getDashboard } from "../../serviceFacades/dashboard";
 
-const makeID = (name) => buildID(ids.BASE, name);
+const makeID = (...names) => buildID(ids.BASE, ...names);
 
 const useStyles = makeStyles((theme) => ({
     root: {
@@ -165,7 +165,7 @@ const cleanDescription = (description) => {
 export const DashboardItem = (props) => {
     const classes = useStyles();
     const { kind, content } = props;
-    const cardID = `${kind}.${content.id}`;
+    const cardID = buildID(kind, content.id);
 
     const description = cleanDescription(content.description);
     const [origination, date] = getOrigination(kind, content);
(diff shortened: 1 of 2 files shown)


src/components/dashboard/index.js
```diff
@@ -32,6 +32,8 @@ import ids from "./ids";
 import * as constants from "./constants";
 import { getDashboard } from "../../serviceFacades/dashboard";
 
+const makeID = (name) => buildID(ids.BASE, name);
```
ash3rz, line 35: Minor, but I think you could refactor this to:
```suggestion
const makeID = (...args) => buildID(ids.BASE, ...args);
```

src/components/dashboard/index.js
```diff
@@ -163,7 +165,7 @@ const cleanDescription = (description) => {
 export const DashboardItem = (props) => {
     const classes = useStyles();
     const { kind, content } = props;
-    const cardID = `${kind}-${content.id}`;
+    const cardID = `${kind}.${content.id}`;
```
ash3rz, line 168: This could just be `buildID(kind, content.id)`

src/components/dashboard/index.js
```diff
@@ -172,7 +174,7 @@ export const DashboardItem = (props) => {
     return (
         <Card
             className={classes.dashboardCard}
-            id={buildID(ids.ITEM_BASE, cardID)}
+            id={makeID(`${ids.ITEM}.${cardID}`)}
```
ash3rz, line 177: With the above refactor to `makeID`, instead of trying to construct the same `parent.child` pattern here, you could then have this be `makeID(ids.ITEM, cardID)`. That way we're relying on `build`, or `buildID` in this case, to maintain that pattern.

src/components/dashboard/index.js
johnworth: This should be fixed now. Thanks for the suggestion.

src/components/dashboard/index.js
johnworth: Also fixed. Thanks for catching this.

src/components/dashboard/index.js
johnworth: Thanks for catching this.

PR #106, opened 2020-05-05
CORE-994: Add support for tool listing.

slr71:
This isn't done yet, but I think it's at a point where I'd like to get some reviews and merge the code soon. I can put subsequent updates in future PRs. Of course I'll address any suggestions that don't involve adding more features in this PR. Any and all suggestions are welcome!


src/components/tools/listing/Listing.js
```diff
@@ -0,0 +1,154 @@
...
+import { useQuery } from "react-query";
+import TableView from "./TableView";
+import { getTools } from "../../../serviceFacades/tools";
+import { injectIntl } from "react-intl";
+import { withI18N } from "@cyverse-de/ui-lib";
+import messages from "../Messages";
+import DEPagination from "../../utils/DEPagination";
+
+/**
+ * Returns the tool listing component.
```
psarando, line 11: I would say that this *is* "the tool listing component" 🤔

src/components/tools/listing/TableView.js
```diff
@@ -0,0 +1,251 @@
...
+        intl,
+        selected,
+        tableId,
+        tools,
+    } = props;
+    return (
+        <TableBody>
+            {error ? (
+                <LoadingError columns={columns} error={error} />
+            ) : !tools || tools.length === 0 ? (
```
psarando, line 167: Could also be
```suggestion
            ) : !(tools?.length) ? (
```

src/common/callApi.js
```diff
@@ -50,6 +50,10 @@ const callApi = (props) => {
         requestOptions.data = body;
     }
 
+    if (params !== null && params !== undefined) {
```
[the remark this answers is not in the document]
slr71: I do like that better. Thanks.

src/serviceFacades/tools.js
```diff
@@ -0,0 +1,53 @@
...
+        order !== undefined &&
+        orderBy != null &&
+        orderBy !== undefined;
+
+    // Determine if the request is supposed to be paginated.
+    const isPaginated =
+        page !== null &&
+        page !== undefined &&
+        rowsPerPage != null &&
+        rowsPerPage !== undefined;
```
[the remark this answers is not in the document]
slr71: Yeah, I'm surprised I didn't see a warning. At any rate, I like the updated code better.

src/components/tools/listing/TableView.js
slr71: 👍

src/components/tools/listing/Listing.js
slr71: Sounds reasonable. I updated the comment.

stories/tools/Listing.stories.js
```diff
@@ -0,0 +1,116 @@
...
+    erroredListing,
+    invalidSortField,
+    listing,
+} from "./ToolMocks";
+
+const toolListingUriRegexp = /\/api\/tools.*/;
+
+export default {
+    title: "Tools",
+};
```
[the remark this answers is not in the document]
slr71: I'll wait for others to chime in on this one.

src/components/tools/listing/Listing.js
```diff
@@ -0,0 +1,154 @@
...
+                order={order}
+                orderBy={orderBy}
+                selected={selected}
+            />
+            {data && data.total > 0 && (
+                <DEPagination
+                    baseId={baseId}
+                    onChange={handleChangePage}
+                    onPageSizeChange={handleChangeRowsPerPage}
+                    page={page + 1}
```
ash3rz, line 145: Not really related to this PR since we have this everywhere, but I realized in `handlePageChange` we have `setPage(newPage - 1)` and here we have `page + 1` which makes me think we can just use `page` everywhere?

src/components/tools/listing/TableView.js
```diff
@@ -0,0 +1,251 @@
...
+
+/**
+ * Returns localized column header information for the tool listing table.
+ * @param {Object} intl - the internationalization object
+ */
+const columnData = (intl) => [
+    {
+        id: ids.NAME,
+        name: formatMessage(intl, "name"),
+        numeric: false,
```
ash3rz, line 34: I believe `numeric` has been replaced with `align`, so this might be come `align: "left"` instead, for example

src/components/tools/listing/TableView.js
```diff
@@ -0,0 +1,251 @@
...
+        tools,
+    } = props;
+    return (
+        <TableBody>
+            {error ? (
+                <LoadingError columns={columns} error={error} />
+            ) : !(tools?.length) ? (
+                <NoTools columns={columns} />
+            ) : (
+                <ToolListing
```
ash3rz, line 170: 👍

stories/tools/Listing.stories.js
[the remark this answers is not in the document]
ash3rz: I'm also used to seeing the default export at the bottom, but that being said, I'm pretty sure I'm the one who put it at the top in the data listing's story 😊 I think this was the first story I created with the new [component story format](https://storybook.js.org/docs/formats/component-story-format/#named-story-exports) and I followed their example with the default export at the top.  I think in my subsequent story files, I put it at the bottom.

Probably for consistency, it'd be better to keep default exports at the bottom.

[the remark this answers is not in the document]
slr71: Cool. I'll move it.

src/components/tools/listing/TableView.js
slr71: It looks like explicitly specifying the `numeric` property to `false` isn't necessary, so I'll simply remove these. I did see `numeric` listed in the properties for `EnhancedTableHead`; it appears that it's used to set the `align` property. I don't know if that's the preferred way to set the alignment, though. Should I avoid using this property in the future?

slr71:
Thanks for the reviews!

src/components/tools/listing/TableView.js
ash3rz: Ah, no, you're correct - for some reason I remembered it being removed entirely, but it's still supported.  `align` makes more sense to me, but either is fine in that case.

PR #108, opened 2020-05-08
CORE-731 Initial App Launch Wizard view

psarando:
This is the initial PR for the App Launch Wizard view. This view has all functionality for displaying and validating app parameters and analysis submission fields, and it can also build the final submission JSON for launching the analysis or saving as a Quick Launch.

This view still needs to be integrated with a next.js page before it can be used in Sonora, but I would like to get this (what I have so far) reviewed before I go much further.

Also, feel free to offer code organization and documentation suggestions.

This could be one of the more complicated forms we have in Sonora so far (the app editor isn't converted yet!) so I want to be sure this code can be followed (and updated) by anyone on the team.

Also, don't feel obligated to review this entire PR if you still feel new to React or Formik, but any questions for clarification are welcome!

---

This App Launch Wizard view is using Material-UI's [horizontal, non-linear stepper](https://material-ui.com/components/steppers/#non-linear) with [bottom navigation buttons](https://material-ui.com/components/bottom-navigation/#bottom-navigation-2).

---

A toolbar at the top displays the App's name (wrapped in a `Paper` component), with its description just below the toolbar.
The toolbar could contain additional actions in the future, such as a button for fetching and displaying the app's documentation.
DE Word Count - Step 1

---

The MUI Stepper and Bottom Navigation automatically handle mobile view, but the stepper labels are hidden on small screens so all 4 steps will fit:
DE Word Count - Step 1 - mobile

The Step content between the Steps header and the Bottom Navigation is at a fixed height, so if a step is very tall, the navigation buttons at the bottom will remain in the same place and not move up or down as the user changes steps.

Note that each Step's label is actually a button that allows the user to select it and go directly to that step.

---
Deprecated and Disabled Apps
---

If an app is deprecated or disabled, the stepper and launch form are entirely replaced with an appropriate message:

DE Word Count - deleted

DE Word Count - disabled

Similarly for apps that use deprecated app parameters (currently only the List Tree parameter):

deprecated params

---
Parameters Step
---

App parameter groups are displayed in expansion panels, which are all automatically expanded. 

DE Word Count - Step 2 Params

Mobile:
DE Word Count - Step 2 Params - mobile

---
Resource Requirements
---

Resource requirements for each step in a DE app are included in an "Advanced" step:

DE Word Count - Step 3 Advanced

---

Resource Requirements for an app with 2 steps (a pipeline workflow):

2-step App Resource Requirements

---
Review and Launch Step
---

The final Review & Launch step displays all the parameters the user has entered (or any set by default). The `Next` button is replaced with a `Launch` button, and a `Save as Quick Launch` button is added in the middle of the Bottom Navigation:
DE Word Count - Step 4 Review Launch

---
Input Parameters
---

Multi-file Selection Input parameter:

Multi-file Input without Selection

With files added to the selector:

Multi-file Input with Selection

---

Inputs in the Review step:

image

---
Text Parameters and Validation Errors
---

Text fields with validation:

Text fields with validation

---

Note that when a field displays a validation error, its step also displays an error icon:

Text Inputs Step 2 error

---

When a field has a validation error, its label will also be marked as an error in the Review step:

Text Inputs Step 4 error

---
Info Text
---

Info text can display sanitized HTML:

image

---
Flag Parameters
---

Checkbox Parameters

In the Review step:

Checkbox Review

---
Number Parameters
---

Numbers with validations:

Numbers with validations

---
List Parameters
---

Lists with validation

List parameters in the Review step:

Lists Review

---
Reference Genome/Sequence/Annotation Parameters
---

The Reference Genome list is fetched from the service only once by the App Launch Form, and only if the form contains at least 1 Reference Genome/Sequence/Annotation parameter. These parameters start by rendering as a loading skeletons, which are also form fields that can display validation errors (if the user clicks the `Launch` button before the list loads):

image

Once the list loads, these fields render as selection lists. However, rather than using the `FormSelectField` from `ui-lib` (which uses the `@material-ui/core/Select` component), I'm using our `FormTextField`, which uses `@material-ui/core/TextField`, which now supports a [select](https://material-ui.com/components/text-fields/#select) prop:

Reference Genome, Sequence, Annotation parameters

---
Agave/TAPIS Apps
---

TAPIS apps do not support the `Retain Inputs` flag:
TAPIS app step 1

Also, our info text about HPC wait-times will now always display in the Review step, rather than displaying in a popup when the user launches the app:
TAPIS app wait times

Notice  that TAPIS apps do not display the `Save as Quick Launch` button, since it's not supported for this app type.

---
Skipping / Hiding Steps
---

If an app has no parameters, then there is no `Parameters` step:
image

---
Known Issues
---

The way I'm building a file/folder selector means the `Browse` button can be misaligned with its `TextField` when either in mobile view (see Parameters Step mobile screenshot above), or for input fields with vs. without help text / validation error messages.

I didn't want to wrap both the button and the text field in the same [form control](https://material-ui.com/api/form-control/), since the docs say that "️Only one input can be used within a FormControl".
I may be able to combine these input fields under 1 component, which could be wrapped, but I would have to experiment with it to see what that does to the `aria` attributes.

---

Another issue I ran into was trying to make an input selector field "touched" once the browse button is used, so that it can display validation errors if the user closes the drawer without selecting anything in a required field; but it appears the form will crash once all items are removed from a multi-select input list if I imperatively use `setFieldTouched` with this kind of Formik array field.

I also ran into Formik issues trying to enable range select and remove for multi-select field, which also did not properly display a `Required` validation error, for some reason. So range select was removed for now (could be fixed with a Formik upgrade).

If the user clicks the `Launch` button, or multi-select inputs are removed individually with the `X` icon in the grid, then the validation error message will display.


src/components/apps/launch/ParamGroups.js
```diff
@@ -0,0 +1,375 @@
...
+                                    (validator) =>
+                                        validator.type ===
+                                        constants.VALIDATOR_TYPE.CHARACTER_LIMIT
+                                );
+                                if (charLimitValidator) {
+                                    fieldProps.inputProps = {
+                                        maxLength: charLimitValidator.params[0],
+                                    };
+                                }
+                            }
```
psarando, line 190: This is the only spot where param validation is checked outside of the `validate` file, since `maxLength` can be set directly on a text field.

src/components/apps/launch/ParamGroups.js
```diff
@@ -0,0 +1,375 @@
...
+                            const fieldName = `groups.${groupIndex}.parameters.${paramIndex}`;
+                            const paramError = getIn(errors, fieldName);
+                            const hasValue = !isEmptyParamValue(param.value);
+
+                            return (
+                                param.isVisible &&
+                                (paramError || hasValue) && (
+                                    <TableRow key={param.id}>
+                                        <TableCell>
+                                            <ParamsReviewLabel
```
psarando, line 352: Parameters will only display in the `Review` step if they have a value set (either by default or by the user), or if they have a validation error.

src/components/apps/launch/ResourceRequirements.js
```diff
@@ -0,0 +1,357 @@
...
+    TableContainer,
+    TableRow,
+    Typography,
+} from "@material-ui/core";
+
+import { ExpandMore } from "@material-ui/icons";
+
+const useStyles = makeStyles(styles);
+
+const ONE_GB = 1024 * 1024 * 1024;
```
psarando, line 45: Maybe this could be refactored as a global util constant (it's reused at least in the app launch stories).

src/components/apps/launch/ResourceRequirements.js
```diff
@@ -0,0 +1,357 @@
...
+                            requirements={limits[0]}
+                            index={0}
+                            defaultMaxCPUCores={defaultMaxCPUCores}
+                            defaultMaxMemory={defaultMaxMemory}
+                            defaultMaxDiskSpace={defaultMaxDiskSpace}
+                        />
+                    ) : (
+                        limits.map((reqs, index) => {
+                            return (
+                                <ExpansionPanel key={reqs.step_number}>
```
psarando, line 203: If there is more than 1 step in the app, then each step gets its own Resource Requirements expansion panel.

src/components/apps/launch/ResourceRequirements.js
```diff
@@ -0,0 +1,357 @@
...
+
+    return requirements.map((stepRequirements) => (
+        <StepResourceRequirementsReview
+            baseId={baseId}
+            key={stepRequirements.step_number}
+            stepRequirements={stepRequirements}
+            headerMessageKey={headerMessageKey}
+        />
+    ));
+}, messages);
```
psarando, line 355: I kinda wish `withI18N` took `messages` as its first param 🤔 
If this looks too awkward to others, I can refactor my components out of the `withI18N` calls.

For example, this could become something like this:
```javascript
const ResourceRequirementsReview = withI18N(ResourceRequirementsReviewBase, messages);
```

src/components/apps/launch/MultiInputSelector.js
```diff
@@ -0,0 +1,324 @@
...
+                                    size="small"
+                                    startIcon={<ClearIcon />}
+                                    onClick={() => {
+                                        setSelected([]);
+
+                                        const filteredPaths = paths.filter(
+                                            (path) => !selected.includes(path)
+                                        );
+
+                                        setFieldValue(
```
psarando, line 179: I'm wondering if using `setFieldValue` here may be causing the validation/touched issues described in the main PR description.

I'm tempted to remove support for range-select and the `Remove` selected button, since the user can remove items individually, and if they need a large number of input paths then they could use a multi-input path list file; but then I think we will run into this same issue for the  multi-input path list creator component 🤔

stories/apps/launch/AgaveWordCount.stories.js
```diff
@@ -0,0 +1,27 @@
...
+            notify={false}
+            defaultOutputDir={ANALYSIS_OUTPUT_DIR}
+            startingPath={STARTING_PATH}
+            submitAnalysis={submitAnalysis}
+            app={AgaveWordCountApp}
+        />
+    );
+};
+
+export default { title: "Apps / Launch" };
```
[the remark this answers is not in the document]
psarando: Yep. This message is currently displayed in a popup dialog when a user launches an HPC app in the current DE, and since popups are really obtrusive and can get annoying fast, we give the user a preference to no longer show the popup in the future.

I think I like this better, where the message is always displayed, but not in an obtrusive popup. So hopefully it's easy to get past but not easily forgotten 👍

src/components/apps/launch/AppLaunchWizard.js
```diff
@@ -0,0 +1,42 @@
...
+
+import constants from "./constants";
+import messages from "./messages";
+
+import AppInfo from "./AppInfo";
+import AppLaunchForm from "./AppLaunchForm";
+
+import { withI18N } from "@cyverse-de/ui-lib";
+
+const deprecatedParamTypes = Object.values(constants.DEPRECATED_PARAM_TYPE);
```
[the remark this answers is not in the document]
psarando: On the other hand, I see no reason to deprecate this parameter in the API. We just don't want to support it in our client 😉

src/components/apps/launch/AppLaunchWizard.js
```diff
@@ -0,0 +1,42 @@
...
+    const { app } = props;
+
+    const hasDeprecatedParams = app?.groups?.reduce(
+        (hasDeprecatedParams, group) =>
+            hasDeprecatedParams ||
+            group.parameters?.find((param) =>
+                deprecatedParamTypes.includes(param.type)
+            ),
+        false
+    );
```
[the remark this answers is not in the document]
psarando: Nice, I think that is a bit cleaner.

src/components/apps/launch/AppLaunchForm.js
```diff
@@ -0,0 +1,600 @@
...
+        step: hasParamsStep ? 2 : 1,
+    };
+
+    const steps = hasParamsStep
+        ? [stepAnalysisInfo, stepParameters, stepReviewAndLaunch]
+        : [stepAnalysisInfo, stepReviewAndLaunch];
+
+    const isLastStep = () => {
+        return activeStep === steps.length - 1;
+    };
```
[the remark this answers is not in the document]
psarando: Ah, cool 👍
These weren't my idea, though. They were copied from the MUI Stepper component example 😆

src/components/apps/launch/AppLaunchWizard.js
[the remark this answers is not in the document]
slr71: Yeah, that's a good point. At least maintaining the code in the API isn't costing us development cycles.

src/components/apps/launch/AppLaunchForm.js
```diff
@@ -0,0 +1,600 @@
...
+                    setReferenceGenomesLoading(false);
+                });
+        }
+
+        return () => {
+            unmounted = true;
+        };
+    }, [props.app, hasReferenceGenomes]);
+
+    const hasParamsStep = hasParams || requirements?.length > 0;
```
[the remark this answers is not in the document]
slr71: I kind of like the idea of putting the resource requirements in the `Review` step instead. They seem fundamentally distinct from the app parameters to me.

src/components/apps/launch/AppLaunchForm.js
```diff
@@ -0,0 +1,600 @@
...
+
+    const stepAnalysisInfo = { label: getMessage("analysisInfo"), step: 0 };
+    const stepParameters = { label: getMessage("parameters"), step: 1 };
+    const stepReviewAndLaunch = {
+        label: getMessage("reviewAndLaunch"),
+        step: hasParamsStep ? 2 : 1,
+    };
```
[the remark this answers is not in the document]
psarando: You might not have seen my edit to this PR's description, but I added some screenshots of Agave apps near the bottom (just above the `Known Issues` section), and I added a screenshot of a no-param Agave app:
no-params screenshot

However, if we all agree the Resource Requirements should be moved into the Review step, then I do have a story for a JupyterLab VICE app with no params.

stories/apps/launch/ReferenceGenomeParams.stories.js
```diff
@@ -0,0 +1,37 @@
...
+            defaultOutputDir={ANALYSIS_OUTPUT_DIR}
+            startingPath={STARTING_PATH}
+            saveQuickLaunch={saveQuickLaunch}
+            submitAnalysis={submitAnalysis}
+            app={ReferenceGenomeApp}
+        />
+    );
+};
+
+export default { title: "Apps / Launch" };
```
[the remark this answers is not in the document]
psarando: I like this impulse, except I think I'll extract another React component for re-use:
```jsx
export default (props) => {
    initMockAxiosFileFolderSelector();

    return (
        <AppLaunchWizard
            notify={false}
            defaultOutputDir={ANALYSIS_OUTPUT_DIR}
            startingPath={STARTING_PATH}
            saveQuickLaunch={saveQuickLaunch}
            submitAnalysis={submitAnalysis}
            {...props}
        />
    );
};
```

Then in the Reference Genome story file, for example:
```jsx
export const ReferenceGenomeParams = () => {
    mockAxios
        .onGet(/\/api\/reference-genomes/)
        .reply(200, ReferenceGenomeListing);

    return <AppLaunchStoryBase app={ReferenceGenomeApp} />;
};
```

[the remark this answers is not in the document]
slr71: I like it! 👍

src/components/apps/launch/AppLaunchForm.js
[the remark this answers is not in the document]
slr71: Ah, yes, that'll definitely work.

stories/apps/launch/DEWordCount.stories.js
```diff
@@ -0,0 +1,40 @@
...
+            notify={false}
+            defaultOutputDir={ANALYSIS_OUTPUT_DIR}
+            startingPath={STARTING_PATH}
+            saveQuickLaunch={saveQuickLaunch}
+            submitAnalysis={submitAnalysis}
+            app={app}
+            defaultMaxCPUCores={8}
+            defaultMaxMemory={4 * ONE_GB}
+            defaultMaxDiskSpace={64 * ONE_GB}
+        />
```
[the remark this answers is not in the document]
ash3rz: Makes sense to me that clicking on submit analysis would navigate to the Analyses listing.  It also makes sense to show the list of quick launches for an app after a user creates one, I would think. 

As for loading, the small thing I've done if it doesn't make sense to have a loading skeleton or navigate to the thing that does have a loading skeleton right away is add a loading spinner somewhere.  We haven't really discussed that in detail though.  We should probably come up with a pattern.

stories/apps/launch/Pipline.stories.js
```diff
@@ -0,0 +1,29 @@
...
+
+import {
+    ANALYSIS_OUTPUT_DIR,
+    STARTING_PATH,
+    initMockAxiosFileFolderSelector,
+    saveQuickLaunch,
+    submitAnalysis,
+} from "./constants";
+
+export const Pipline = () => {
```
ash3rz, line 14: tpyo?  Also in the file name

stories/apps/launch/Pipline.stories.js
```diff
@@ -0,0 +1,29 @@
...
+            defaultOutputDir={ANALYSIS_OUTPUT_DIR}
+            startingPath={STARTING_PATH}
+            saveQuickLaunch={saveQuickLaunch}
+            submitAnalysis={submitAnalysis}
+            app={PipelineApp}
+        />
+    );
+};
+
+export default { title: "Apps / Launch" };
```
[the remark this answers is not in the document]
ash3rz: Yes, we had a discussion in a thread about adding double-clicking.  @sri2k1us had a lot of points against double-clicking.

stories/apps/launch/data/TextParamsApp.js
```diff
@@ -0,0 +1,85 @@
...
+            label: "Text Input",
+            parameters: [
+                {
+                    description: "",
+                    arguments: [],
+                    name: "",
+                    type: "Info",
+                    validators: [],
+                    label:
+                        "<h4>Info Text!</h4><p>Can display <b>HTML</b>:</p><p>The <b>Dark Arts</b> better be worried,<br/><em>oh boy</em>!</p>",
```
ash3rz, line 28: 😁

src/components/apps/launch/AppLaunchForm.js
```diff
@@ -0,0 +1,600 @@
...
+const paramConfigsReducer = (configs, group) => {
+    group.parameters.forEach((param) => {
+        const { id, type } = param;
+
+        if (type !== constants.PARAM_TYPE.INFO) {
+            let { value } = param;
+
+            switch (type) {
+                case constants.PARAM_TYPE.FLAG:
+                    value = value && value !== "false";
```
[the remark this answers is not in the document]
psarando: I'm using the [FormCheckboxStringValue](https://github.com/cyverse-de/ui-lib/blob/2f9df43cf68b79baf1dd3f85a21710964b5b9273/src/util/FormField.js#L124-L138), which stores a string value of "true" or "false" in the form, rather than a literal boolean. I'm using the string values so I don't have to do any conversions to display the selected value in the Review step, but that means converting the string value to a literal boolean for the submission configs here.

If it seems clearer, I could swap these fields to use a regular [FormCheckbox](https://github.com/cyverse-de/ui-lib/blob/2f9df43cf68b79baf1dd3f85a21710964b5b9273/src/util/FormField.js#L89-L106), and convert the boolean values to display strings in the Review step instead. Maybe that's even better, so we can choose what string to display in the review step if we want something other than "true" or "false" 🤔

src/components/apps/launch/ReferenceGenomeSelect.js
```diff
@@ -0,0 +1,91 @@
...
+} from "@material-ui/core";
+import { Skeleton } from "@material-ui/lab";
+
+const useStyles = makeStyles((theme) => ({
+    skeleton: {
+        height: theme.spacing(6),
+    },
+}));
+
+const FormFieldLoading = ({
```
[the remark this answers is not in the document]
psarando: I was wondering that as well. Especially if we'll need more selection lists that load from endpoint calls 🤔

src/components/apps/launch/validate.js
```diff
@@ -0,0 +1,291 @@
...
+                            });
+                        }
+                        break;
+
+                    default:
+                        break;
+                }
+
+                if (validatorMsg) {
+                    errorMsg = validatorMsg;
```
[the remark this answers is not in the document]
psarando: Yes, `validators` is an array. I did intentionally go with only displaying 1 validation error at a time, and which one displays depends on the order the app integrator adds them to the param.

The Number Params story has an Integer param with 3 validators, even if they're not very realistic. So in this story, the validations are arranged so that `Must be greater than 1` will display for numbers less than 2, then `Must be between 3 and 7` will display for 2, 8, and 9, then `Must be less than 10` for numbers 10 or higher.

src/components/apps/launch/ParamGroups.js
```diff
@@ -0,0 +1,375 @@
...
+                            fieldProps = {
+                                id: paramFormId,
+                                name,
+                                component: Typography,
+                                variant: "body1",
+                                gutterBottom: true,
+                                dangerouslySetInnerHTML: {
+                                    __html: sanitizeHtml(param.label),
+                                },
+                            };
```
sri2k1us, line 173: Same Q. Why do we have to set inner html ?

src/components/apps/launch/ParamGroups.js
sri2k1us: Hmm 🤔 What if they have accidentally left a field. I am leaning toward displaying all the parameters in the review step irrespective of user selection.

src/components/apps/launch/ReferenceGenomeSelect.js
```diff
@@ -0,0 +1,91 @@
...
+    const helperTextID = buildDebugId(loadingFieldID, "helper-text");
+
+    return (
+        <FormControl fullWidth error={!!errorMsg}>
+            <InputLabel htmlFor={loadingFieldID} required={required}>
+                {label}
+            </InputLabel>
+            <Skeleton
+                id={loadingFieldID}
+                aria-describedby={helperTextID}
```
sri2k1us, line 57: 👍 I haven't been adding `aria-describedby` to loading skeleton. I Should go back and add them.

src/components/apps/launch/ResourceRequirements.js
```diff
@@ -0,0 +1,360 @@
...
+                            )}
+                        />
+                    }
+                >
+                    <Typography variant="subtitle1">
+                        <ResourceRequirementsHeader
+                            headerMessageKey={headerMessageKey}
+                            step={step_number + 1}
+                        />
+                    </Typography>
```
sri2k1us, line 288: I think i had previously requested this feature in DE as well. Is it possible to add the step name here ? An end user coming to DE just to use an `app` will have no understanding of multi-step/single step app.

src/components/apps/launch/messages.js
```diff
@@ -0,0 +1,81 @@
...
+        analysisParameters: "Analysis Parameters",
+        appDeprecated:
+            "This application has been deprecated." +
+            " If you need access to it, please contact" +
+            ' <a href="mailto:support@cyverse.org">support@cyverse.org</a>.',
+        appParamsDeprecated:
+            "This application uses 1 or more parameters that have been deprecated." +
+            " If you still need access to this application, please contact" +
+            ' <a href="mailto:support@cyverse.org">support@cyverse.org</a>.',
+        appUnavailable: "This app is temporarily unavailable.",
```
sri2k1us, line 40: We should display intercom button (first check intercom availability using `useIntercom` context) for contacting support whenever possible. That way, they get queued for our team. Emails to `support@cyverse.org` will go to `general` queue in intercom and can get lost.

stories/apps/launch/Pipline.stories.js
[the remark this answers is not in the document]
sri2k1us: Double clicks are strongly discouraged in web apps. See https://core-sw.slack.com/archives/C2CK881B5/p1587963471458900?thread_ts=1587688778.431600

sri2k1us:
I will add my comments on Screenshots here 

- Am not entirely sure how an `AppBar` within `AppBar (Sonora layout)` will render in the page.
- App description can use little more styling to make it standout.
- More styling around disable, deleted and deprecated app messages. For example the text could be in `error` theme.
- Avoid displaying support email address and use intercom button when possible.
- Some spacing between `Remove` and `Browse` button in multi-input select
- May be change `Browse` to `Add` ? Not sure but it makes more sense to pair `Add` with `Remove` 🤔 
- The icon used for `Launch Analysis` is the same icon for going to VICE analyses page.  May be switch to using `import ExitToAppIcon from '@material-ui/icons/ExitToApp';` for VICE ?

src/components/apps/launch/AppInfo.js
```diff
@@ -0,0 +1,48 @@
...
+            </Typography>
+
+            {(app?.deleted || app?.disabled || hasDeprecatedParams) && (
+                <Typography
+                    variant="body2"
+                    gutterBottom
+                    dangerouslySetInnerHTML={{
+                        __html: formatMessage(intl, unavailableMsgKey),
+                    }}
+                />
```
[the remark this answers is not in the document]
psarando: The HTML link in the message will only render if the message is set as the inner HTML. Otherwise the raw tags will display with the message.

src/components/apps/launch/ParamGroups.js
psarando: This text is coming from the app integrator, and in order to support a limited set of HTML (as we do in the current UI), we have to sanitize the string then set it as the inner HTML (again so the raw tags don’t display in the text).

src/components/apps/launch/ReferenceGenomeSelect.js
psarando: I’m only doing that here because there might be a validation message in the helper text. This was basically copied from the MUI example for custom form control labels 🙂

src/components/apps/launch/ResourceRequirements.js
psarando: It will be possible with an update to the app description endpoint, or I can look for another endpoint we can call to get this info.

src/components/apps/launch/messages.js
psarando: This message was just copied from the current UI, but I’ll look into updating it. 👍🏼

src/components/apps/launch/AppInfo.js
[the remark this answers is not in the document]
sri2k1us: May be look into - https://formatjs.io/docs/react-intl/components#rich-text-formatting

src/components/apps/launch/ParamGroups.js
psarando: In that case I don’t see too much of a difference between the param step and the review step. If there are only 1 or 2 required params, and 10 or 20 optional params, then I would think having 10+ blank fields in the review step would cause a lot of noise.

This reminds me to find an app in prod with a lot of optional params (I’ve seen at least one with close to 20 params spread in a bunch of groups before).

sri2k1us: The difference is you cannot edit values directly in the review step ;)

Exactly why I would like see all parameters in the review step. If there are 20 parameters, it is easy a miss one even though it is optional but needed for that analyses for that user.  Again these parameters will be grouped in the review step as well. So I don't see why we cannot include them here.

sri2k1us: Visual noise is something we need to understand from watching the user interact with this screen. If it is evident that users are distracted with this, I am in for removing / hiding these as an optional feature.

src/components/apps/launch/AppLaunchForm.js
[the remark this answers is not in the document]
ash3rz: Ah, I see.  I didn't notice the usage of `FormCheckboxStringValue`, so I was mostly confused about that.  I'll leave it up to you if you want to switch.  Now that I'm aware of it, it makes sense, so leaving it is fine too 👍

src/components/apps/launch/validate.js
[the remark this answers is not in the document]
ash3rz: Got it. Thanks for the clarification and example!

src/components/apps/launch/ParamGroups.js
psarando: I guess my thinking was that if the user wanted to review all the parameters, including the ones left blank, then the user could simply review the Params step, and there's no point duplicating that info in the Review step.

Although simply showing the param labels and values, without Info text and the groups between the params in the Review step, is a different view from the Params step 🤔

I agree that we should get input from the sci-team for this.

In the meantime, I found a public app (with 119 analysis completed so far) that has over 50 parameters that I can use in a story, and maybe provide screenshots of the alternate Review step views.

psarando: I was also imagining an app that had a lot of optional parameters, but the user only needed to pick 1 or 2 out of the bunch, and in that scenario I thought only showing the parameters the user entered would be more convenient.

Using the actual app from prod with ~50 parameters, however, I see an example where most of those params are required, or have a default value already set. There are also a couple of pairs of params that are not marked as required, but the user must still provide a value for 1 of the params in the pair. In this scenario, I can see your point where they could easily overlook one of these not-required-but-still-required params if they're hidden from the Review step.

sri2k1us: Thanks @psarando for digging into this further. If It is not too much trouble, can you find out how many apps are in prod with 1. <= 10 params  2. 10 - 25 params  3.  > 25 before next UX meeting. And then we can discuss more about this in our upcoming UX meeting

sri2k1us: Again my comments on this issue are not show stopper. So feel free to merge anytime 👍

psarando:
> * The icon used for `Launch Analysis` is the same icon for going to VICE analyses page.  May be switch to using `import ExitToAppIcon from '@material-ui/icons/ExitToApp';` for VICE ?

What about using the `PlayArrow` icon for `Launch Analysis`?
We're already using that for launching a `Quick Launch`.

src/components/apps/launch/messages.js
psarando: I'll have to save this for another PR, because it looks like our Intercom config is setup to detect our custom Intercom launcher by ID, and there should only be one element with that ID in the DOM:
https://github.com/cyverse-de/sonora/blob/94991edb75081c6cebb2d3c051c98513d6d88204/src/pages/_app.js#L34
Maybe we can refactor it to use some other selector 🤔

sri2k1us:
> > * The icon used for `Launch Analysis` is the same icon for going to VICE analyses page.  May be switch to using `import ExitToAppIcon from '@material-ui/icons/ExitToApp';` for VICE ?
> 
> What about using the `PlayArrow` icon for `Launch Analysis`?
> We're already using that for launching a `Quick Launch`.

Sounds good to me. So `Launch Analysis` can use `PlayArrow` and `FastForward` for Quick Launch ?

src/components/apps/launch/messages.js
sri2k1us: Please see https://github.com/cyverse-de/sonora/blob/2ed058ec05473e90dfe172c2fee177f92d893910/src/components/utils/error/ErrorHandler.js#L91 on how to display a button to intercom support button. That reminds me that I need to check if intercom is enabled.

psarando: Ah, thanks!

> That reminds me that I need to check if intercom is enabled here as well.

In that case, I'll create a `showIntercom` function in `/src/common/intercom.js`, then I'll refactor `ErrorHandler` to use that as well 👍

psarando:
> Sounds good to me. So `Launch Analysis` can use `PlayArrow` and `FastForward` for Quick Launch ?

In this app launch wizard I was going to still use the `Save` (disk) icon for saving a Quick Launch, but I think it makes sense to keep using the `PlayArrow` for using a Quick Launch (from the apps listing or apps details).

psarando:
Thanks for the reviews!

I've updated this PR's description and screenshots, so I'll go ahead and merge this now, but any additional reviews are still welcome 👍

PR #110, opened 2020-05-15
Add @next/bundle-analyzer in a non-ad-hoc way

ianmcorvidae:
A small thing, but still probably nice to have outside my local copy. Means that if you run a next build with `ANALYZE=true` as an environment variable, you'll get an analysis of the webpack bundles (server & client both, though client is the more important one for us usually).

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -1,10 +1,14 @@
 const config = require("config");
 
-module.exports = {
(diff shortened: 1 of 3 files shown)


ianmcorvidae:
Cool, thanks for the lookover

PR #111, opened 2020-05-19
Chromatic delay

ash3rz:
Most of our snapshots in Chromatic are of loading skeletons.  This should wait until the fake axios calls have completed their delay (2 seconds) and then I added another half a second just to account for any animations/transitions (like fade in or slide in or whatever else used by MUI).  Hopefully with the wait in place, we'll stop seeing those really minor differences in components that didn't actually change as well.

Also I fixed some storybook errors about missing props.

[Chromatic Build](https://www.chromatic.com/build?appId=5ebad5c6b991d60022b1e72e&number=21)

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,9 +1,10 @@
 import React, { useEffect } from "react";
 import { withConsole } from "@storybook/addon-console";
-import { addDecorator, configure } from "@storybook/react";
+import { addDecorator, addParameters, configure } from "@storybook/react";
 import { CyVerseAnnouncer } from "@cyverse-de/ui-lib";
 import { ThemeProvider } from "@material-ui/core/styles";
 import theme from "../src/components/theme/default";
+import { AXIOS_DELAY } from "../stories/axiosMock";
 import {
     UserProfileProvider,
     useUserProfile,
@@ -48,5 +49,7 @@ addDecorator((storyFn) => (
 
 //redirect console error / logs / warns to action logger
 addDecorator((storyFn, context) => withConsole()(storyFn)(context));
+
+addParameters({ chromatic: { delay: AXIOS_DELAY + 500 } });
 
 configure(require.context("../stories", true, /\.stories\.js$/), module);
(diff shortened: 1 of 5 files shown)


ianmcorvidae:
Poking through the baselines on Chromatic. I see that the Apps Listing Test is still a skeleton. The rest I've marked as accepted baselines, which seems to be a shared thing. Not sure if we need to do something to handle the apps listing or what.

ianmcorvidae:
I think this may be because we have to make several calls on that page, chained -- one to get the categories, and then another to list the apps within the category. So, since the calls are chained, it needs at least twice the mock axios timeout, I guess

ash3rz:
> I think this may be because we have to make several calls on that page, chained -- one to get the categories, and then another to list the apps within the category. So, since the calls are chained, it needs at least twice the mock axios timeout, I guess

Good catch! I can bump up the delay for that particular story.  In that case, I wonder if it'd be preferable to set the delay on each individual story instead so each story has control over how much delay they need.  I set it globally just because I figured it'd be convenient to generally not have to worry about it.

ash3rz:
> This seems reasonable to me. When I was working on the tool listing stuff, I found myself wondering if it might be helpful to remove the global Axios delay entirely and to create separate stories to test the loading masks. I don't know how much of an additional burden that would place on us, though.

OK, that gives me an idea.  Maybe there's a way instead to turn the Axios delay on and off, maybe with a global storybook knob or something.  If so, by default we could have it disabled and then as you're developing, you can choose to enable it. I do personally like having the loading added in because I will otherwise forget about it, but I get your point about not wanting it all the time. Let me explore that idea.

slr71:
> > This seems reasonable to me. When I was working on the tool listing stuff, I found myself wondering if it might be helpful to remove the global Axios delay entirely and to create separate stories to test the loading masks. I don't know how much of an additional burden that would place on us, though.
> 
> OK, that gives me an idea. Maybe there's a way instead to turn the Axios delay on and off, maybe with a global storybook knob or something. If so, by default we could have it disabled and then as you're developing, you can choose to enable it. I do personally like having the loading added in because I will otherwise forget about it, but I get your point about not wanting it all the time. Let me explore that idea.

I like that idea! 👍

ash3rz:
I think it's possible to have a global configuration for turning on and off the Axios delay in Storybook, but I'm not immediately finding something that's working.  For now, I'll merge this so we can at least get our snapshots to be more useful.

Thanks for the reviews!

PR #112, opened 2020-05-20
Dashboard Updates

johnworth:
This PR does the following:
* Changes the title and subtitle to a CardHeader.
* Assigns an avatar icon based on the content type.
* Changes the primary color of the CardHeader and open button based on the section. Needed in order to provide a bit of color to the dashboard.
* Adds a default description of "No description was provided." for the cards.
* Links to the planned details pages for apps and analyses using `router.push`.
* Adjusts the sizes of the cards slightly to allow more items per row.
* Modifies field limits and card layouts for mobile.

Limitations:
* Links for Apps and Analyses currently 404 since the routing for those hasn't been implemented yet and are outside the scope of this PR.

Notes:
* Colors for the cards come from the existing CyVerse palette in ui-lib.

image

image

image

image

image

image

### DIFF
--- a/src/components/dashboard/constants.js
+++ b/src/components/dashboard/constants.js
@@ -23,3 +23,5 @@ export const USER_SUFFIX = "@iplantcollaborative.org";
 export const CYVERSE = "CyVerse";
 export const DESC_MAX_LENGTH = 140;
 export const TITLE_MAX_LENGTH = 31;
+export const TITLE_MAX_LENGTH_SMALL = 25;
+export const SUBHEADER_MAX_LENGTH_SMALL = 39;
--- a/src/components/dashboard/constants.js
+++ b/src/components/dashboard/constants.js
@@ -6,6 +6,8 @@
  * @module dashboard/constants
  */
 
+import meta from "../../constants";
+
 export const KIND_ANALYSES = "analyses";
 export const KIND_APPS = "apps";
 export const KIND_FEEDS = "feeds";
@@ -19,7 +21,7 @@ export const SECTION_NEWS = "news";
 export const SECTION_EVENTS = "events";
 export const SECTION_ITEM_LIMIT = 8;
 
-export const USER_SUFFIX = "@iplantcollaborative.org";
+export const USER_SUFFIX = `@${meta.IPLANT}.org`;
 export const CYVERSE = "CyVerse";
 export const DESC_MAX_LENGTH = 140;
 export const TITLE_MAX_LENGTH = 31;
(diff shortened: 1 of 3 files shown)


src/components/dashboard/index.js
```diff
@@ -144,14 +325,70 @@ const cleanUsername = (username) => {
...
+const cleanTitle = (title, isLarge = true) => {
+    let comparitor;
+
+    if (isLarge) {
+        comparitor = constants.TITLE_MAX_LENGTH;
+    } else {
+        comparitor = constants.TITLE_MAX_LENGTH_SMALL;
+    }
+
+    return cleanField(title, comparitor);
```
ash3rz, line 350: This comes down to preference, but just a reminder that this type of simple conditional could also be written with the `? :` ternary syntax

src/components/dashboard/index.js
```diff
@@ -130,6 +195,111 @@ const getOrigination = (kind, content) => {
...
+            retval = (
+                <Event color="primary" classes={{ colorPrimary: colorClass }} />
+            );
+            break;
+        default:
+            retval = (
+                <Apps color="primary" classes={{ colorPrimary: colorClass }} />
+            );
+    }
+    return retval;
```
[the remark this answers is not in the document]
johnworth: Yeah, I'm not sure what the right thing to do is. In Go, doing that would be 100% acceptable, but in other languages like Python and Java I've been dinged in reviews for having multiple return statements in a function/method. I've been flip/flopping on this in JS for a bit, now.

src/components/dashboard/index.js
```diff
@@ -144,14 +325,70 @@ const cleanUsername = (username) => {
     return user;
 };
 
-const cleanDescription = (description) => {
-    let desc;
-    if (description.length > constants.DESC_MAX_LENGTH) {
-        desc = description.slice(0, constants.DESC_MAX_LENGTH) + "...";
+const cleanField = (field, comparitor) => {
```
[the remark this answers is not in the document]
johnworth: Should be fixed.

src/components/dashboard/constants.js
```diff
@@ -21,3 +22,6 @@ export const SECTION_ITEM_LIMIT = 8;
 export const USER_SUFFIX = "@iplantcollaborative.org";
```
sri2k1us, line 22: Minor. I use something similar in analyses too - https://github.com/cyverse-de/sonora/blob/b597a4f14a2198676576b72679dbd33c44866417/src/constants.js#L27 and https://github.com/cyverse-de/sonora/blob/b597a4f14a2198676576b72679dbd33c44866417/src/components/analyses/listing/TableView.js#L284. I guess we don't need this declared again here and just the use one from `src/constants.js` ?

johnworth: I changed the constant so it's based on the top-level `IPLANT` constant. Trying to limit inter-component dependencies, otherwise.

PR #115, opened 2020-05-28
CORE-994: added some tests for tool listings.

slr71:


src/__tests__/tools.js
```diff
@@ -0,0 +1,35 @@
...
+});
+
+afterEach(() => {
+    mockAxios.reset();
+});
+
+test("Tool Table View renders", () => {
+    const component = renderer.create(<ToolListingTest />);
+    let tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
```
psarando, line 22: I thought we wanted to move away from snapshots and use `chromatic` instead?
So in my last PR I simply rendered a component and unmounted it, to ensure it wouldn't crash in either phase.
```suggestion
    component.unmount();
```

slr71: I thought that matching the snapshot was using chromatic. 😕 I could easily be mistaken, though. I'm still a newbie when it comes to UI work.

psarando: I think chromatic automatically creates its own snapshots and rendered images based on the stories that are added in a PR.

My last PR #108 didn't include any snapshot files for the components I created, but chromatic still rendered all the stories I added.

sri2k1us: Chromatic automates snapshot creation and comparison. No need to manually create and update snapshot with each change 🎉

slr71: Ah, okay. Thanks. I'll update the code.

slr71:
I think I've addressed everything, so I'm going to merge this now. Please feel free to ping me if there's anything I missed. I can fix any remaining problems in a subsequent pull request.

PR #116, opened 2020-05-29
Add "chromatic-ignore" classname

ash3rz:
This should keep the ErrorHandler from showing as having a change in Chromatic.  See https://www.chromatic.com/docs/ignoring-elements

[Chromatic build](https://www.chromatic.com/build?appId=5ebad5c6b991d60022b1e72e&number=64) with 0 diffs reported


ash3rz:
Thanks for the reviews!

PR #120, opened 2020-06-09
VICE Admin Reorg and Analysis Mgmt

johnworth:
This PR moves the tables into separate tabs and adds action buttons to each analysis row that allows admins to extend the time limit, trigger file outputs, trigger file inputs, cancel/exit the analysis, and save outputs and exit the analysis.

Each action button should have a popper that provides feedback to the actions. The exit and save and exit actions will collapse the row, delete it, and then cause the set of rows to get refetched.

Additionally, the rows are now retrieved every 10 seconds, so newly launched analyses should show up in the table automatically without a full page refresh.

Screen Shot 2020-06-18 at 12 57 33 PM

Screen Shot 2020-06-18 at 12 57 59 PM

Screen Shot 2020-06-18 at 12 58 05 PM

Screen Shot 2020-06-18 at 12 58 31 PM

Screen Shot 2020-06-18 at 12 58 53 PM

### DIFF
--- a/src/components/vice/admin/index.js
+++ b/src/components/vice/admin/index.js
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useState, useEffect } from "react";
 
 import { useQuery, useMutation, queryCache } from "react-query";
 
@@ -217,10 +217,6 @@ const VICEAdminTabs = ({ data }) => {
     const tabPanelID = (name) => id(ids.ROOT, "admin", "tab-panels", name);
     const ariaControls = (name) => `vice-admin-tabs-${name}`;
 
-    const [analysisRows, setAnalysisRows] = useState(
-        data ? getAnalyses(data) : []
-    );
-
     const [mutantExit] = useMutation(exit, {
         onSuccess: () => queryCache.refetchQueries(getDataQueryName),
     });
@@ -236,6 +232,12 @@ const VICEAdminTabs = ({ data }) => {
     const [mutantDownloadInputs] = useMutation(downloadInputFiles, {
         onSuccess: () => queryCache.refetchQueries(getDataQueryName),
     });
+
+    const [analysisRows, setAnalysisRows] = useState([]);
+
+    useEffect(() => {
+        setAnalysisRows(getAnalyses(data));
+    }, [data]);
 
     const orderOfTabs = [
         "analyses",
@@ -327,7 +329,10 @@ const VICEAdminTabs = ({ data }) => {
 const VICEAdmin = () => {
     const classes = useStyles();
 
-    const { status, data, error } = useQuery(getDataQueryName, getData);
+    const { status, data, error } = useQuery(getDataQueryName, getData, {
+        refetchInterval: 10000,
+    });
+
     const isLoading = status === "loading";
     const hasErrored = status === "error";
 
--- a/src/components/vice/admin/index.js
+++ b/src/components/vice/admin/index.js
@@ -213,7 +213,9 @@ const VICEAdminTabs = ({ data }) => {
     const tabPanelID = (name) => id(ids.ROOT, "admin", "tab-panels", name);
     const ariaControls = (name) => `vice-admin-tabs-${name}`;
 
-    const analysisRows = data ? getAnalyses(data) : [];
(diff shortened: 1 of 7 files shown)


src/components/vice/admin/index.js
```diff
@@ -191,10 +210,133 @@ const VICEAdminSkeleton = () => {
...
+        </TabContext>
+    );
+};
+
 const VICEAdmin = () => {
     const classes = useStyles();
 
-    const { status, data, error } = useQuery("vice-admin", getData);
+    const { status, data, error } = useQuery(getDataQueryName, getData, {
+        refetchInterval: 10000,
```
sri2k1us, line 337: Minor. This can be moved to constant.

src/components/vice/admin/index.js
```diff
@@ -191,10 +210,133 @@ const VICEAdminSkeleton = () => {
...
+    });
+
     const isLoading = status === "loading";
```
sri2k1us, line 340: I think `loading` is defined as a constant in global constants file. We can use that here. Also we can define `error` as a global constant while we are at it.

johnworth: I'll add it to the component-specific constants file. I'm wary of global or widely-scoped values.

johnworth:
> 
> 
> LGTM. Some feedback from screenshots
> 
>     * May be just the active tab can be in primary to give better contrast. See app or data details for example.
> 
>     * It will be helpful to allow admins to copy App ID , External ID and User ID.

I'll work on the contrast in the tabs for this PR since it affects usability.

I'm gonna hold off on the copy buttons since they're gonna require a pretty major refactor and users can still highlight and copy the values manually.

johnworth:
Here's the new look for the tabs that I just uploaded.

Capture

src/components/vice/admin/index.js
```diff
@@ -191,12 +224,146 @@ const VICEAdminSkeleton = () => {
...
+                    scrollButtons="auto"
+                    indicatorColor="secondary"
+                >
+                    {orderOfTabs.map((tabName, index) => (
+                        <Tab
+                            label={msg(tabName)}
+                            id={tabID(tabName)}
+                            key={tabID(tabName)}
+                            value={`${index}`}
+                            aria-controls={ariaControls(tabName)}
```
ash3rz, line 286: Minor - everybody gets this part confused - but `aria-controls` is supposed to establish a relationship by ID, it's not a label. So the tab's value in `aria-controls` should match the ID of the tab panel it refers to.  It's not 100% clear in the MUI examples, but it is there.  I think in your case, this should be `tabPanelID(tabName)`

src/components/vice/admin/table/index.js
```diff
@@ -17,79 +17,181 @@ import {
...
-    dataEntryLabel: {
-        marginRight: theme.spacing(1),
-        fontWeight: 500,
-    },
-}));
-
-const ExtendedDataCard = ({ columns, row, collapseID }) => {
+const ActionButtonsSkeleton = () => {
+    return (
+        <>
```
ash3rz, line 44: Looks like this fragment isn't needed

src/components/vice/admin/table/index.js
```diff
@@ -17,79 +17,181 @@ import {
...
+    const { status, data, error } = useQuery(
+        ["async-data", row.externalID],
+        asyncData
+    );
+
+    const isLoading = status === "loading";
+    const hasErrored = status === "error";
+
+    if (hasErrored) {
+        console.log(error);
```
ash3rz, line 93: We should probably add in the same error handling we have in other components, using components from `/utils/error`

src/components/vice/admin/index.js
johnworth: Thanks for pointing this out, should be fixed now.

src/components/vice/admin/table/index.js
johnworth: Fixed, thanks.

src/components/vice/admin/table/index.js
johnworth: I'm going to be working on another set of features for the admin page soon, I'll include this in those PRs.

PR #121, opened 2020-06-13
CORE-984 Initial next.js page for displaying the App Launch view

psarando:
This PR will add the initial `next.js` page for displaying the App Launch view.
The app name click handler in the apps listing view will now route to this page.

This PR will also add an App Launch loading skeleton and loading error display, as well as a submission loading mask over only the stepper and bottom navigation buttons so that form data is not reset in case of an app launch (or quick launch save) submission error.

The the [CreateQuickLaunchDialog](https://github.com/cyverse-de/ui/blob/master/react-components/src/apps/quickLaunch/CreateQuickLaunchDialog.js) will also be imported and integrated with the launch form's `Save as Quick Launch` action.

The App Launch page displaying Step 1 for the DE Word Count app:
App Launch page - DE Word Count - Step 1

Step 1 for the DE Word Count app in mobile:
App Launch page - DE Word Count - Step 1 - mobile

The Create Quick Launch dialog in this page:
App Launch page - DE Word Count - Create Quick Launch dialog

The Create Quick Launch dialog on mobile:
App Launch page - DE Word Count - Create Quick Launch dialog - mobile

Note that if an error occurs while saving a Quick Launch, then only the error message is displayed in the dialog, but the user can close this dialog and open the error details dialog from the app launch form:
App Launch page - DE Word Count - Create Quick Launch dialog error

Create Quick Launch dialog error on mobile:
App Launch page - DE Word Count - Create Quick Launch dialog error - mobile

The App Launch page displaying a loading error:
App Launch page - loading error

The App Launch page displaying a loading error on mobile:
App Launch page - loading error - mobile

The App Launch page displaying a disabled app message:
App Launch page - disabled

The App Launch page displaying a disabled app message on mobile:
App Launch page - disabled - mobile

### TODO

The App Launch Form still needs to be populated with defaults from user preferences.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -12,5 +12,6 @@ module.exports = withBundleAnalyzer({
         INTERCOM_COMPANY_NAME: config.get("intercom.company_name"),
         ADMIN_GROUPS: config.get("admin.groups"),
         ADMIN_GROUP_ATTRIBUTE: config.get("admin.group_attribute_name"),
+        IRODS_HOME_PATH: config.get("irods.home_path"),
     },
 });
--- a/next.config.js
+++ b/next.config.js
@@ -4,6 +4,8 @@ const withBundleAnalyzer = require("@next/bundle-analyzer")({
     enabled: process.env.ANALYZE === "true",
 });
 
+const ONE_GiB = 2 ** 30;
+
 module.exports = withBundleAnalyzer({
     publicRuntimeConfig: {
         INTERCOM_APP_ID: config.get("intercom.app_id"),
@@ -15,14 +17,14 @@ module.exports = withBundleAnalyzer({
         IRODS_HOME_PATH: config.get("irods.home_path"),
         TOOLS_PRIVATE_MAX_CPU_LIMIT: config.has("tools.private.max_cpu_limit")
             ? config.get("tools.private.max_cpu_limit")
-            : null,
+            : 8,
         TOOLS_PRIVATE_MAX_MEMORY_LIMIT: config.has(
             "tools.private.max_memory_limit"
         )
             ? config.get("tools.private.max_memory_limit")
-            : null,
+            : 16 * ONE_GiB,
         TOOLS_PRIVATE_MAX_DISK_LIMIT: config.has("tools.private.max_disk_limit")
             ? config.get("tools.private.max_disk_limit")
-            : null,
+            : 512 * ONE_GiB,
     },
 });
--- a/next.config.js
+++ b/next.config.js
@@ -13,5 +13,16 @@ module.exports = withBundleAnalyzer({
         ADMIN_GROUPS: config.get("admin.groups"),
         ADMIN_GROUP_ATTRIBUTE: config.get("admin.group_attribute_name"),
         IRODS_HOME_PATH: config.get("irods.home_path"),
+        TOOLS_PRIVATE_MAX_CPU_LIMIT: config.has("tools.private.max_cpu_limit")
+            ? config.get("tools.private.max_cpu_limit")
+            : null,
+        TOOLS_PRIVATE_MAX_MEMORY_LIMIT: config.has(
+            "tools.private.max_memory_limit"
+        )
+            ? config.get("tools.private.max_memory_limit")
+            : null,
+        TOOLS_PRIVATE_MAX_DISK_LIMIT: config.has("tools.private.max_disk_limit")
+            ? config.get("tools.private.max_disk_limit")
+            : null,
     },
 });
--- a/src/components/apps/launch/AppLaunchFormSkeleton.js
+++ b/src/components/apps/launch/AppLaunchFormSkeleton.js
@@ -1,0 +1,51 @@
+/**
+ * @author psarando
+ *
+ * A Loading Mask for the App Launch Form.
+ */
+import React from "react";
+
+import {
+    BottomNavigation,
+    BottomNavigationAction,
+    Container,
+    Step,
+    Stepper,
+    StepButton,
+    TextField,
+    Typography,
+} from "@material-ui/core";
+import { Skeleton } from "@material-ui/lab";
+
+export default () => (
+    <>
+        <Skeleton variant="rect" width="100%">
+            <Stepper alternativeLabel nonLinear>
+                <Step>
+                    <StepButton>&nbsp;</StepButton>
+                </Step>
+            </Stepper>
+        </Skeleton>
+        <Container component="fieldset" maxWidth="md">
+            <Typography variant="h3" gutterBottom>
+                <Skeleton variant="text" />
+            </Typography>
+            <Typography variant="h3" gutterBottom>
+                <Skeleton variant="rect" width="100%">
+                    <TextField fullWidth rows={3} />
+                </Skeleton>
+            </Typography>
+            <Typography variant="h3" gutterBottom>
+                <Skeleton variant="text" />
+            </Typography>
+            <Typography variant="h3" gutterBottom>
+                <Skeleton variant="text" />
+            </Typography>
+        </Container>
+        <Skeleton variant="rect" width="100%">
+            <BottomNavigation>
+                <BottomNavigationAction />
+            </BottomNavigation>
+        </Skeleton>
+    </>
+);
--- a/src/components/apps/launch/AppLaunchFormSkeleton.js
+++ b/src/components/apps/launch/AppLaunchFormSkeleton.js
@@ -5,6 +5,8 @@
  */
 import React from "react";
 
+import styles from "./styles";
+
 import {
     BottomNavigation,
     BottomNavigationAction,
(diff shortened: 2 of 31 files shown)


src/components/apps/listing/TableView.js
```diff
@@ -207,6 +218,14 @@ function TableView(props) {
...
                                                 isDisabled={app.disabled}
                                                 name={app.name}
+                                                onAppNameClicked={() =>
+                                                    router.push(
+                                                        getAppLaunchPath(
+                                                            app.system_id,
+                                                            app.id
+                                                        )
+                                                    )
+                                                }
```
psarando, line 228: I added the `router` call here, since there's already a precedence for calling `router.push` in the analyses listing component, but I'm not sure if this is the right place for this call. Note that this will throw an exception in the listing stories for apps and analyses, since `router` is undefined there.

The alternative might be passing down a callback from the page component which can call `router.push` (similar to the new `launch` page in this PR).

src/pages/apps/[systemId]/[appId]/launch.js
```diff
@@ -0,0 +1,98 @@
...
+                submitAnalysis(submission)
+                    .then((resp) => {
+                        onSuccess(resp);
+                        router.push(`/${NavigationConstants.ANALYSES}`);
+                    })
+                    .catch((error) => {
+                        onError(error);
+                        setAppError(error);
+                    });
+            }}
```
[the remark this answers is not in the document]
Comment (sri2k1us) [reply src/pages/apps/[systemId]/[appId]/launch.js:89]: I haven't reviewed this PR. But I think we should use `react-query` here. Also, instead of navigating to `Analyses` listing here, we should navigate to job submission confirmation page. i.e `/apps/<id>/launch/submitted`. This page will show how many analyses are currently in submitted / running state along with their quota.
This page will provide users with option to 
- Go back to app listing 
- Launch another one of the same app 
- Goto analyses listing 
- Save this as quick launch. 
In future, this page can show live status of this analysis, load status of condor, k8, HPC OSG resources.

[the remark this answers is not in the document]
Comment (sri2k1us) [reply src/pages/apps/[systemId]/[appId]/launch.js:89]: Ohh all this is not needed for this PR. For now navigating to Analyses using `react-query` is good.

src/components/apps/listing/TableView.js
sri2k1us: I guess passing callback from the page should work. I can refactor analyses listing as well.

src/components/apps/launch/AppLaunchFormSkeleton.js
```diff
@@ -0,0 +1,74 @@
...
+    TextField,
+    Typography,
+    makeStyles,
+} from "@material-ui/core";
+import { Skeleton } from "@material-ui/lab";
+
+const useStyles = makeStyles(styles);
+
+export const StepperSkeleton = () => (
+    <Skeleton width="100%">
```
ash3rz, line 26: Minor, but a static ID somewhere on the loading skeleton is usually helpful for testing

src/components/apps/listing/TableView.js
ash3rz: I'm also for having the pages define the routing.  At least in the case for Data, it was necessary to make the component reusable not within a page (as in the data selection drawer).

src/pages/_app.js
```diff
@@ -108,6 +108,7 @@ function MyApp({ Component, pageProps, intercom, admin }) {
         intercom,
         intercom.appId,
         intercom.enabled,
+        irods,
```
ash3rz, line 111: Looks like GitHub actions is throwing a warning that `tools` is also missing as a dependency here.

stories/apps/launch/LoadingError.stories.js
```diff
@@ -0,0 +1,34 @@
...
+import { withKnobs, boolean } from "@storybook/addon-knobs";
+
+import AppLaunchStoryBase from "./AppLaunchStoryBase";
+
+const errorObject = {
+    response: {
+        status: 404,
+        data: {
+            error_code: "ERR_NOT_FOUND",
+            reason: "Something went arwy",
```
ash3rz, line 12: Minor tpyo?  `awry` 😄

src/pages/_app.js
psarando: Ah, good catch, thanks 👍

stories/apps/launch/LoadingError.stories.js
psarando: I uh... did that on purpose 👀 
It's even more arwy than awry 😆

src/components/utils/paths.js
```diff
@@ -0,0 +1,15 @@
...
+import NavigationConstants from "../../common/NavigationConstants";
+
+/**
+ * Builds a path to the App Launch Wizard for the app with the given IDs.
+ *
+ * @param {string} systemId The app's system ID.
+ * @param {string} appId The app's ID.
+ */
+export const getAppLaunchPath = (systemId, appId) =>
+    `/${NavigationConstants.APPS}/${systemId}/${appId}/launch`;
```
sri2k1us, line 15: This can be moved to`src/components/apps/utils.js`. We already have `utils.js` for analyses and data :)

src/server/configuration.js
```diff
@@ -238,3 +241,9 @@ export const amqpUri = config.get("amqp.amqp_uri");
  *
  */
 export const amqpExchangeName = config.get("amqp.exchange_name");
+
+/**
+ * Is irods home path
+ * @type {string}
+ */
+export const irodsHomePath = config.get("irods.home_path");
```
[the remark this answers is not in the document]
sri2k1us: I wonder if we really need this config at all. Right now, `bootstrap` returns a `data_info` property 

` data_info: {
        user_home_path: "/iplant/home/jmittelmeier",
        user_trash_path: "/iplant/trash/home/de-irods/jmittelmeier",
        base_trash_path: "/iplant/trash/home/de-irods",
    },`

Can we simply add irods home path to this?

next.config.js
```diff
@@ -12,5 +12,17 @@ module.exports = withBundleAnalyzer({
...
+            ? config.get("tools.private.max_cpu_limit")
+            : null,
+        TOOLS_PRIVATE_MAX_MEMORY_LIMIT: config.has(
+            "tools.private.max_memory_limit"
+        )
+            ? config.get("tools.private.max_memory_limit")
+            : null,
+        TOOLS_PRIVATE_MAX_DISK_LIMIT: config.has("tools.private.max_disk_limit")
+            ? config.get("tools.private.max_disk_limit")
+            : null,
```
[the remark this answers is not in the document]
psarando: I set these to `null` here so that the `ResourceRequirements` component in the app launch form will use its own reasonable defaults:
https://github.com/cyverse-de/sonora/blob/0c60284dba2e9e3c328e6db4b25b09776243f594/src/components/apps/launch/ResourceRequirements.js#L97-L111

I can refactor those defaults so they're set here instead, which is probably better.

src/server/configuration.js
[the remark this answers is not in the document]
psarando: This will be required when the `data-info` service is down, or maybe if `bootstrap` hasn't responded yet. We use it in the current UI this way:
https://github.com/cyverse-de/ui/blob/58224562a7b44d4900cea6b5b195a5bd1fab128c/de-lib/src/main/java/org/iplantc/de/client/models/UserInfo.java#L201-L216

So it appears the `launch` page is now imitating the logic of that `UserInfo` model, and constructing a reasonable assumption of the user's home folder, but it's only using it to construct an assumption of the user's default output folder.

Ultimately the user's default output folder should come from their preferences, and their home folder will come from the `bootstrap` response, but whatever will be responsible for storing those preferences or bootstrap values might also need this config to manually create a default home folder in case a service can't be reached?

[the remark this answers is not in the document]
sri2k1us: Sounds good. Can we launch a job when data-info is down ? I was under the assumption we could not.

I agree - users default folder should come from user preferences / bootstrap. I know my PR for preferences is not ready yet but, I am prefetching and caching bootstrap info. In case of failure, there will be retries as well. So if we don't have bootstrap info in our code, then bootstrap service is really down. And in that case, can `apps/analyses` service get the users home folder / default output folder from `data-info `. Does `launch` service expects the output folder set when launching a job or can it be empty?

src/pages/apps/[systemId]/[appId]/launch.js
```diff
@@ -53,6 +56,25 @@ export default () => {
...
+        ({ submission }) => submitAnalysis(submission),
+        {
+            onSuccess: (resp, { onSuccess }) => {
+                // return a promise so that mutate only resolves after this callback
+                return queryCache
+                    .refetchQueries(getAnalysesQueryKey)
+                    .then(() => {
+                        router.push(`/${NavigationConstants.ANALYSES}`);
+                        onSuccess(resp);
+                    });
```
Comment (sri2k1us) [inline src/pages/apps/[systemId]/[appId]/launch.js:69]: Do we have refetch analyses here ? Won't get get fetched by default when analyses page is loaded ?

next.config.js
```diff
@@ -15,14 +17,14 @@ module.exports = withBundleAnalyzer({
...
         )
             ? config.get("tools.private.max_memory_limit")
-            : null,
+            : 16 * ONE_GiB,
         TOOLS_PRIVATE_MAX_DISK_LIMIT: config.has("tools.private.max_disk_limit")
             ? config.get("tools.private.max_disk_limit")
-            : null,
+            : 512 * ONE_GiB,
     },
 });
```
sri2k1us, line 30: 👍

src/server/configuration.js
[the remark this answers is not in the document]
sri2k1us: We can move this to a design meeting. I have few more thoughts / questions about apps, app launch and analyses.

src/pages/apps/[systemId]/[appId]/launch.js
Comment (psarando) [reply src/pages/apps/[systemId]/[appId]/launch.js:69]: I'm not that familiar with `react-query` yet, so I don't really know. I tried asking in a previous comment (https://github.com/cyverse-de/sonora/pull/121#discussion_r439698956) but didn't get a specific answer.

I can try removing the refetch and see what happens 🤷‍♂️

Comment (sri2k1us) [reply src/pages/apps/[systemId]/[appId]/launch.js:69]: Oops. Sorry :) My guess and based on my understanding this `refetch` call is not needed. Lets see how wrong I am 😜

src/server/configuration.js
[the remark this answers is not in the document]
psarando: We can move this to a design meeting, but a few points to keep in mind for the meeting:

I wasn't intending on keeping this logic of building a default output folder in this page, it's just here until it can be grabbed from the `bootstrap` code. I was defending keeping this config, though, since this config and the logic that uses it was added as part of the UI resiliency epic (cyverse-de/ui#50).

> Sounds good. Can we launch a job when data-info is down ? I was under the assumption we could not.
> 

No, I don't think we can, but I think there's a possibility that `data-info` doesn't respond for the `bootstrap` call, but is available again once the user is ready to launch the analysis.

> So if we don't have bootstrap info in our code, then bootstrap service is really down. And in that case, can `apps/analyses` service get the users home folder / default output folder from `data-info `? Does `launch` service expects the output folder set when launching a job or can it be empty?

Well the `bootstrap` endpoint is in `terrain`, so if the entire `bootstrap` endpoint is down then nothing works 😉 
However, the `bootstrap` endpoint gets the path info from `data-info`, but it can still get as much info as possible if `data-info` is down, and only in that case will the home and trash paths be missing for the UI.

Finally, there is a known issue with an analysis output folder that I think is also an issue in the current UI. If the default output folder is not available when the launch form is created, and the user picks an output folder, then this form will assume that the user wants to use that exact folder for outputs, and the job won't automatically create a subfolder under it:
https://github.com/cyverse-de/sonora/blob/0c60284dba2e9e3c328e6db4b25b09776243f594/src/components/apps/launch/AppLaunchForm.js#L211

This means if there is already a file under that folder that has the same name as an output file, it will be overwritten by this job.

Again, this form is imitating the logic of the current UI, but we can add a ticket for this issue and fix it in another PR if necessary (I don't know the history behind the decision to hide the `create_output_subdir` flag from the user).

src/pages/apps/[systemId]/[appId]/launch.js
Comment (sri2k1us) [reply src/pages/apps/[systemId]/[appId]/launch.js:69]: Did this work ?

src/server/configuration.js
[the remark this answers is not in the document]
sri2k1us: Thanks for detailed response. I guess what have I in my mind is new approach. I will try elaborate more on this in the meeting.

src/pages/apps/[systemId]/[appId]/launch.js
Comment (psarando) [reply src/pages/apps/[systemId]/[appId]/launch.js:69]: Yep, looks like it works 👍

src/pages/apps/[systemId]/[appId]/launch.js
```diff
@@ -67,6 +68,20 @@ export default () => {
         }
     );
 
+    const [addQuickLaunchMutation] = useMutation(
+        ({ quickLaunch }) => addQuickLaunch(quickLaunch),
+        {
+            onSuccess: (resp, { onSuccess }) => {
+                // TODO route to app details or QL listing page?
```
Comment (sri2k1us) [inline src/pages/apps/[systemId]/[appId]/launch.js:75]: 👍

sri2k1us:
LGTM 👍

psarando:
Thanks for the reviews! 👍

PR #122, opened 2020-06-18
User Preferences

sri2k1us:
- Basic User Preferences
- Options to reset to defaults or update

Notes:
- Default folder validation is not done with Formik. I spent too much time to figure this out. @ash3rz helped too. But I still saw exceptions randomly. So I am using Material UI error prop with state.

- Keyboard shortcut preferences are not implemented. i.e For example, Hitting Control + Shift + A wont open apps yet.

- General preferences are not implemented. i.e For example, Remember last path won't work

- Webhooks preferences UI not implemented yet.

image

image

image


src/components/layout/CyVerseAppBar.js
```diff
@@ -260,8 +266,38 @@ function CyverseAppBar(props) {
...
+                            30 * 1000
+                        ),
+                    onError: (e) => {
+                        showErrorAnnouncer(
+                            formatMessage(intl, "bootstrapError"),
+                            e
+                        );
+                    },
+                },
+            });
```
sri2k1us, line 290: Prefetch bootstrap and cache it. If error, this will retry 3 times.

src/components/layout/CyVerseAppBar.js
```diff
@@ -446,31 +482,33 @@ function CyverseAppBar(props) {
...
-                    className={
-                        activeView === NavigationConstants.SETTINGS
-                            ? classes.listItemActive
-                            : classes.listItem
-                    }
+            {userProfile?.id && (
+                <Tooltip
+                    title={formatMessage(intl, "settings")}
+                    placement="right"
+                    arrow
```
sri2k1us, line 489: Hide preferences Side Menu option when not logged in.

src/components/preferences/General.js
```diff
@@ -0,0 +1,274 @@
...
+                    acceptedType={ResourceTypes.FOLDER}
+                    onConfirm={(selection) => {
+                        setOpenFileBrowser(false);
+                        onNewDefaultOutputFolder(selection);
+                    }}
+                    baseId={baseId}
+                    multiSelect={false}
+                />
+            )}
+        </>
```
sri2k1us, line 270: Without the `defaultOutputFolder` condition, for some weird reason, the `SelectionDrawer` will always open the users home folder and ignore the `startingPath`. I haven't debugged it entirely yet.

stories/preferences/Preferences.stories.js
```diff
@@ -0,0 +1,87 @@
...
+        defaultFileSelectorPath: "/iplant/home/jmittelmeier",
+        closeKBShortcut: "Q",
+        appsKBShortcut: "A",
+        system_default_output_dir: {
+            id: "/iplant/home/jmittelmeier/analyses_qa",
+            path: "/iplant/home/jmittelmeier/analyses_qa",
+        },
+        default_output_folder: {
+            id: "/iplant/home/jmittelmeier/analyses_qa",
+            path: "/iplant/home/jmittelmeier/analyses_qa",
```
sri2k1us, line 69: I am setting the path value to both `path` and `id`. Selection Drawer needs to be  modified to return the UUID before it can be set here.

src/components/preferences/Preferences.js
```diff
@@ -0,0 +1,463 @@
...
+    useEffect(() => {
+        //get from cache if not fetch now.
+        const ip = userProfile?.attributes.ip;
+        const prefCache = queryCache.getQueryData([BOOTSTRAP_KEY, { ip }]);
+        if (prefCache) {
+            preProcessData(prefCache);
+        } else {
+            setBootstrapKey([BOOTSTRAP_KEY, { ip }]);
+        }
+    }, [userPref, userProfile]);
```
sri2k1us, line 108: Fetch back cached bootstrap info. If not, call bootstrap service again and cache the result.

src/components/preferences/Preferences.js
```diff
@@ -0,0 +1,463 @@
...
+                        return bootstrapData;
+                    }
+                });
+            },
+            onError: (e) => {
+                showErrorAnnouncer(formatMessage(intl, "savePrefError"), e);
+            },
+        }
+    );
+
```
sri2k1us, line 164: Once preferences is updated by user, update the cache with new values.

src/components/preferences/Preferences.js
```diff
@@ -0,0 +1,466 @@
...
+                            return bootstrapData;
+                        }
+                    }
+                );
+            },
+            onError: (e) => {
+                showErrorAnnouncer(formatMessage(intl, "savePrefError"), e);
+            },
+        }
+    );
```
sri2k1us, line 166: Update cache once preferences is saved.

src/components/preferences/Preferences.js
```diff
@@ -0,0 +1,466 @@
...
+            prefConstants.keys.DEFAULT_OUTPUT_FOLDER,
+            userPref.system_default_output_dir.path
+        );
+        setShowRestoreConfirmation(false);
+    };
+
+    const setDefaultFolder = (setFieldValue, newFolder) => {
+        setFieldValue(prefConstants.keys.DEFAULT_OUTPUT_FOLDER, newFolder);
+        setDefaultOutputFolder(newFolder);
+    };
```
psarando, line 279: I tried looking at the `react-query` docs to see if a query can be used as a promise, since that's the only way to do async validation with Formik, but I didn't find a way.

I'm not sure if you tried something like this, but maybe one way to let Formik handle validation, instead of manually setting it with state, is call `setFetchDetailsKey` here, instead of calling `setFieldValue` right away.

Then in the  `onSuccess` callback of that query, the entire folder details object can be set with `setFieldValue`. Once that is called it will trigger validation, and then inside `validateShortCuts` (which should be renamed) we can check for write permissions and set an error message if necessary.

This would also require using `defaultOutputFolder.path` for the `value` of the text field for display in the form, and also in `handleSubmit`.

src/server/auth.js
```diff
@@ -26,6 +27,9 @@ export const getUserID = (req) => {
  */
 export const getUserProfile = (req) => {
     const accessToken = req?.kauth?.grant?.access_token;
+    logger.info(
+        "####### Client IP ####### => " + req?.connection?.remoteAddress
+    );
```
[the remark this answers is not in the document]
sri2k1us: @slr71 and I decided to leave it for some time to see if it get the correct ip once it deployed with K8s.

src/components/preferences/Preferences.js
sri2k1us: Good Idea. Let me give this a shot!

sri2k1us: Welp. Doesn't seem to work. I also see roadblock with this approach. For example, when I try to restore defaults, that causes exceptions.
`A component is changing an uncontrolled input of type text to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa)` 

May be we can revisit this when we add `webhooks` preferences.

src/components/layout/messages.js
```diff
@@ -5,6 +5,7 @@ const intlData = {
         all: "All",
         analyses: "Analyses",
         apps: "Apps",
+        bootstrapError: "Unable to get bootstrap info.",
```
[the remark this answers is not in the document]
sri2k1us: Make sense. I will refactor.

src/components/preferences/Preferences.js
psarando: When you got the "uncontrolled input" warning, were you still defaulting the `value` of the text field to an empty string if there is no folder path?
`value={defaultOutputFolder?.path || ""}`

sri2k1us: Yes.

src/components/preferences/General.js
```diff
@@ -0,0 +1,274 @@
...
+                        />
+                        <Button
+                            color="primary"
+                            id={build(baseId, ids.BROWSE_BUTTON)}
+                            className={classes.browseButton}
+                            onClick={() => setOpenFileBrowser(true)}
+                            variant="outlined"
+                        >
+                            {getMessage("browse")}
+                        </Button>
```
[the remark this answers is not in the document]
sri2k1us: Just realized, we will need to a new page like `error.js` to get this done. But I think we decided in one of our meetings that we need it anyway. So will go ahead and add it.

src/components/layout/messages.js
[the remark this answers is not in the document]
sri2k1us: Just realized, we will need to add a new page like `error.js` to get this done. But I think we decided in one of our meetings that we need it anyway. So will go ahead and add it.

src/components/apps/launch/AppLaunchForm.js
```diff
@@ -374,11 +375,20 @@ const AppLaunchForm = (props) => {
...
+        label: formatMessage(intl, "parameters"),
+        step: 1,
+    };
+    const stepAdvanced = {
+        label: formatMessage(intl, "advancedSettings"),
+        step: 2,
+    };
     const stepReviewAndLaunch = {
-        label: getMessage("reviewAndLaunch"),
+        label: formatMessage(intl, "reviewAndLaunch"),
```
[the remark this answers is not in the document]
sri2k1us: I just checked it again. Looks like I still get the warning about same key being reused.

src/components/preferences/Shortcuts.js
```diff
@@ -0,0 +1,118 @@
...
+    const { baseId } = props;
+    const classes = useStyles();
+    return (
+        <>
+            <Typography variant="h6" className={classes.sectionHeader}>
+                {getMessage("keyboardShortcutHeaderLbl")}
+            </Typography>
+            <Grid container spacing={2} className={classes.grid}>
+                <GridLabelValue
+                    label={getMessage("appsViewShortcutLbl")}
```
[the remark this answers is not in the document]
sri2k1us: am not sure if understand your suggestion here. Can you please elaborate ?

src/components/apps/launch/AppLaunchForm.js
```diff
@@ -390,6 +390,7 @@ const AppLaunchForm = (props) => {
         saveQuickLaunch,
         startingPath,
         submitAnalysis,
+        intl,
```
sri2k1us, line 393: Missed it during rebase. Added it back.

src/components/layout/CyVerseAppBar.js
```diff
@@ -242,10 +244,42 @@ function CyverseAppBar(props) {
...
+            retry: 1,
+            //copied from react-query doc. Add exponential delay for retry.
+            retryDelay: (attempt) =>
+                Math.min(attempt > 1 ? 2 ** attempt * 1000 : 1000, 30 * 1000),
+            onError: (e) => {
+                console.log("error on bootstrap=>" + e);
+                setBootstrapError(e);
+            },
+        },
+    });
```
sri2k1us: `prefetchQuery` was not invoking `onError` for some reason so I switched it back to `useQuery`. But I didn't notice any operational or performance differences.

src/components/utils/error/ErrorHandler.js
```diff
@@ -54,9 +55,15 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 function ClientInfo(props) {
+    const [browser, setBrowser] = useState();
```
sri2k1us, line 58: Since this is now in a page, access to `window` is only available from `useEffect`

src/components/preferences/Preferences.js
```diff
@@ -122,13 +125,23 @@ function Preferences(props) {
...
+    useEffect(() => {
+        if (bootstrapError) {
+            const errorString = JSON.stringify(bootstrapError);
+            setBootstrapError(null);
+            router.push(
+                `/${NavigationConstants.ERROR}?errorInfo=` + errorString
+            );
+        }
+    }, [bootstrapError, router]);
+
```
sri2k1us, line 137: Serialize error object and pass it as query string to `error` page.

src/pages/error.js
```diff
@@ -0,0 +1,17 @@
...
+ */
+import React from "react";
+import { useRouter } from "next/router";
+import ErrorHandler from "../components/utils/error/ErrorHandler";
+
+export default function Error() {
+    const router = useRouter();
+    const errorInfo = router.query?.errorInfo;
+    const errorObj = errorInfo ? JSON.parse(errorInfo) : null;
+    return <ErrorHandler errorObject={errorObj} baseId="error" />;
```
sri2k1us, line 16: Parse `error` string into `errorObject`. `Nextjs` takes care of encoding and decoding.

src/components/layout/CyVerseAppBar.js
```diff
@@ -242,10 +244,42 @@ function CyverseAppBar(props) {
...
     }, [clientConfig, setConfig]);
 
+    useEffect(() => {
+        if (bootstrapError) {
+            const errorString = JSON.stringify(bootstrapError);
+            setBootstrapError(null);
+            router.push("/error?errorInfo=" + errorString);
+        }
+    }, [bootstrapError, router]);
+
```
sri2k1us: Serialize error object to error string.

src/components/utils/error/ErrorHandler.js
```diff
@@ -54,9 +55,15 @@ const useStyles = makeStyles((theme) => ({
...
+    useEffect(() => {
+        setBrowser(Bowser.getParser(window.navigator.userAgent));
+    }, []);
```
sri2k1us, line 61: Since this is now in a page, access to window is only available from useEffect

src/components/preferences/Shortcuts.js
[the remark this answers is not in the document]
psarando: Originally I was thinking of using a `label={getMessage(...)}` prop on the on the `FormTextField` instead of using a `GridLabelValue`, but the layout isn't the same.

I was also trying to point out the benefit of using the `label` prop on MUI text fields: if you inspect elements of a MUI text field with this prop set, then you will see that they use a `label` element with a `for` attribute, which should let assistive tech know that's the label for the input field.

This is also what I was getting at with my example in https://github.com/cyverse-de/sonora/pull/122#discussion_r445277180

I thought there were other aria attributes involved, but I might have been thinking of this example instead: https://material-ui.com/api/form-control/

src/components/utils/GridLabelValue.js
```diff
@@ -10,17 +10,14 @@ import { Grid, Typography } from "@material-ui/core";
...
 // space than the first column, labels
-export default function GridLabelValue(props) {
-    const { label, children } = props;
-
+export default function GridLabelValue({ label, children, ...props }) {
     return (
         <>
             <Grid item sm={3} xs={6} md={3} lg={3}>
-                <Typography component="span" variant="subtitle2">
+                <Typography component="span" {...props}>
```
psarando, line 17: Do we still want to leave `variant="subtitle2"` as a default setting? It can still be overridden by a `variant` prop passed by a parent component.
```suggestion
                <Typography component="span" variant="subtitle2" {...props}>
```

src/components/layout/messages.js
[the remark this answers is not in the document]
psarando: Looks like this message can be removed now.

src/components/preferences/messages.js
```diff
@@ -0,0 +1,41 @@
+export default {
+    locales: "en-us",
+    messages: {
+        analysesViewShortcutLbl: "Open analyses view",
+        analysisStatusEmailLbl: "Email me when my analysis status changes",
+        appsViewShortcutLbl: "Open apps view",
+        bootstrapError: "Unable to get bootstrap info.",
```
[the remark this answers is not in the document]
psarando: Looks like this message can be removed now.

src/components/utils/GridLabelValue.js
sri2k1us: Makes sense.

sri2k1us:
Thanks for the reviews.

PR #123, opened 2020-06-20
CORE-1028: allow admins to impersonate other users.

slr71:
I still need to test this, but I thought I'd submit a draft PR in case anyone wants to take a look.

### DIFF
--- a/src/server/api/fileio.js
+++ b/src/server/api/fileio.js
@@ -18,10 +18,23 @@ export default function fileIORouter() {
 
     logger.info("************ Adding File I/O handlers **********");
 
-    logger.info("adding the /api/upload handler ");
+    logger.info("adding the POST /api/upload handler");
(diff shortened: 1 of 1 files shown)


src/server/auth.js
```diff
@@ -4,19 +4,117 @@
...
+};
+
+/**
+ * Stops impersonation.
+ * @param {Object} req
+ * @param {Object} res
+ * @param {Function} next
+ */
+export const stopImpersonation = (req, res, next) => {
+    delete req.session[IMPERSONATION_TOKEN_KEY];
```
ash3rz, line 116: Not really a comment on your PR, but I had to look this up.  I had no idea a simple `delete` existed :smile: The syntax looks so strange to me, I thought it was pseudo code

slr71:
I moved this to another branch so that I could work on something else. I'll submit another PR when we have things working better.

PR #124, opened 2020-06-24
Analysis and Data hot fix

sri2k1us:
- Fix NPE/Undefined error when notifications were processed in analyses view.
- Check if the notification is of type `analyses` before processing notification in analyses view.
- Hide upload options when in read-only folder.
- Add feedback (circular progress) when a user clicks on `Import` button
- Fix row level dot-menu for URL import, Local import and Upload tracking dialog.

Note:
We can just push https://github.com/cyverse-de/sonora/commit/7c4b28e58cdb844aa537dcac8444cd73d950c0b2 if anyone is not comfortable with fixes for data menu

### DIFF
--- a/src/components/analyses/listing/Listing.js
+++ b/src/components/analyses/listing/Listing.js
@@ -143,10 +143,11 @@ function Listing(props) {
                 const category = message.type;
                 if (
                     category?.toLowerCase() ===
-                    constants.NOTIFICATION_CATEGORY.ANALYSIS.toLowerCase()
+                        constants.NOTIFICATION_CATEGORY.ANALYSIS.toLowerCase() &&
+                    data
                 ) {
                     const analysisStatus = message.payload.status;
-                    const found = data?.analyses?.find(
+                    const found = data.analyses?.find(
                         (analysis) => analysis.id === message.payload.id
                     );
 
@@ -154,12 +155,12 @@ function Listing(props) {
                         if (analysisStatus !== found.status) {
                             found.status = analysisStatus;
                             found.enddate = message.payload.enddate;
-                            setData({ analyses: [...data?.analyses] });
+                            setData({ analyses: [...data.analyses] });
                         }
                     } else {
                         //add a new analysis record and remove the last record from the page
                         //to maintain page size
-                        if (data?.analyses.length === rowsPerPage) {
+                        if (data.analyses?.length === rowsPerPage) {
                             const newPage = data.analyses.slice(
                                 0,
                                 data.analyses.length - 1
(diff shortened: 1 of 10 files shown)


src/components/data/listing/TableView.js
```diff
@@ -254,182 +280,222 @@ function TableView(props) {
...
+            </TableContainer>
+            <FileBrowser
+                id={localUploadId}
+                handleUploadFiles={handleUploadFiles}
+            />
+            <UploadDialog
+                open={uploadDialogOpen}
+                handleClose={() => setUploadDialogOpen(false)}
+            />
+            <URLImportDialog
```
ash3rz, line 493: If I'm understanding right, we've basically moved up the responsibility for these upload and import dialogs here, which is good.  But I think that also means they can be removed from the render functions in wherever else they were originally (I think `Toolbar` and maybe also `UploadMenuBtn`)?  I may have missed something.

sri2k1us: Hmm. Honestly that didn't cross my mind. Although when I Look at the code, I still think it is needed at both the places. `Toolbar` has its own dot menu and table view has its own dot menu.

src/components/analyses/listing/Listing.js
```diff
@@ -141,40 +141,40 @@ function Listing(props) {
...
+                                analyses: [message.payload, ...newPage],
+                            });
+                        } else if (data?.analyses.length === 0) {
+                            //if page is empty...
+                            setData({ analyses: [message.payload] });
+                        } else {
+                            setData({
+                                analyses: [message.payload, ...data?.analyses],
+                            });
+                        }
```
psarando, line 177: I think this can still throw an error.

When this listing is initially rendered, the `data` state is `null`, then the listing is fetched with a query; but if an analysis notification arrives while the listing is still loading, then `data` will still be `null` here, and the "uniterable" error will be thrown.

Another `if` check might prevent it (here or above when checking `category`):
```suggestion
                        } else if (data) {
                            setData({
                                analyses: [message.payload, ...data?.analyses],
                            });
                        }
```

sri2k1us: 🤦 I had this change but removed to test one last time but I never put it back and then got distracted by data view issues. I should go to sleep!

sri2k1us: But that why we have code reviews 👍 (To save people like me.)

src/components/data/listing/TableView.js
psarando: Yeah, I don't see a reference in here to the toolbar, however the `Listing` component uses both the toolbar and this `TableView`. So maybe it's possible to pull up only 1 definition of each of these dialogs and the browser into the `Listing` component, and then the toolbar and this table view can make callbacks up to the `Listing` component to open the dialogs?

sri2k1us:
Thanks for the reviews.

PR #125, opened 2020-06-29
Refactor VICE admin UI to use react-table

johnworth:
Refactors the VICE admin UI to use react-table. The major improvement here is the standardized state management API, which will enable future support for features such as:
- per-column filters
- global filters
- reordering rows
- resizing rows

Right now I'm only really taking advantage of the hidden columns and row expansion features, which are used to push columns into the collapsed part of each row when the screen becomes too small to display them.

Previously, this PR was being used to add bulk actions to the VICE admin UI, but it was growing too large to be easily reviewed. I refocused the PR on adding react-table and supporting the existing feature set to keep it from getting overwhelming. The bulk actions feature is still planned.

### Unexpanded
image
### Expanded
image
### Sorted by username
image

### Unexpanded
image
### Expanded
image
### Sorted by username
image

### Unexpanded
image
### Expanded
Without action buttons showing
image
With action buttons showing
image
### Sorted by username
image

### DIFF
--- a/src/components/vice/admin/table/actionButtons.js
+++ b/src/components/vice/admin/table/actionButtons.js
@@ -54,10 +54,14 @@ export default ({
     const classes = useStyles();
     const [open, setOpen] = useState(false);
 
+    const externalID = row.original.externalID;
+
     const { status, data, error } = useQuery(
-        ["async-data", row.externalID],
+        ["async-data", externalID],
         asyncData
     );
+
+    console.log(row);
 
     const isLoading = status === "loading";
     const hasErrored = status === "error";
@@ -70,8 +74,10 @@ export default ({
         let tlErr;
         let tlData;
 
+        console.log(data);
+
         try {
-            tlData = dataFn(data.analysisID, row.externalID);
+            tlData = dataFn(data.analysisID, externalID);
         } catch (err) {
             tlErr = err;
         }
@@ -80,6 +86,8 @@ export default ({
             ? AnnouncerConstants.ERROR
             : AnnouncerConstants.SUCCESS;
         const text = tlErr ? tlErr.message : msg(msgKey);
+
+        console.log(`text: ${text}    variant: ${variant}`);
 
         announce({ text, variant });
 
@@ -102,7 +110,7 @@ export default ({
             ) : (
                 <>
                     <ActionButton
-                        baseID={row.externalID}
+                        baseID={externalID}
                         name="extendTimeLimit"
                         handler={handleExtendTimeLimit}
                         popperMsgKey="timeLimitExtended"
@@ -110,7 +118,7 @@ export default ({
                     />
 
                     <ActionButton
-                        baseID={row.externalID}
+                        baseID={externalID}
                         name="downloadInputs"
                         handler={handleDownloadInputs}
                         popperMsgKey="downloadInputsCommandSent"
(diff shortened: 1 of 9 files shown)


src/components/vice/admin/table/actionButtons.js
```diff
@@ -0,0 +1,151 @@
...
+        } catch (err) {
+            tlErr = err;
+        }
+
+        setAnchorEl(event.currentTarget);
+        setPopperMessage(tlErr ? tlErr.message : msg(msgKey));
+        setOpen(true);
+
+        return tlData;
+    };
```
sri2k1us, line 80: Instead of the using the `Popper` component, we can use the `annoucer` that shows a `toast` message.
`import {
    announce,
    AnnouncerConstants,
} from "@cyverse-de/ui-lib";`
`announce({
                text: <errorMessage>,
                variant: AnnouncerConstants.ERROR,
 });`

sri2k1us:
For others who are new to `react-table` - this example is good to review since it documents the entire capability - https://codesandbox.io/s/github/tannerlinsley/react-table/tree/master/examples/material-UI-kitchen-sink?file=/src/components/EnhancedTable.js

src/components/vice/admin/table/actionButtons.js
```diff
@@ -0,0 +1,151 @@
...
+                setOpen(false);
+            }
+        }, 3000);
+        return () => clearInterval(timerID);
+    });
+
+    return (
+        <div className={classes.actions}>
+            {isLoading ? (
+                <ActionButtonsSkeleton />
```
ash3rz, line 94: An ID here would be helpful

src/components/vice/admin/table/index.js
```diff
@@ -14,169 +9,28 @@ import {
...
     useMediaQuery,
-    Button,
-    Popper,
+    IconButton,
+    TableSortLabel,
 } from "@material-ui/core";
 
-import { Skeleton } from "@material-ui/lab";
-
+// import { Skeleton } from "@material-ui/lab";
```
ash3rz, line 22: Looks like this can be removed

src/components/vice/admin/table/actionButtons.js
johnworth: Alrighty, this should be switched over to an Announcer. The switch over also simplified the code a good bit and exposed some i18n issues lurking about, so thanks for the suggestion.

src/components/vice/admin/table/index.js
johnworth: Done, thanks.

src/components/vice/admin/table/actionButtons.js
johnworth: Done, thanks.

src/components/vice/admin/table/actionButtons.js
```diff
@@ -0,0 +1,151 @@
...
+    const { status, data, error } = useQuery(
+        ["async-data", row.externalID],
+        asyncData
+    );
+
+    const isLoading = status === "loading";
+    const hasErrored = status === "error";
+
+    if (hasErrored) {
+        console.log(error);
```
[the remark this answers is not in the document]
johnworth: Took a little of bit of poking around, but I finally got the announcer to work here. I also changed things so that the action buttons don't get rendered if the request to get async data about the analysis fails. Also, each button can will cause an announcer to display, one with a message telling you that the action was successful or one to display an error message.

PR #126, opened 2020-06-30
Selection drawer updates

sri2k1us:
- Make the drawer not scroll
- Add scrolling to data table
- Fix selection drawer typography

image

image

### DIFF
--- a/src/components/data/SelectionDrawer.js
+++ b/src/components/data/SelectionDrawer.js
@@ -5,7 +5,7 @@
  * resources as input to other components.
  */
 
-import React, { useCallback, useEffect, useRef, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 
 import { build, getMessage, withI18N } from "@cyverse-de/ui-lib";
 import {
@@ -47,7 +47,7 @@ function SelectionToolbar(props) {
         onClose,
         acceptedType,
         multiSelect,
-        getToolbarHeight,
+        setToolbarRef,
     } = props;
     const theme = useTheme();
     const classes = useStyles();
@@ -57,12 +57,8 @@ function SelectionToolbar(props) {
 
     // useEffect will run on toolbarRef value assignment
     useEffect(() => {
-        // The 'current' property contains info of the reference:
-        // align, title, ... , width, height, etc.
-        if (toolbarRef.current && getToolbarHeight) {
-            getToolbarHeight(toolbarRef.current.offsetHeight);
-        }
-    }, [toolbarRef, getToolbarHeight]);
+        setToolbarRef(toolbarRef);
+    }, [toolbarRef, setToolbarRef]);
 
     useEffect(() => {
         setDisplayPath(currentPath);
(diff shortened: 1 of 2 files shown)


src/components/data/SelectionDrawer.js
```diff
@@ -170,6 +201,16 @@ function SelectionDrawer(props) {
...
 
+    const [toolbarHeight, setToolbarHeight] = useState(0);
+    const [appBarHeight] = useComponentHeight();
+
+    const toolbarHeightCallback = useCallback(
+        (height) => {
+            setToolbarHeight(height);
+        },
+        [setToolbarHeight]
+    );
```
ash3rz, line 212: I believe this could be simplified since `useComponentHeight` does all the calculating for you.  So this `useState` and `useCallback` could turn into:
```
const [toolbarHeight, setToolbarRef] = useComponentHeight();
```

src/components/data/SelectionDrawer.js
```diff
@@ -181,26 +222,33 @@ function SelectionDrawer(props) {
...
+                            baseId={id}
+                            currentPath={currentPath}
+                            setCurrentPath={setCurrentPath}
+                            selectedTotal={selectedTotal}
+                            getSelectedResources={getSelectedResources}
+                            acceptedType={acceptedType}
+                            multiSelect={multiSelect}
+                            onConfirm={onConfirm}
+                            onClose={onClose}
+                            getToolbarHeight={toolbarHeightCallback}
```
ash3rz, line 247: Then, this prop could be replaced with `setToolbarRef={setToolbarRef}`

src/components/data/SelectionDrawer.js
```diff
@@ -40,10 +47,23 @@ function SelectionToolbar(props) {
...
+    const toolbarRef = useRef(null);
+
+    // useEffect will run on toolbarRef value assignment
+    useEffect(() => {
+        // The 'current' property contains info of the reference:
+        // align, title, ... , width, height, etc.
+        if (toolbarRef.current && getToolbarHeight) {
+            getToolbarHeight(toolbarRef.current.offsetHeight);
+        }
+    }, [toolbarRef, getToolbarHeight]);
```
ash3rz, line 65: And finally, this `useEffect` could be replaced with `setToolbarRef(toolbarRef)`

src/components/data/SelectionDrawer.js
sri2k1us: This worked out. I still had to add offset for pagination toolbar but :thum

ash3rz: Ah right, still need the pagination height factored in - good catch 👍

src/components/data/SelectionDrawer.js
```diff
@@ -57,12 +57,8 @@ function SelectionToolbar(props) {
...
     // useEffect will run on toolbarRef value assignment
     useEffect(() => {
-        // The 'current' property contains info of the reference:
-        // align, title, ... , width, height, etc.
-        if (toolbarRef.current && getToolbarHeight) {
-            getToolbarHeight(toolbarRef.current.offsetHeight);
-        }
-    }, [toolbarRef, getToolbarHeight]);
+        setToolbarRef(toolbarRef);
+    }, [toolbarRef, setToolbarRef]);
```
ash3rz, line 61: I think this can actually just be `setToolbarRef(toolbarRef);` without the `useEffect` because `useComponentHeight` is already tracking the ref with its own `useEffect`.  Small thing though

sri2k1us: It works without `useEffect` but throws a big stack trace in browser console the first time you open the Selection Drawer. So I will keep this.

sri2k1us:
Thanks for reviews.

PR #127, opened 2020-06-30
User job limit mgmt in VICE admin UI

johnworth:
This PR adds basic concurrent job limit management to the VICE admin UI.

Note: I used the screenshot functionality in Firefox, which does not render the bottom border of the text fields in the phone and tablet layouts. It's actually there, I promise. :)

Limitations:
* Basic support: look up a user's current setting, set it to another value.
* Does not list all settings for all users (should be done in another PR).

# Desktop
image

# Phone
image

# Tablet
image


src/components/vice/admin/joblimits/index.js
```diff
@@ -0,0 +1,196 @@
...
+        onError: (e) => {
+            showErrorAnnouncer(
+                formatMessage(intl, "jobLimitUpdateError", {
+                    username,
+                    currentLimit: newLimit,
+                }),
+                e
+            );
+        },
+    });
```
sri2k1us, line 111: Just as a FYI, we can still use the `useQuery` in `1.x` without the enabled flag. To do that you set the query key to a `falsy` value with `useState` and then update to a value with `key` along with required params when needed.

sri2k1us:
LGTM :shipit: . I had one comment but I can fix it once my PR for `RQ 2.x` is out.

src/components/vice/admin/joblimits/index.js
johnworth: 👍

PR #128, opened 2020-06-30
Apps dot menu

sri2k1us:
image

- App filtering in small screen. I feel this UX is more cleaner on small screen and I was able to remove a lot of CSS that was needed to manage widths across different screen sizes. Also, no more ellipsis.

image

### DIFF
--- a/src/components/apps/toolbar/AppsDotMenu.js
+++ b/src/components/apps/toolbar/AppsDotMenu.js
@@ -9,7 +9,13 @@ import ids from "../ids";
 import messages from "../messages";
 
 import { build, DotMenu, getMessage, withI18N } from "@cyverse-de/ui-lib";
-import { ListItemIcon, ListItemText, MenuItem } from "@material-ui/core";
+import {
+    ListItemIcon,
+    ListItemText,
+    MenuItem,
+    useMediaQuery,
+    useTheme,
+} from "@material-ui/core";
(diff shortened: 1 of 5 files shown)


src/components/apps/toolbar/AppsDotMenu.js
```diff
@@ -0,0 +1,73 @@
...
+        baseId,
+        ButtonProps,
+        detailsEnabled,
+        onDetailsSelected,
+        onFilterSelected,
+    } = props;
+    const theme = useTheme();
+    const matches = useMediaQuery(theme.breakpoints.down("xs"));
+    return (
+        <>
```
ash3rz, line 32: Looks like this fragment isn't needed

src/components/apps/toolbar/AppsDotMenu.js
```diff
@@ -0,0 +1,73 @@
...
+                                onDetailsSelected();
+                            }}
+                        >
+                            <ListItemIcon>
+                                <Info fontSize="small" />
+                            </ListItemIcon>
+                            <ListItemText primary={getMessage("details")} />
+                        </MenuItem>
+                    ),
+                    matches && (
```
psarando, line 52: Not sure what `matches` means here 🤔 
Maybe `mobileView` is more clear?

sri2k1us:
Thanks for the review.

PR #129, opened 2020-07-06
Update react-query to 2.x

sri2k1us:


src/components/analyses/listing/Listing.js
```diff
@@ -115,9 +125,10 @@ function Listing(props) {
             .join(",");
 
         setAnalysesKey([
-            "getAnalyses",
+            ANALYSES_LISTING_QUERY_KEY,
```
ash3rz, line 128: I only did a quick look-over of the changes in v2, but it seems like with the introduction to the `enabled` flag, we don't need to keep the keys in state?  I thought we were keeping the keys in state initially because the previous method was to have a falsy query key to prevent a query from running.  Now it seems like we can remove that and rely on the `enabled` flag.

src/components/apps/listing/Listing.js
```diff
@@ -191,18 +210,23 @@ function Listing({ baseId, onRouteToApp }) {
     }, [data, detailsOpen, selected]);
 
     useEffect(() => {
-        setDetailsEnabled(selected && selected.length === 1);
+        const enabled = selected && selected.length === 1;
+        setDetailsEnabled(enabled);
+        //      setDetailsQueryEnabled(enabled);
```
ash3rz, line 215: Seems like this comment can be removed

src/serviceFacades/analyses.js
```diff
@@ -1,5 +1,9 @@
 import callApi from "../common/callApi";
 
+const ANALYSES_LISTING_QUERY_KEY = "fetchAnalysesListingKey";
+const ANALYSIS_PARAMS_QUERY_KEY = "fetchAnalysisParamsKey";
+const ANALYSIS_HISTORY_QUERY_KEY = "fetchAnalysisHistoryKey";
```
ash3rz, line 5: 👍

src/components/analyses/listing/Listing.js
sri2k1us: I think we will still need to keep the key in state at least for analyses and app listings. For eg: in analyses listing, the key contains additional parameters like `pagesize`, `offset`, `filters` etc. Although, all these parameters are in state, the `filter` needs little bit of processing before we fire off the query. That happens between line `86 - 125` in `/analyses/listing/Listing.js`. For data listing, we can get away with not having the `key` in state but it does include `uploadsCompleted` which forces a `refresh`. Not sure if there is another approach to managing these keys.

ash3rz: This is ultimately minor, so feel free to disregard. If I understand correctly, you can control when a query runs by either managing when the query key gets set or setting the enabled flag to true/false. And instead of picking one of those methods, it looks like we're kind of doing both is what I'm saying.

I haven't looked too deeply at the code, so it may be the case that removing the keys from state and moving the components of the keys into state (like the filters you were talking about since `page`, `orderBy`, etc. are already in state) would end up being the same number of things in state. In that case, it doesn't really make a difference and definitely disregard my comment.

psarando:
It looks like the chromatic snapshot for `Tools: Errored Listing Test` changed: it's showing the loading skeleton instead of the 400 error message now.

I can view the diff by clicking the `UI Tests`, but I can't comment on it in chromatic (I guess because we reached our snapshot limit this month).

sri2k1us:
> It looks like the chromatic snapshot for `Tools: Errored Listing Test` changed: it's showing the loading skeleton instead of the 400 error message now.
> 
> I can view the diff by clicking the `UI Tests`, but I can't comment on it in chromatic (I guess because we reached our snapshot limit this month).

I checked this in SB. While it took some time to get the actual error msg displayed, it did came eventually. So I am thinking this is a matter of timing issue although am not sure what changed caused that. See screenshot:

image

sri2k1us:
Thanks for the reviews.

PR #130, opened 2020-07-06
Update material-UI and other libs

sri2k1us:
Notes
- This is on top of #129
- Material-UI renamed `ExpansionPanel` to `Accordion`

PR #131, opened 2020-07-07
Router

sri2k1us:
- Break up large router file according to end-points.
- Note this is on top of #130. Review from https://github.com/cyverse-de/sonora/pull/131/commits/0f2f254b7cd601f341ba2539d06f804e58b3ea79


src/server/api/user.js
```diff
@@ -0,0 +1,38 @@
...
+    api.get(
+        "/user-info",
+        auth.authnTokenMiddleware,
+        terrainHandler({
+            method: "GET",
+            pathname: "/secured/user-info",
+        })
+    );
+
+    logger.info("adding the GET /api/dashboard handler");
```
ash3rz, line 24: I wouldn't think to look here in the `user` file to find this endpoint.  Might just be me though.

src/server/api/data.js
```diff
@@ -0,0 +1,144 @@
...
+        terrainHandler({
+            method: "POST",
+            pathname: "/secured/filetypes/type",
+            headers: {
+                "Content-Type": "application/json",
+            },
+        })
+    );
+
+    logger.info("adding the POST /share handler");
```
ash3rz, line 130: Is this endpoint exclusive to data sharing?  I think we use this for apps and analyses sharing as well, which might make this hard to find in the `data` file

src/server/api/user.js
sri2k1us: I was debating between putting it in a separate file vs putting it `user` and ultimately decided to put it here. I will move it a separate file.

src/server/api/data.js
sri2k1us: Thats true. I will move this a separate file.

src/server/index.js
```diff
@@ -68,8 +81,20 @@ app.prepare()
             getNotifications(authn.getUserID(request), ws);
         });
 
-        logger.info("adding the api router to the express server");
-        server.use("/api", apiRouter());
+        logger.info(
+            "$$$$$$$$ adding the api router to the express server $$$$$$$$$"
```
psarando, line 85: 🤑

sri2k1us:
Thanks for the reviews.

src/server/api/router.js
```diff
@@ -1,551 +0,0 @@
...
-        "/quicklaunches",
-        auth.authnTokenMiddleware,
-        terrainHandler({
-            method: "POST",
-            pathname: "/quicklaunches",
-            headers: {
-                "Content-Type": "application/json",
-            },
-        })
-    );
```
psarando, line 367: Oops, missed this in my review, but it looks like this endpoint wasn't ported into a new file, so it's missing now.

sri2k1us: 😖

PR #132, opened 2020-07-10
Try out next-i18next for i18N.

sri2k1us:


public/static/locales/en/common.json
```diff
@@ -0,0 +1,29 @@
...
+    "searchAriaLabel": "search",
+    "searchButtonAriaControl": "search button",
+    "searchFilterAriaControl": "search filter menu",
+    "searchFilterAriaLabel": "search filter",
+    "searchFilterMenuAriaLabel": "select search filter",
+    "settings": "Settings",
+    "userMenuAriaLabel": "user menu",
+    "vice": "VICE",
+    "viewOutput": "View output"
+}
```
sri2k1us, line 29: This is the translation file. This can be broken up into several namespaces like data, apps etc...

src/pages/_app.js
```diff
@@ -145,33 +163,14 @@ function MyApp({ Component, pageProps, intercom, admin, irods, tools }) {
...
-    };
-    return clientConfig;
+MyApp.getInitialProps = async ({ Component, ctx }) => {
+    let pageProps = {};
+
+    if (Component.getInitialProps) {
+        pageProps = await Component.getInitialProps(ctx);
+    }
+
+    return { pageProps, namespacesRequired: ["common"] };
```
sri2k1us, line 173: This is where is spent the most time figuring out. Turns out I need to move the config step into the function and just return pageProps here.

src/server/index.js
```diff
@@ -73,6 +76,9 @@ app.prepare()
             res.redirect(req.url.replace(/^\/login/, ""));
         });
 
+        logger.info("Use translations");
+        server.use(nextI18NextMiddleware(nextI18next));
+
```
sri2k1us, line 81: Load translations files based on the language.

src/components/layout/CustomIntercom.js
```diff
@@ -64,4 +61,4 @@ function CustomIntercom(props) {
     }
 }
 
-export default CustomIntercom;
+export default withTranslation("common")(CustomIntercom);
```
ash3rz, line 64: Is it possible to pull from multiple files?  Like if my component needs a string from `common` and also a string from `data`?

src/pages/_app.js
ash3rz, line 173: Is `namespacesRequired` supposed to be a list of all the message files we have?  I wonder if there's a way we can generate that, maybe with `fs`, so we don't have to maintain this list ourselves.

src/pages/_app.js
```diff
@@ -145,33 +163,14 @@ function MyApp({ Component, pageProps, intercom, admin, irods, tools }) {
     );
 }
 
-MyApp.getInitialProps = async (ctx) => {
-    const { publicRuntimeConfig = {} } = getConfig() || {};
```
ash3rz, line 149: Was there a reason to move this stuff into the `useEffect`?  Personally, I liked here, since it was a little more out of the way.  Minor thing though.

src/components/layout/CustomIntercom.js
sri2k1us: Yes. use can use an array syntax here.

src/pages/_app.js
sri2k1us: Yes. I only have one but we can add any array of namespaces.

src/pages/_app.js
sri2k1us: Yes. I got an error mentioned here - https://github.com/isaachinman/next-i18next/issues/124. I could make it work with this change.

sri2k1us:
Do you know some code in Sonora where we do this ? I can try to check this use case 

> Looks significantly better than our current library!
> 
> I left some comments and one other question I had - we've run into edge cases with `react-intl` where the parent component is wrapped with i18n and uses one messages file, and a child component is also wrapped with i18n and uses a different messages file. In some cases, the child component can't actually find its messages because the parent's messages are overriding it. It'd be good to test that somewhere and see what happens. We're trying to fix that with our own HOC to combine the parent and child messages [here](https://github.com/cyverse-de/ui-lib/blob/3669382931a745da031ac2a88700789c63ffa04a/src/util/I18NWrapper.js#L14-L20).

src/components/dashboard/index.js
```diff
@@ -9,7 +9,7 @@ import React from "react";
 import { useRouter } from "next/router";
 import clsx from "clsx";
 import { useQuery } from "react-query";
-import { injectIntl } from "react-intl";
+import { useTranslation } from "react-i18next";
```
sri2k1us, line 12: We can use `useTranslation` hook instead of using the `withTranslation` HOC

src/components/layout/CyVerseAppBar.js
```diff
@@ -208,6 +207,7 @@ function CyverseAppBar(props) {
     const router = useRouter();
     const ref = useRef();
     const [config, setConfig] = useConfig();
+    const { t } = useTranslation(["common", "search"]);
```
sri2k1us, line 210: Demonstrate use of array syntax to import multiple namespaces.

src/pages/dashboard.js
```diff
@@ -5,3 +5,13 @@ import Dashboard from "../components/dashboard";
...
+
+DashboardPage.getInitialProps = async ({ Component, ctx }) => {
+    let pageProps = {};
+
+    if (Component.getInitialProps) {
+        pageProps = await Component.getInitialProps(ctx);
+    }
+
+    return { pageProps, namespacesRequired: ["dashboard"] };
+};
```
sri2k1us, line 17: I removed the `getInitialProps` from `_apps.js` and moved it here cos I was seeing warning in the console
`You have not declared a namespacesRequired array on your page-level component: DashboardPage. This will cause all namespaces to be sent down to the client, possibly negatively impacting the performance of your app.`

Also see - 

https://github.com/isaachinman/next-i18next#4-declaring-namespace-dependencies

ash3rz: So if I'm understanding correctly, then each page in the `pages` directory will need to have this same thing but with whatever message files that page will need?

src/components/layout/CustomIntercom.js
```diff
@@ -19,7 +18,8 @@ import { Badge, IconButton, Tooltip, useTheme } from "@material-ui/core";
 import LiveHelpIcon from "@material-ui/icons/LiveHelp";
 
 function CustomIntercom(props) {
-    const { intercomUnreadCount, t } = props;
+    const { intercomUnreadCount } = props;
+    const { t } = useTranslation("common");
```
ash3rz, line 22: Just picking a random spot to comment because the quoting thing is harder to follow - for an example of issues we've had with `react-intl` and having 2 components working together with 2 different messages files: `withErrorAnnouncer`.  The error announcer is using the `utils` directory messages file, and any component that uses `withErrorAnnouncer` will be using its own messages file.  `getMessage` for some reason didn't work (when using the `Details` button, the details text could never be found, IIRC), but `formatMessage` did work.

src/pages/dashboard.js
sri2k1us: Thats correct.

sri2k1us: That way it won't download translation files for the whole app. Instead only files related to that page or downloaded.

src/components/layout/CustomIntercom.js
sri2k1us: Let me check this.

src/pages/dashboard.js
ash3rz: OK, so if during development my component needs a message from a file my page hasn't used before, I have to add it to the `useTranslation` hook in my component and also remember to add it to the page's `getInitialProps`? The `getInitialProps` part might be a little annoying / easy to forget. I was hoping we could have what you had before with one global `getInitialProps` that uses `fs` to import all the message files automatically and never have to worry about updating it manually, but I still think the convenience we get at the component level is far, far better than what we have with `react-intl`.  I'm still for moving forward with this change 👍

sri2k1us: No. You don't update the `getInitialProps` during development. I just double checked. Updating just the `useTranslation` with namespace should work. As I said, this is a performance optimization so that the page won't download all the translation file the app.

sri2k1us: Just make sure you do update the `getInitialProps` when you put a PR :) / changing something for production deployment.

ash3rz: Got it - that's not that bad then!  We'll just all need to be mindful during PR review to catch that too.  Hopefully we get an error or something to catch it quickly otherwise.

src/i18n.js
```diff
@@ -0,0 +1,5 @@
+const NextI18Next = require("next-i18next").default;
+
+module.exports = new NextI18Next({
+    defaultLanguage: "en",
+});
```
psarando, line 5: Is this syntax required? Could we use the import/export syntax we usually use?
```suggestion
import NextI18Next from "next-i18next";

export default new NextI18Next({
    defaultLanguage: "en",
});
```

src/server/index.js
```diff
@@ -47,6 +47,9 @@ function buildNavigationRouteRegexp() {
 // Configure the Keycloak client.
 const keycloakClient = authn.getKeycloakClient();
 
+const nextI18next = require("../i18n");
+const nextI18NextMiddleware = require("next-i18next/middleware").default;
```
psarando, line 51: Can we use `import` syntax here?
```suggestion
import nextI18next from "../i18n";
import nextI18NextMiddleware from "next-i18next/middleware";
```
Except these lines would then have to be moved to the top of the file with the rest.

src/components/dashboard/index.js
```diff
@@ -178,30 +171,30 @@ const useStyles = makeStyles((theme) => ({
     },
 }));
 
-const getOrigination = (kind, content, intl) => {
+const getOrigination = (kind, content, t) => {
```
psarando, line 174: Future cleanup: Components could just call `useTranslation` where needed, rather than grabbing `t` from props passed down by a parent component.

I guess we could have already done that with the `injectIntl` HOC, but at least for me it seems more cumbersome to wrap a small component that's not used anywhere else in `injectIntl` 🆚 calling `useTranslation` inside it, especially if wrapping it in an HOC alters the entire component's indentation 😅

sri2k1us: Actually I did in other spots in this file. I missed it here.

src/i18n.js
sri2k1us: Yes this syntax is required (I double checked) due to some polyfill issues. I can't find the link any more that describes the issue but I can confirm `import` syntax won't work.

src/server/index.js
sri2k1us: This worked.

PR #133, opened 2020-07-11
CORE-769 Analysis relaunch page

psarando:
This PR will add an analysis relaunch page and update the analyses listing `Relaunch` actions to route to this page.
It also adds a route to the `/analyses/relauncher` endpoint and adds multi-relaunch support to the analyses listing component.

This PR also includes bug fixes for app integrator default input file/folder values and default reference genome values.

### DIFF
--- a/src/components/analyses/toolbar/AnalysesDotMenu.js
+++ b/src/components/analyses/toolbar/AnalysesDotMenu.js
@@ -35,6 +35,7 @@ function DotMenuItems(props) {
         detailsEnabled,
         handleInteractiveUrlClick,
         handleGoToOutputFolder,
+        handleRelaunch,
         handleBatchIconClick,
         isBatch,
         isVICE,
@@ -81,6 +82,7 @@ function DotMenuItems(props) {
                 id={build(baseId, ids.MENUITEM_RELAUNCH)}
                 onClick={() => {
                     onClose();
+                    handleRelaunch(selectedAnalyses[0]);
                 }}
             >
                 <ListItemIcon>
--- a/src/components/analyses/toolbar/AnalysesDotMenu.js
+++ b/src/components/analyses/toolbar/AnalysesDotMenu.js
@@ -82,7 +82,7 @@ function DotMenuItems(props) {
                 id={build(baseId, ids.MENUITEM_RELAUNCH)}
                 onClick={() => {
                     onClose();
-                    handleRelaunch(selectedAnalyses[0]);
+                    handleRelaunch(selectedAnalyses);
                 }}
             >
                 <ListItemIcon>
--- a/src/components/apps/launch.js
+++ b/src/components/apps/launch.js
@@ -1,0 +1,95 @@
+/**
+ * @author psarando
+ *
+ * A component for displaying the App Launch Wizard for a given app.
+ * Also handles calling analysis submission and quick-launch-save endpoints.
+ */
+import React from "react";
+
+import { useRouter } from "next/router";
+import { useMutation } from "react-query";
+
+import NavigationConstants from "../../common/NavigationConstants";
+
+import { useConfig } from "../../contexts/config";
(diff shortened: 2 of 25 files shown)


src/components/apps/launch.js
```diff
@@ -0,0 +1,95 @@
+/**
+ * @author psarando
+ *
+ * A component for displaying the App Launch Wizard for a given app.
+ * Also handles calling analysis submission and quick-launch-save endpoints.
```
psarando, line 5: I moved some common logic that was in the app launch page and also in the new relaunch page into this component.

Depending on what we decide in our UX review meeting, this file could be renamed `src/components/apps/launch/index.js`.

psarando:
I think I had GitHub Actions turned off on my fork during dev, so that I could push WIP commits and rebase without constantly creating Chromatic snapshots. I think the Actions were still off when I submitted this PR, so there are no Chromatic links attached to this PR yet, but here it is:
https://www.chromatic.com/pullrequest?appId=5ebad5c6b991d60022b1e72e&number=133

src/components/analyses/toolbar/AnalysesDotMenu.js
```diff
@@ -81,6 +82,7 @@ function DotMenuItems(props) {
                 id={build(baseId, ids.MENUITEM_RELAUNCH)}
                 onClick={() => {
                     onClose();
+                    handleRelaunch(selectedAnalyses[0]);
```
ash3rz, line 85: Are we still supporting relaunching multiple analyses? In the old DE, it just displays a message about what relaunching multiple analyses is intended to do instead of showing the app launch wizard.

psarando: Ah yes! Good catch. I need to add support for that and call a different endpoint, instead of routing to the relaunch page.
I'll switch this to `Draft` in the meantime.

src/components/apps/launch/AppLaunchForm.js
```diff
@@ -194,16 +194,27 @@ const initValues = ({
...
+            }
+
             if (paramType === constants.PARAM_TYPE.MULTIFILE_SELECTOR) {
-                value = defaultValue || [];
+                value = defaultValue?.path || [];
             }
+
             if (paramType === constants.PARAM_TYPE.FLAG) {
                 value = defaultValue && defaultValue !== "false";
             }
```
sri2k1us, line 211: Just double checking. Is `false` a string here ?

src/pages/analyses/[analysisId]/relaunch.js
```diff
@@ -0,0 +1,60 @@
...
+        config: {
+            enabled: relaunchQueryEnabled,
+            onSuccess: setApp,
+            onError: setRelaunchError,
+        },
+    });
+
+    const loading = relaunchStatus === constants.LOADING;
+
+    return <AppLaunch app={app} appError={relaunchError} loading={loading} />;
```
Comment (sri2k1us) [inline src/pages/analyses/[analysisId]/relaunch.js:59]: `appError` could also use renaming to `launchError`

src/components/layout/Notifications.js
```diff
@@ -71,23 +71,28 @@ function Notifications(props) {
...
+                notification?.payload?.analysisresultsfolder;
+
+            const completed = status === analysisStatus.COMPLETED;
+            const failed = status === analysisStatus.FAILED;
+
+            const variant = completed
+                ? AnnouncerConstants.SUCCESS
+                : failed
+                ? AnnouncerConstants.ERROR
+                : AnnouncerConstants.INFO;
```
ash3rz, line 85: 👍

src/components/apps/launch/AppLaunchForm.js
psarando: Yes, this value is sent as a string by the service.

psarando:
Thanks for the reviews! 👍

PR #137, opened 2020-07-14
Fix data details permission listing issue.

sri2k1us:


src/server/api/data.js
```diff
@@ -136,5 +136,5 @@ export default function dataRouter() {
         })
     );
 
-   return api;
+    return api;
```
sri2k1us, line 139: Prettier was adamant that the formatting is off here.

sri2k1us:
Thanks for the reviews.

PR #139, opened 2020-07-15
Bootstrap without IP address.

sri2k1us:


src/server/api/user.js
```diff
@@ -40,5 +40,38 @@ export default function userRouter() {
...
+    api.get(
+        "/redirect-uris",
+        auth.authnTokenMiddleware,
+        terrainHandler({
+            method: "GET",
+            pathname: "/secured/oauth/redirect-uris",
+        })
+    );
+
     return api;
```
sri2k1us, line 76: Turns out I had missed preferences handlers as well during `router` refactor.

sri2k1us:
Thanks for the reviews.

PR #143, opened 2020-07-22
Add url query params for app listing to fix back button.

sri2k1us:
- Add   `order,
    orderBy,
    page,
    rowsPerPage,
    filter,
    category` to URL query params.
- Fix Icon for Grid View Type
- Remove `All` from filter options. 
- Set `Browse All Apps` as default option.
Todo:
- `as` - The url that will be shown in the browser. (if we decide to, in a Separate PR)

### DIFF
--- a/src/components/apps/listing/Listing.js
+++ b/src/components/apps/listing/Listing.js
@@ -373,11 +373,9 @@ function Listing({
             }
             setCategory(category);
             setSelected([]);
-            if (selectedPage === 0) {
-                setPage(0);
-            }
+            setPage(0);
         },
-        [selectedPage]
+        []
     );
 
     const handleFilterChange = (filter) => {
--- a/src/components/apps/listing/Listing.js
+++ b/src/components/apps/listing/Listing.js
@@ -13,8 +13,6 @@ import TableView from "./TableView";
 import intlData from "../messages";
 import AppsToolbar from "../toolbar/Toolbar";
 
-import { getAppTypeFilters } from "../toolbar/AppNavigation";
-
 import appType from "../../models/AppType";
 import DEPagination from "../../utils/DEPagination";
 
@@ -36,16 +34,28 @@ import { withI18N } from "@cyverse-de/ui-lib";
 import { injectIntl } from "react-intl";
 import { queryCache, useMutation, useQuery } from "react-query";
 
-function Listing({ baseId, onRouteToApp }) {
+function Listing({
+    baseId,
+    onRouteToApp,
+    onRouteToListing,
+    selectedPage,
+    selectedRowsPerPage,
+    selectedOrder,
+    selectedOrderBy,
+    selectedFilter,
+    selectedCategory,
+}) {
     const [isGridView, setGridView] = useState(false);
-    const [order, setOrder] = useState("asc");
-    const [orderBy, setOrderBy] = useState("name");
+
+    const [order, setOrder] = useState(selectedOrder);
+    const [orderBy, setOrderBy] = useState(selectedOrderBy);
+    const [page, setPage] = useState(selectedPage);
+    const [rowsPerPage, setRowsPerPage] = useState(selectedRowsPerPage);
+    const [filter, setFilter] = useState(selectedFilter);
+    const [category, setCategory] = useState(selectedCategory);
+
     const [selected, setSelected] = useState([]);
     const [lastSelectIndex, setLastSelectIndex] = useState(-1);
-    const [page, setPage] = useState(0);
-    const [rowsPerPage, setRowsPerPage] = useState(25);
-    const [selectedCategory, setSelectedCategory] = useState(null);
-    const [filter, setFilter] = useState(getAppTypeFilters()[0]);
 
     const [data, setData] = useState(null);
     const [agaveAuthDialogOpen, setAgaveAuthDialogOpen] = useState(false);
@@ -75,7 +85,7 @@ function Listing({ baseId, onRouteToApp }) {
 
     //a query with falsy key will not execute until key is set truthy val
     const {
-        status: appInCategoryStatus,
+        isFetching: appInCategoryStatus,
         error: appsInCategoryError,
     } = useQuery({
         queryKey: appsInCategoryKey,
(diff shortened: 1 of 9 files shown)


src/components/apps/toolbar/AppNavigation.js
```diff
@@ -101,41 +94,40 @@ function AppNavigation(props) {
...
-    const allAppsCategory = getAllAppsCategory(constants.BROWSE_ALL_APPS);
+
+    const iconMap = new Map();
+    iconMap.set(constants.APPS_UNDER_DEV, <LockIcon />);
+    iconMap.set(constants.APPS_SHARED_WITH_ME, <FolderSharedIcon />);
+    iconMap.set(constants.FAV_APPS, <FavoriteIcon />);
+    iconMap.set(constants.MY_PUBLIC_APPS, <GroupWorkIcon />);
+    iconMap.set(constants.BROWSE_ALL_APPS, <AppsIcon />);
+    iconMap.set(constants.HPC, <StorageIcon />);
+
```
sri2k1us, line 105: Having the `icons` in the inside the `Category` object caused problems during `JSON.stringfy` and `JSON.parse`. React barked saying I cannot render object.

src/pages/apps.js
```diff
@@ -1,19 +1,68 @@
...
+    const selectedFilter = query.selectedFilter
+        ? JSON.parse(query.selectedFilter)
+        : null;
 
+    const selectedCategory = query.selectedCategory
+        ? JSON.parse(query.selectedCategory)
+        : {
+              name: constants.BROWSE_ALL_APPS,
+              id: constants.BROWSE_ALL_APPS_ID,
+          };
```
sri2k1us, line 36: Default settings set here.

src/components/models/AppType.js
```diff
@@ -1,5 +1,4 @@
 const appType = {
-    all: "All",
```
sri2k1us, line 2: I removed the `All` from filters since Nirav mentioned that `Filter` `All` is confusing. With `All` option removed, users can filter by `appType` but if they want to see everything, they can simply hit `x` button next to the filter to clear it which will automatically show everything.

sri2k1us:
> I didn't spot any problems, although I was a little confused by the fact that the UI uses a fake category ID to browse all apps. As always, take any UI review that I do with a grain of salt. I'm still a newbie at this stuff.

So "Browse All Apps" is a dummy Category am creating to add to the drop down in the UI. The Category name and Id are purely used to determine that user has selected to view all the apps. Thats about it. Its `ID` won't be used to retrieve anything from `terrain`

slr71:
> So "Browse All Apps" is a dummy Category am creating to add to the drop down in the UI. The Category name and Id are purely used to determine that user has selected to view all the apps. Thats about it. Its `ID` won't be used to retrieve anything from `terrain`

I think my confusion came primarily from the order in which I was reading the code. A few lines of code made a lot more sense once I realized that the dummy category existed. 😆

src/components/apps/listing/Listing.js
```diff
@@ -36,16 +34,28 @@ import { withI18N } from "@cyverse-de/ui-lib";
...
     const [isGridView, setGridView] = useState(false);
-    const [order, setOrder] = useState("asc");
-    const [orderBy, setOrderBy] = useState("name");
+
+    const [order, setOrder] = useState(selectedOrder);
+    const [orderBy, setOrderBy] = useState(selectedOrderBy);
+    const [page, setPage] = useState(selectedPage);
+    const [rowsPerPage, setRowsPerPage] = useState(selectedRowsPerPage);
+    const [filter, setFilter] = useState(selectedFilter);
+    const [category, setCategory] = useState(selectedCategory);
```
ash3rz, line 55: 👍

src/components/apps/listing/Listing.js
```diff
@@ -301,26 +355,29 @@ function Listing({ baseId, onRouteToApp }) {
...
                 appType.agave.toLowerCase()
             ) {
                 setFilter(null);
             }
-            setSelectedCategory(selectedCategory);
+            setCategory(category);
             setSelected([]);
-            setPage(0);
+            if (selectedPage === 0) {
+                setPage(0);
```
ash3rz, line 377: Seems like having this if statement might cause issues if I'm initially browsing All Apps on page 100 and then try to view my Apps Under Development which may only have 1 page.

src/components/apps/toolbar/AppNavigation.js
```diff
@@ -101,41 +94,40 @@ function AppNavigation(props) {
...
 
     const preProcessData = useCallback(
         (data) => {
             const privateCat = data.categories.find(
-                (cat) => cat.system_id === "de"
+                (cat) => cat.system_id === systemId.de
             );
             const hpcCat = data.categories.find(
-                (cat) => cat.system_id === "agave"
+                (cat) => cat.system_id === systemId.agave
```
ash3rz, line 119: 👍

src/components/apps/toolbar/AppNavigation.js
```diff
@@ -101,41 +94,40 @@ function AppNavigation(props) {
...
-            } else {
-                setCategories(privateCat.categories);
+                categoryList = categoryList.concat(hpcCat);
             }
+            categoryList = categoryList.concat(allAppsCategory());
+            setCategories(categoryList);
             handleAppNavError(null);
         },
-        [setCategories, handleAppNavError]
+        [allAppsCategory, handleAppNavError]
```
ash3rz, line 130: I'm not seeing a warning here, but I would think `setCategories` might still need to be in the useEffect's dependency array

src/components/apps/utils.js
```diff
@@ -13,3 +13,23 @@ import NavigationConstants from "../../common/NavigationConstants";
  */
 export const getAppLaunchPath = (systemId, appId) =>
     `/${NavigationConstants.APPS}/${systemId}/${appId}/launch`;
+
+/**
+ *
+ * @param {stirng} order The apps listing sort order, asc or desc.
```
ash3rz, line 19: Tpyo in `stirng`

src/components/apps/toolbar/AppNavigation.js
sri2k1us: I will try adding it and see 🤷‍♂️

src/components/apps/listing/Listing.js
sri2k1us: Let me double check this.

sri2k1us: This was indeed a bug. Fixed it. Thanks.

sri2k1us:
thanks for the reviews.

PR #144, opened 2020-07-23
Use react-i18next in VICE admin

johnworth:
Switch over the VICE admin UI over to using the new react-i18next stuff. No visible UI changes, from what I saw.


johnworth:
> 
> 
> LGTM. But we need to add `vice-admin` to page to prevent it downloading all the translations. I know this step is something we need to start remembering. I remember seeing some kind of warning in server console. Not sure though.
> 
> ```
> VICEAdminPage.getInitialProps = async ({ Component, ctx }) => {
>     let pageProps = {};
>     if (Component.getInitialProps) {
>         pageProps = await Component.getInitialProps(ctx);
>     }
> 
>     return { pageProps, namespacesRequired: ["vice-admin"] };
> };
> ```

This fix should be in now.

johnworth:
Thanks for the reviews.

PR #145, opened 2020-07-23
Update guildlines for i18next

sri2k1us:


guidelines.md
```diff
@@ -81,96 +81,47 @@ Create a test for your component(s) located in the [/src/__tests__](/src/__tests
...
-    )
-}
-
--export default MyComponent;
-+export default withI18N(MyComponent, intlData);
-
 ```
 
-3. Update your component to fetch the internationalized text with the corresponding key you created. The [getMessage(key)](https://github.com/cyverse-de/ui-lib/blob/78880901c263c14ea697a5abd9b607fbd776ec4b/src/util/I18NWrapper.js#L29-L34) helper function covers roughly 80% of the use cases
+2. Import all the translation json file(s) into your react component with filename used as namespace. For example, if you have named the translation files `common.json` and `search.json`, you can import those translations as
```
ash3rz, line 101: There should also be mention of what needs to get added on the page level with `getInitialProps` I believe?

guidelines.md
```diff
@@ -81,96 +81,47 @@ Create a test for your component(s) located in the [/src/__tests__](/src/__tests
...
-+            <span
-+                title={formatMessage(props.intl, "stringTitle")}
-+            >
-+                {getMessage("header")}
-+            </span>
-            <span>{getMessage("namedHeader", {values: {name: props.name}})}</span>
+ -           <span>`Welcome ${name}`</span>
+ -           <span>Discovery Environment</span>
+ +           <span>{t("welcome",{ name: name })}
+ +           <span>{t("deTitle")}</span>
```
ash3rz, line 118: 👍

guidelines.md
```diff
@@ -220,4 +171,5 @@ It's recommended that 2 developers review each major pull request though there's
 1. Your component should meet all of the above and align with our UI/UX [goals](#ui--ux-goals)
 1. Tests should pass - `npm test`
 1. Lint checking should pass - `npm run lint`
+1. Format checking should pass - `npx prettier --check ./src/**/*.js ./.storybook/**/*.js ./stories/**/*.js`
```
ash3rz, line 174: I believe this can just be `npm run format` or maybe we could add a script in `package.json` that only does the `--check` flag

guidelines.md
sri2k1us: That right. I was in double mind on whether to put it here and make the instructions complicated. I will go ahead and add it.

guidelines.md
sri2k1us: `npm run format` formats the whole project which makes it little intrusive. I will probably add a check task.

sri2k1us:
Thanks for review.

PR #147, opened 2020-07-28
Launch improvements

sri2k1us:
- Set `Stepper` height dynamically.
- Highlight app `sections`.
- Replace `BottomNavigationBar` with `Toolbar`.
- Fix mobile view launch.
- Remove `Retain Input` flag
- 

Laptop:
image

image

image

image

iphone 5 / SE:

image

image

image

image

~~Notice - that long descriptions get truncated into `ellipsis` and hidden in mobile view. But we add a `i` button next to app name for user to view app details and documentation.~~ - AppInfo bar has dynamic height now.

> Overall looks good.
> 
> I know there are more changes coming after today's UX meeting, but we didn't get to the mobile view in the meeting, so I will mention that it would be great to find a way to prevent side-scrolling in mobile view, if possible.

I checked the mobile view, the side scroll happens only for apps like `Tuxedo` cos of the wall of text. We are setting the text using `inner HTML` which messes up word wrap I think 🤔 . Here is a view of another app in iphone 5/ SE. 

image

This app has negligible amount of  Side-scroll but I think that can be fixed too.
Note: I still don't know what causes that `Browse` button to misalign in mobile view. I will look into later.

### DIFF
--- a/src/components/apps/launch/AnalysisInfoForm.js
+++ b/src/components/apps/launch/AnalysisInfoForm.js
@@ -36,6 +36,7 @@ const AnalysisInfoForm = ({ formId, appType, startingPath }) => (
             label={getMessage("analysisName")}
             required={true}
             name="name"
+            size="small"
             component={FormTextField}
             inputProps={{ maxLength: 255 }}
         />
--- a/src/components/apps/launch/AnalysisInfoForm.js
+++ b/src/components/apps/launch/AnalysisInfoForm.js
@@ -8,7 +8,6 @@ import React from "react";
 
 import { FastField } from "formik";
 
-import constants from "../../../constants";
 import ResourceTypes from "../../models/ResourceTypes";
 
 import ids from "./ids";
@@ -18,7 +17,6 @@ import InputSelector from "./InputSelector";
 
 import {
     build as buildDebugId,
-    FormCheckbox,
     FormMultilineTextField,
     FormTextField,
     getMessage,
@@ -37,6 +35,7 @@ const AnalysisInfoForm = ({ formId, appType, startingPath }) => (
             required={true}
             name="name"
             size="small"
+            helperText={getMessage("analysisNameHelp")}
             component={FormTextField}
             inputProps={{ maxLength: 255 }}
         />
@@ -63,18 +62,6 @@ const AnalysisInfoForm = ({ formId, appType, startingPath }) => (
             startingPath={startingPath}
             acceptedType={ResourceTypes.FOLDER}
         />
-        {appType !== constants.APP_TYPE_EXTERNAL && (
-            <FastField
-                id={buildDebugId(
-                    formId,
-                    ids.LAUNCH_ANALYSIS_GROUP,
-                    ids.APP_LAUNCH_RETAIN_INPUTS
-                )}
-                label={getMessage("retainInputsLabel")}
-                name="debug"
-                component={FormCheckbox}
-            />
-        )}
     </>
 );
 
--- a/src/components/apps/launch/AppInfo.js
+++ b/src/components/apps/launch/AppInfo.js
@@ -26,7 +26,6 @@ import {
     Hidden,
     Link,
     makeStyles,
-    Paper,
     Typography,
     useMediaQuery,
     useTheme,
@@ -60,13 +59,8 @@ const LoadingErrorDisplay = injectIntl(({ intl, baseId, loadingError }) => {
     );
 });
 
-const AppInfo = ({
-    app,
-    baseId,
-    hasDeprecatedParams,
-    loading,
-    loadingError,
-}) => {
+const AppInfo = React.forwardRef((props, ref) => {
+    const { app, baseId, hasDeprecatedParams, loading, loadingError } = props;
     const unavailableMsgKey = app?.deleted
         ? "appDeprecated"
         : app?.disabled
@@ -78,35 +72,32 @@ const AppInfo = ({
     const theme = useTheme();
     const isMobile = useMediaQuery(theme.breakpoints.down("xs"));
     return (
-        <>
-            <Paper className={classes.appInfoContainer} elevation={0}>
+        <div ref={ref}>
+            <Typography
+                variant={isMobile ? "subtitle2" : "h6"}
+                className={classes.appInfoTypography}
+            >
+                {loadingError ? (
+                    <LoadingErrorDisplay
+                        baseId={baseId}
+                        loadingError={loadingError}
+                    />
+                ) : loading ? (
+                    <Skeleton width={250} />
+                ) : (
+                    app?.name
+                )}
+            </Typography>
+            <Hidden xsDown>
                 <Typography
-                    variant={isMobile ? "subtitle2" : "h6"}
                     className={classes.appInfoTypography}
+                    variant="body2"
+                    display="block"
+                    gutterBottom
                 >
-                    {loadingError ? (
-                        <LoadingErrorDisplay
-                            baseId={baseId}
-                            loadingError={loadingError}
-                        />
-                    ) : loading ? (
-                        <Skeleton width={250} />
-                    ) : (
-                        app?.name
-                    )}
+                    {loading ? <Skeleton /> : app?.description}
                 </Typography>
-                <Hidden xsDown>
-                    <Typography
-                        className={classes.appInfoTypography}
-                        variant="body2"
-                        display="block"
-                        gutterBottom
-                        noWrap={true}
-                    >
-                        {loading ? <Skeleton /> : app?.description}
-                    </Typography>
-                </Hidden>
-            </Paper>
+            </Hidden>
             <Box m={2}>
                 {(app?.deleted || app?.disabled || hasDeprecatedParams) && (
                     <Typography color="error" variant="body1" gutterBottom>
@@ -130,8 +121,8 @@ const AppInfo = ({
                     </Typography>
                 )}
             </Box>
-        </>
+        </div>
     );
-};
+});
 
 export default AppInfo;
--- a/src/components/apps/launch/AppInfo.js
+++ b/src/components/apps/launch/AppInfo.js
@@ -65,10 +65,10 @@ const AppInfo = ({
         : hasDeprecatedParams
         ? "appParamsDeprecated"
         : null;
-        const classes = useStyles();
+    const classes = useStyles();
     return (
         <>
-            <Paper className={classes.appInfoContainer} elevation={0}> 
+            <Paper className={classes.appInfoContainer} elevation={0}>
                 <Typography variant="h6" className={classes.appInfoTypography}>
                     {loadingError ? (
                         <LoadingErrorDisplay
@@ -81,7 +81,11 @@ const AppInfo = ({
                         app?.name
                     )}
                 </Typography>
-                <Typography variant="body2" gutterBottom className={classes.appInfoTypography}>
+                <Typography
+                    variant="body2"
+                    gutterBottom
+                    className={classes.appInfoTypography}
+                >
                     {loading ? <Skeleton /> : app?.description}
                 </Typography>
             </Paper>
--- a/src/components/apps/launch/AppInfo.js
+++ b/src/components/apps/launch/AppInfo.js
@@ -23,11 +23,13 @@ import {
 
 import {
     Box,
+    Hidden,
     Link,
     makeStyles,
     Paper,
     Typography,
-    Hidden,
+    useMediaQuery,
+    useTheme,
 } from "@material-ui/core";
 
 import { Skeleton } from "@material-ui/lab";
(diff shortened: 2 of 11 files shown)


src/components/apps/launch/AppInfo.js
```diff
@@ -77,13 +95,19 @@ const AppInfo = ({
...
+                        variant="body2"
+                        display="block"
+                        gutterBottom
+                        noWrap={true}
+                    >
+                        {loading ? <Skeleton /> : app?.description}
+                    </Typography>
+                </Hidden>
+            </Paper>
             <Box m={2}>
```
sri2k1us, line 110: Replaced `Toolbar` with `Paper` so that the launch screen surface looks uniform (white bg).

src/components/apps/launch/AppLaunchForm.js
```diff
@@ -48,21 +51,30 @@ import {
...
 import { ArrowBack, ArrowForward, PlayArrow, Save } from "@material-ui/icons";
 
 const useStyles = makeStyles(styles);
 
+//Includes sonora appbar height, appInfo bar height and bottom navigation bar height.
+//Stepper height is calculated dynamically.
+//TODO: calculate these number dynamically.
+const NAVIGATION_BAR_HEIGHT = 200;
+const MOBILE_NAVIGATION_BAR_HEIGHT = 150;
+
```
sri2k1us, line 77: Needed for dynamic height calculation.

src/components/apps/launch/AppLaunchForm.js
```diff
@@ -88,60 +100,56 @@ const StepperBottomNavigation = ({
...
+                    endIcon={<ArrowForward />}
+                    size="small"
+                    onClick={handleNext}
+                >
+                    {isMobile ? "" : getMessage("next")}
+                </Button>
+            )}
+        </Toolbar>
+    );
+};
```
sri2k1us, line 152: I replaced `BottomNavigation` with `Toolbar` for few reasons:
1. `BottomNavigation` was not allowing me set `dense` padding 
2.  Removing labels caused icons to misalign
3.  I wasn't able to set Flex button layout.
4. Other minor `css` annoyances (that may be due to my css ignorance)

src/components/apps/launch/AppLaunchForm.js
```diff
@@ -427,6 +437,16 @@ const AppLaunchForm = (props) => {
             )
         );
 
+    const stepperRef = React.useRef(null);
+    const [stepperHeight, setStepperRef] = useComponentHeight();
+    const appBarHeight =
+        stepperHeight +
+        (isMobile ? MOBILE_NAVIGATION_BAR_HEIGHT : NAVIGATION_BAR_HEIGHT);
+
```
sri2k1us, line 445: Was able to save some space for `mobile` view by removing text and reducing `margins` and `padding`.

src/components/apps/launch/ParamGroups.js
```diff
@@ -312,15 +342,23 @@ const ParamsReviewValue = ({ param }) => {
...
         case constants.PARAM_TYPE.REFERENCE_ANNOTATION:
             if (value?.name) {
-                return value.name;
+                return (
+                    <Typography className={classes.paramsReview}>
+                        {value.name}
+                    </Typography>
+                );
             }
             break;
```
sri2k1us, line 364: Use `Typography` instead of `plain` text.

src/components/layout/CyVerseAppBar.js
```diff
@@ -580,10 +580,7 @@ function CyverseAppBar(props) {
                         >
                             <MenuIcon />
                         </IconButton>
-                        <Typography>
-                            {t("discovery")}
-                            &nbsp;{t("environment")}
-                        </Typography>
+                        <Typography>{t("deTitle")}</Typography>
                     </Hidden>
```
sri2k1us, line 584: Noticed this was displaying `html` tags in title so I went ahead and fixed it.

src/components/apps/launch/ParamGroups.js
```diff
@@ -123,28 +122,35 @@ const buildParamId = (baseId, paramIndex, type) => {
...
-                <Typography variant="subtitle1">{group.label}</Typography>
-            </AccordionSummary>
-            <AccordionDetails className={classes.accordionDetails}>
+                {getMessage("section", {
+                    values: { groupNumber: index + 1, totalGroups: noOfGroups },
+                })}
+                {group.label}
+            </Typography>
+            <Divider className={classes.paramsSectionHeader} />
+            <Grid container spacing={3}>
```
sri2k1us, line 153: `Grid` layout allows the parameters fields to expand to fill the screen according to screen size.

src/components/apps/launch/ParamGroups.js
```diff
@@ -123,28 +122,35 @@ const buildParamId = (baseId, paramIndex, type) => {
...
         referenceGenomesLoading,
         startingPath,
     } = props;
+    const classes = useStyles();
+    const theme = useTheme();
+    const isMobile = useMediaQuery(theme.breakpoints.down("xs"));
 
     return (
-        <Accordion id={baseId} defaultExpanded>
-            <AccordionSummary
```
psarando, line 140: I think we agreed in today's meeting to stick with Accordions for now, for the way they offset the parameter groups from each other, but we also wanted better styling for the group header/summary 👍

src/components/apps/launch/AppLaunchWizard.js
```diff
@@ -27,14 +28,15 @@ function AppLaunchWizard(props) {
...
-        <>
+        <Paper>
             <AppInfo
                 baseId={baseId}
                 loading={loading}
                 loadingError={appError}
                 app={app}
                 hasDeprecatedParams={hasDeprecatedParams}
             />
+            <Divider />
```
psarando, line 39: 👍

src/components/apps/launch/AppLaunchForm.js
```diff
@@ -574,104 +596,119 @@ const AppLaunchForm = (props) => {
...
-
-                                {values.requirements && (
-                                    <ResourceRequirementsReview
-                                        baseId={stepIdReview}
-                                        requirements={values.requirements}
+                                </StepContent>
+                            </PageWrapper>
+
+                            <PageWrapper appBarHeight={appBarHeight}>
+                                <StepContent
```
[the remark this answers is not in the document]
sri2k1us: I thought it was MUI component somehow looking at it name.

src/components/apps/launch/AnalysisInfoForm.js
```diff
@@ -62,18 +62,6 @@ const AnalysisInfoForm = ({ formId, appType, startingPath }) => (
...
-                    ids.LAUNCH_ANALYSIS_GROUP,
-                    ids.APP_LAUNCH_RETAIN_INPUTS
-                )}
-                label={getMessage("retainInputsLabel")}
-                name="debug"
-                component={FormCheckbox}
-            />
-        )}
     </>
 );
```
sri2k1us, line 67: Removed `Retain Inputs` and added helper text for analysis name.

src/components/apps/launch/AppInfo.js
sri2k1us: Actually the `Paper` here was excess and not required. So I removed it. But I needed an element to set `ref` to calculated `AppInfo` bar height dynamically. So added a `div`.

src/components/apps/launch/AppLaunchForm.js
```diff
@@ -48,100 +51,108 @@ import {
...
-            <BottomNavigationAction
-                id={buildDebugId(formId, ids.BUTTONS.STEP_NEXT)}
-                label={getMessage("next")}
-                value="next"
-                icon={<ArrowForward />}
-            />
-        )}
-    </BottomNavigation>
-);
+const StepperBottomNavigation = React.forwardRef((props, ref) => {
```
sri2k1us, line 97: Needed to set `ref` for functional components. See https://github.com/reactjs/reactjs.org/issues/2120

src/components/apps/launch/AppLaunchForm.js
```diff
@@ -427,6 +441,26 @@ const AppLaunchForm = (props) => {
...
+        (isMobile ? MOBILE_NAVIGATION_BAR_HEIGHT : NAVIGATION_BAR_HEIGHT);
+
+    React.useEffect(() => {
+        setStepperRef(stepperRef);
+    }, [stepperRef, setStepperRef]);
+
+    React.useEffect(() => {
+        setBottomNavRef(bottomNavRef);
+    }, [bottomNavRef, setBottomNavRef]);
+
```
sri2k1us, line 463: Calculate `Height` for `AppInfo, Stepper and Bottom Nav bar` dynamically.

src/components/apps/launch/ParamGroups.js
sri2k1us: This change is done. 
image

image

sri2k1us: I have used `info` color and contrast text. The section numbers are under the `group` title. If anyone has a feedback, please bring it up in next UX meeting.

src/components/apps/launch/messages.js
```diff
@@ -32,6 +32,7 @@ export default {
         advancedSettings: "Advanced Settings (optional)",
         analysisInfo: "Analysis Info",
         analysisName: "Analysis Name",
+        analysisNameHelp: "Enter analysis name",
```
psarando, line 35: I think this could use more elaboration. Maybe let the user know that this analysis name will also become the first part of the output folder name (by default), and they can change it here.

sri2k1us:
If it not problem, I can address issues with Mobile view in a separate PR since it's not a road block to anyhting.I anticipate a bigger change for that. I will go head merge this once I fix the skeleton.

src/components/apps/launch/messages.js
sri2k1us: Not sure how to convey that concisely here. Might want to get that sorted out. I will defer that to Ux meeting.

sri2k1us:
May be https://material-ui.com/components/steppers/#mobile-stepper for Mobile view?

src/components/apps/launch/AppLaunchForm.js
ash3rz: I don't believe these useEffects are needed.  `useComponentHeight` already has a useEffect for when the ref changes

src/components/apps/launch/AppLaunchForm.js
```diff
@@ -48,100 +51,108 @@ import {
...
 } from "@material-ui/core";
 
 import { ArrowBack, ArrowForward, PlayArrow, Save } from "@material-ui/icons";
 
 const useStyles = makeStyles(styles);
 
+// sonora appbar height.
+//Stepper height, appinfo height and bottom nav height is calculated dynamically.
+const NAVIGATION_BAR_HEIGHT = 105;
+const MOBILE_NAVIGATION_BAR_HEIGHT = 75;
```
ash3rz, line 74: Probably not for this PR, but a later PR - the Sonora app bar height isn't hardcoded in `_app` or `CyVerseAppBar`, it's also figured out dynamically.  I'd like us to keep with that pattern if MUI ever decides to change these heights.

src/components/apps/launch/AppLaunchWizard.js
```diff
@@ -25,25 +28,33 @@ function AppLaunchWizard(props) {
             deprecatedParamTypes.includes(param.type)
         )
     );
+    const appInfoRef = React.useRef(null);
+    const [appInfoHeight, setAppInfoRef] = useComponentHeight();
+
+    React.useEffect(() => {
+        setAppInfoRef(appInfoRef);
+    }, [appInfoRef, setAppInfoRef]);
```
ash3rz, line 36: This useEffect should also not be needed

ash3rz: Sorry, these comments weren't super clear - you still need to do `setAppInfoRef(appInfoRef)`, but that line doesn't need to be within a useEffect, is what I meant.

src/components/apps/launch/AppLaunchForm.js
sri2k1us: Feels like we talked about this before - https://github.com/cyverse-de/sonora/pull/126#discussion_r447908442

src/components/apps/launch/AppLaunchForm.js
sri2k1us: Ya we have to figure out how to pass down the ref between appbar  and app launch.

src/components/apps/launch/AppLaunchForm.js
ash3rz: Ah, I forgot about that convo. Does the same error happen here?  I'm unsure as to why that could be 😕

sri2k1us: Removing it actually pushed the nav bar out of bounds. Surprisingly I didn't get an exception in console 😕 . I am going to save the investigation step for a later time. For now am gonna leave the `useEffect` in place.

sri2k1us:
Thanks for the reviews.

src/components/apps/launch/AppLaunchForm.js
sri2k1us: Small update - putting the `setBottomNavRef(bottomNavRef);` outside the `useEffect` does result in `too many rerenders` error. When I wrote the previous commented out the whole `useEffect` without moving `setBottomNavRef(bottomNavRef);`.

PR #148, opened 2020-07-29
CORE-1104 Update Sonora for async tasks

ash3rz:
The only move action from the DE we have right now in Sonora is delete.  I made the message shorter than what's in the DE because I wanted the announcer to be small enough for mobile.  Any wording suggestions there are welcome.

I also started converting Data over to using react-i18next.  This PR only has the Data Listing converted.  Next PR will have all of Data converted, just still need to validate nothing broke.


ash3rz:
Thanks for the review! Since this is a pretty small PR, I'll go ahead and merge, though additional reviews still welcome!

PR #150, opened 2020-07-30
More i18n

sri2k1us:
- Clean up layout and search 
- Port `analyses ` to `i18next`.


src/pages/_app.js
```diff
@@ -166,3 +166,13 @@ function MyApp({ Component, pageProps }) {
...
+
+MyApp.getInitialProps = async ({ Component, ctx }) => {
+    let pageProps = {};
+
+    if (Component.getInitialProps) {
+        pageProps = await Component.getInitialProps(ctx);
+    }
+
+    return { pageProps, namespacesRequired: ["common"] };
+};
```
sri2k1us, line 178: I guess this is the important piece of puzzle. Since our layout component is included in every page, adding a `namespacesRequired` seems to remove the warnings. I haven't updated the next-i18next version. That could be handled in separate PR.

src/pages/search.js
```diff
@@ -21,3 +21,13 @@ export default function Search() {
...
 }
+
+Search.getInitialProps = async ({ Component, ctx }) => {
+    let pageProps = {};
+
+    if (Component.getInitialProps) {
+        pageProps = await Component.getInitialProps(ctx);
+    }
+
+    return { pageProps, namespacesRequired: ["common"] };
```
ash3rz, line 32: The other pages have a shortened version of the `getInitialProps` function - is there a reason this one is different?

sri2k1us: Nope..I just forgot to update this.

sri2k1us:
Thanks for the reviews.

public/static/locales/en/analyses.json
```diff
@@ -0,0 +1,83 @@
...
+    "copyAnalysisId":"Copy Analysis ID",
+    "currentStatus":"Current Status",
+    "date":"Date",
+    "delete":"Delete",
+    "details":"Details",
+    "email":"Email",
+    "emptyValue":"-",
+    "endDate":"End date",
+    "extendTime":"Extend time limit",
+    "extendTimeLimitMessage":"This analysis is scheduled to be terminated at {timeLimit}. Do you wish to extend the time limit?",
```
ash3rz, line 35: Missed one more spot 😉

sri2k1us: oops..I have #154 on top this...Will fix there....

PR #151, opened 2020-07-30
added some debugging code to try to pinpoint the source of the auth t…

slr71:
…imeout problem

This is intended to be a temporary change. I haven't been able to reproduce the authentication timeout issue locally, so I'm hoping that we can pinpoint it by adding some logging and an endpoint that we can call to get detailed information about the authentication. This code can go away as soon as we identify the source of the problem.

### DIFF
--- a/src/server/api/debug.js
+++ b/src/server/api/debug.js
@@ -1,0 +1,30 @@
+import express from "express";
+
+import logger from "../logging";
+import c from "cookie";
+
+export default function debugRouter() {
+    const api = express.Router();
+
+    logger.info("************ Adding Debug handlers **********");
+
+    logger.info("adding the GET /apps/categories handler");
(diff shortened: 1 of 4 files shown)


src/server/api/debug.js
```diff
@@ -0,0 +1,30 @@
...
+
+import logger from "../logging";
+import c from "cookie";
+
+export default function debugRouter() {
+    const api = express.Router();
+
+    logger.info("************ Adding Debug handlers **********");
+
+    logger.info("adding the GET /apps/categories handler");
```
ash3rz, line 11: Might be good to remove this line to avoid confusion

slr71: Oops! I just removed that line.

slr71:
Thanks for the reviews!

PR #152, opened 2020-07-31
Convert Data components to using react-i18next

ash3rz:
Should be straightforward, the only weird thing is where App Launch and Data are sharing the same logic for disk resource name validation.  I updated the validate function to accept either the `t` function from react-i18next or `getMessage` to work with react-intl until App Launch can be converted over.


src/components/data/utils.js
```diff
@@ -32,12 +31,12 @@ function getEncodedPath(path) {
 const getFolderPage = (folderPath) =>
     `${constants.PATH_SEPARATOR}${NavigationConstants.DATA}${constants.PATH_SEPARATOR}ds${folderPath}`;
 
-const validateDiskResourceName = (name) => {
+const validateDiskResourceName = (name, i18nFn) => {
```
sri2k1us, line 34: 👍

ash3rz:
I think we've got the i18n stuff figured out now and I'm not seeing any warnings in the console.  I'll go ahead and merge this, but reviews still welcome!

PR #155, opened 2020-08-05
CORE-1116: detect when a user has logged out and update the CyVerseAppBar accordingly.

slr71:
The one case that isn't handled at this time is the case where the user logs out and is on an authenticated page. We'd ideally like to leave the user on that page and display the login error. The catch is that we need to be able to detect when the user is on a page that requires authentication vs. one that doesn't. One idea might be to invalidate all `react-query` queries. I _think_ that would have the desired effect, I don't think that there's a good way to invalidate all queries without knowing the first element in every query key.

We could consolidate the query key definitions and then write code to invalidate each set of queries, but that's a big change. I'm not sure if now is the best time to tackle that or not.

At any rate, if anyone has any other ideas, I'd love to hear them.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -15,6 +15,9 @@ module.exports = withBundleAnalyzer({
         ADMIN_GROUPS: config.get("admin.groups"),
         ADMIN_GROUP_ATTRIBUTE: config.get("admin.group_attribute_name"),
         IRODS_HOME_PATH: config.get("irods.home_path"),
+        SESSION_POLL_INTERVAL_MS: config.has("sessions.poll_interval_ms")
+            ? config.get("sessions.poll_interval_ms")
+            : null,
         TOOLS_PRIVATE_MAX_CPU_LIMIT: config.has("tools.private.max_cpu_limit")
             ? config.get("tools.private.max_cpu_limit")
             : 8,
--- a/next.config.js
+++ b/next.config.js
@@ -17,7 +17,7 @@ module.exports = withBundleAnalyzer({
         IRODS_HOME_PATH: config.get("irods.home_path"),
         SESSION_POLL_INTERVAL_MS: config.has("sessions.poll_interval_ms")
             ? config.get("sessions.poll_interval_ms")
-            : null,
+            : 5000,
         TOOLS_PRIVATE_MAX_CPU_LIMIT: config.has("tools.private.max_cpu_limit")
             ? config.get("tools.private.max_cpu_limit")
             : 8,
--- a/src/components/layout/CyVerseAppBar.js
+++ b/src/components/layout/CyVerseAppBar.js
@@ -27,7 +27,6 @@ import {
     bootstrap,
     BOOTSTRAP_KEY,
     USER_PROFILE_QUERY_KEY,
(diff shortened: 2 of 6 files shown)


next.config.js
```diff
@@ -15,6 +15,9 @@ module.exports = withBundleAnalyzer({
         ADMIN_GROUPS: config.get("admin.groups"),
         ADMIN_GROUP_ATTRIBUTE: config.get("admin.group_attribute_name"),
         IRODS_HOME_PATH: config.get("irods.home_path"),
+        SESSION_POLL_INTERVAL_MS: config.has("sessions.poll_interval_ms")
+            ? config.get("sessions.poll_interval_ms")
+            : null,
```
sri2k1us, line 20: We can return a `default` value of `5000` here instead of returning null like how we did for resource (CPU, memory etc..) defaults. That would eliminate the changes in `serviceFacades/users.js`.

slr71: I have to admit that I couldn't get myself to let go of the check for a null value where the configuration was being used. I thought about putting the default value both here and in `serviceFacades/users.js`, but that felt wrong to me. I also thought about defining a default value as a constant here and including this file in `CyVerseAppBar.js`. That felt wrong too. My brain eventually rebelled and I returned null here so that I could let the null check handle the case where the configuration value was undefined. That option seemed the least repulsive to my brain at the time.

As hard as it is for me to do, I think I can convince myself to let go of the null check in `CyVerseAppBar.js`. The upshot is that I'm not just a programmer. I'm a _crazy_ programmer. 🤪

sri2k1us: Feel free to merge as it is..It was just a suggestion. So no biggie.

slr71: I think you're right. My brain was just being stubborn yesterday. 😆

src/components/layout/CyVerseAppBar.js
```diff
@@ -231,8 +231,7 @@ function CyverseAppBar(props) {
         config: {
             enabled: true,
             onSuccess: setUserProfile,
-            cacheTime: Infinity,
-            staleTime: Infinity,
+            refetchInterval: clientConfig.sessions.poll_interval_ms,
```
slr71, line 234: I ended up changing the `?.` operators to simply `.` because the value should be defined everywhere.

slr71:
While testing this change again, I noticed that there were several errors appearing in storybook. I don't think that the errors are related to these changes, though.

slr71:
Thanks for the review!

slr71:
Er...thanks for the reviews!

src/components/layout/CyVerseAppBar.js
sri2k1us: @slr71 - removing the `?` throws an error on `Sonora` start-up. I guess `clientConfig` prop could be `null` during startup ? 🤔

sri2k1us: You could run the profile query after `config` state is set.

slr71: I didn't notice that error. Thanks for catching that. I'll update it today.

PR #156, opened 2020-08-06
next-i18next fixes

ash3rz:
This ended up being more changes than I expected 😬 To be honest, I have no idea if all of these changes are needed to fix storybook, but I wanted to start over and eliminate all possible variables, so we hopefully don't run into any issues for a while.

* I bumped up to the latest version of `next-i18next` (had to remove the server side middleware, which I don't think we were using)
* Followed the `next-i18next` docs and more importantly the [sample](https://github.com/isaachinman/next-i18next/tree/master/examples/simple) which uses the `i18n` file with the `next-i18next` instance for everything.  We were trying to use `react-i18next` in our pages and components instead of our instance.  The docs say that `next-i18next` uses `react-i18next` "under the hood" so this may have been causing issues.
* The docs mentioned two config parameters that are required that we didn't have - `localePath` and `otherLanguages`
* Took inspiration from a [github issue](https://github.com/isaachinman/next-i18next/issues/324) where someone else was trying to set up storybook.  I turned that into a wrapper component that storybook and jest can both use.

I ran the tests, ran storybook, ran dev, ran a prod build, and didn't see any issues.  Please try this on your machines too.


guidelines.md
```diff
@@ -105,7 +105,7 @@ Generally, you'll need to:
 3. Once you have imported the translation files, you can use them in your react component.
 ```
 import React from "react";
-+import { useTranslation } from "react-i18next";
++import { useTranslation } from "i18n";
```
sri2k1us, line 108: 👍

src/i18n.js
```diff
@@ -1,5 +1,50 @@
...
+ * @returns {JSX.Element}
+ * @constructor
+ */
+function I18nProviderWrapper({ children }) {
+    return (
+        <Suspense fallback={"Loading..."}>
+            <I18nextProvider i18n={i18n}>{children}</I18nextProvider>
+        </Suspense>
+    );
+}
```
sri2k1us, line 41: 👍

ash3rz:
Thanks for the reviews and for checking it out on your machines too.  Hopefully it doesn't blow up now on Jenkins 🤞

PR #157, opened 2020-08-07
CORE-1116: prevent polling for the user profile until after the confi…

slr71:
…guration settings have been loaded

I'm not really satisfied with this PR, but I wanted to push these changes as quickly as possible. I'm going to continue to try to find a better solution.


slr71:
Thanks for the review!

slr71:
Thanks for the review!

PR #158, opened 2020-08-08
CORE-777 Add Delete menu item to Analyses listing

psarando:
This PR will add a `Delete` menu item to the Analyses listing's dot menu, add `/analyses/shredder` API handlers, and correctly invalidate all analyses listing queries after a relaunch (bonus fix).


src/components/analyses/listing/DeleteConfirmationDialog.js
```diff
@@ -0,0 +1,65 @@
...
+    DialogTitle,
+    useMediaQuery,
+    useTheme,
+} from "@material-ui/core";
+
+import { build } from "@cyverse-de/ui-lib";
+
+import ids from "../ids";
+
+function DeleteConfirmationDialog({ baseId, open, onClose, confirmDelete }) {
```
psarando, line 24: This is almost an exact copy of the `src/components/analyses/listing/MultiRelaunchWarningDialog.js`. The only difference being the base debug ID, dialog title and content text, and the OK button's `onClick` confirmation function.

Maybe these should be refactored to use a common, generic confirmation dialog.

src/components/analyses/listing/Listing.js
```diff
@@ -374,12 +396,13 @@ function Listing(props) {
...
                 handleGoToOutputFolder={handleGoToOutputFolder}
+                handleDelete={handleDelete}
                 handleRelaunch={handleRelaunch}
                 handleBatchIconClick={handleBatchIconClick}
             />
             <TableView
-                loading={isFetching || relaunchLoading}
-                error={error || relaunchError}
+                loading={isFetching || deleteLoading || relaunchLoading}
+                error={error || deleteError || relaunchError}
```
sri2k1us, line 405: Sorry I missed this during `relaunch` PR but `errors` which happen during mutations should be handled with `announcer`. Blocking the listing will prevent the users from retrying these operations.
 I think we agreed during our alpha test review meeting that 
1. Any listing error should be redirected to `error.js` page.Users can then simply hit browser `Back` button to retry the listing.
2. Any other error should be handled by displaying an `announcer` with ` Details` button if applicable.

public/static/locales/en/common.json
```diff
@@ -15,11 +16,12 @@
     "logout": "Logout, {{name}}",
     "more": "More",
     "newNotificationAriaLabel": "Show New Notifications",
+    "ok": "OK",
```
ash3rz, line 19: 👍

psarando:
Thanks for the reviews! 👍

PR #159, opened 2020-08-11
Global Search

sri2k1us:
Notes:
- Detailed search result page is not implemented yet.
- Drop down results capped to `10` items per category.
- Selecting a file from search results will display the parent folder since file viewer is not implemented yet.

- Grouped results
image

image

- iPhone 5 / SE
image

image

- No results
image

- Filtered by apps (or search only apps):
image


src/components/apps/toolbar/Toolbar.js
```diff
@@ -121,6 +124,18 @@ function AppsToolbar(props) {
...
+                        variant="outlined"
+                        disableElevation
+                        color="primary"
+                        onClick={viewAllApps}
+                        startIcon={<FilterListIcon />}
+                    >
+                        {getMessage("viewAllApps")}
+                    </Button>
+                )}
                 <div className={classes.divider} />
```
sri2k1us, line 139: If the user is viewing an individual app listing from Search, then they can click on this button to view all apps from there.

src/components/search/GlobalSearchField.js
```diff
@@ -3,139 +3,485 @@
...
+    let path = selectedOption._source.path;
+
+    //SS route to parent folder for the file util we have file viewers ready in sonora.
+    if (type === ResourceTypes.FILE) {
+        path = getParentPath(path);
+        icon = <DescriptionIcon />;
+    }
+
+    const href = `/${NavigationConstants.DATA}/${constants.DATA_STORE_STORAGE_ID}`;
+    const as = `/${NavigationConstants.DATA}/${constants.DATA_STORE_STORAGE_ID}${path}`;
```
sri2k1us, line 162: Tricky. Adding `[...pathItems]` to the `href` will cause whole app to refresh on selecting data result.

src/components/search/GlobalSearchField.js
```diff
@@ -3,139 +3,485 @@
...
+    return (
+        <Link href={href} as={as} passHref>
```
sri2k1us, line 164: more implementation details - https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-function-component

src/components/search/GlobalSearchField.js
```diff
@@ -3,139 +3,485 @@
...
+            t("searchError"),
+            analysesSearchError || appsSearchError || dataSearchError
+        );
+    }
+
+    const loading = searchingAnalyses || searchingApps || searchingData;
+
+    const renderCustomOption = (option) => {
+        switch (option?.resultType) {
+            case t("data"):
```
ianmcorvidae, line 387: Unless I'm mistaken, this `resultType` field isn't rendered anywhere, and given that I think we shouldn't be translating it for use in conditionals like this. In particular, it creates more potential for weird bugs, especially if and when we actually start getting things translated. For example, what happens if the server side doesn't have correct language information while doing SSR, but then the client side tries to do react rehydration using a specific language and produces a different result? Or, what happens if we're switching or branching on a string such that multiple things are translated to the same thing and the conditional loses a branch (don't think that's likely here, but anyway)? These sorts of bugs involving the i18n system or language-switching can't happen if these sorts of things that users don't see just use English, or opaque values instead of any particular language (though that can make writing the code harder, of course).

src/components/search/GlobalSearchField.js
```diff
@@ -3,139 +3,485 @@
...
+                id={build(ids.SEARCH, ids.SEARCH_FILTER_MENU)}
+                value={filter}
+                onChange={handleFilterChange}
+                disableUnderline
+                className={classes.searchFilter}
+                size="small"
+                variant={isMobile ? "outlined" : "standard"}
+                renderInput={() => <TextField size="small" disableUnderline />}
+            >
+                <MenuItem value={t("all")}>{t("all")}</MenuItem>
```
ianmcorvidae, line 478: These `value`s are like the conditionals too, I think, best if we can avoid those being translated strings (the text inside the `MenuItem` should be, of course, just not the `value`)

src/pages/analyses.js
```diff
@@ -24,7 +25,8 @@ export default function Analyses() {
     const selectedPage = parseInt(query.selectedPage) || 0;
     const selectedRowsPerPage = parseInt(query.selectedRowsPerPage) || 25;
     const selectedOrder = query.selectedOrder || constants.SORT_DESCENDING;
-    const selectedOrderBy = query.selecetdOrderBy || "startdate";
+    const selectedOrderBy =
+        query.selecetdOrderBy || analysisFileds.START_DATE.key;
```
ianmcorvidae, line 29: `selecetd` another typo? Looks like it existed before this PR though

src/components/search/GlobalSearchField.js
sri2k1us: I can easily switch to using non translated values but I still don't understand how translated values will cause a problem.

sri2k1us: When i18n stuff is broken, our app becomes unusable. For example, the drop down for the filter won't show proper values.

sri2k1us: One last thing ..our default language is set to `en` in our i18n configs
` defaultLanguage: "en",`

src/components/search/dataSearchQueryBuilder.js
```diff
@@ -0,0 +1,61 @@
...
+) => {
+    if (userHomeDir) {
+        return {
+            query: {
+                all: [
+                    {
+                        type: "label",
+                        args: {
+                            exact: false,
+                            negated: false,
```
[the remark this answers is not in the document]
sri2k1us: I believe it is being by search service. Here is the results for comparison
- without `negated`
image

- with negated.
image

src/components/search/GlobalSearchField.js
sri2k1us: Oops I almost forgot, the `resultType` is actually rendered. The `group label` you see in the search result is the `resultType`

src/components/search/GlobalSearchField.js
sri2k1us: I Updated the `value` to be a `const` here.

src/components/search/GlobalSearchField.js
ianmcorvidae: If it's being displayed, then for sure it should stay as the translated value. We can worry about the potential bugs I'm talking about if they come up later.

src/components/search/dataSearchQueryBuilder.js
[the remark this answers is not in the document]
ianmcorvidae: I'm not sure what's up here. The searches being created by the backend should be exactly the same with or without `negated` -- it appears that maybe some searches don't include the `path` clause in the backend though -- here's the queries from your user, today:

```
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"tes\"]","service":"search","time":"2020-08-13T06:19:46-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\"]","service":"search","time":"2020-08-13T06:19:46-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"tes\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T06:20:44-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T06:20:45-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriramNaN\"]","service":"search","time":"2020-08-13T06:30:56-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"tes\",path=\"/iplant/home/sriramNaN\"]","service":"search","time":"2020-08-13T06:30:56-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"tes\",path=\"/iplant/home/sriramNaN\"]","service":"search","time":"2020-08-13T06:32:10-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"tes\",path=\"/iplant/home/sriramNaN\"]","service":"search","time":"2020-08-13T06:32:32-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriramNaN\"]","service":"search","time":"2020-08-13T06:32:32-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[None:[path=\"/iplant/home/shared\"],label~\"test\"]","service":"search","time":"2020-08-13T06:33:36-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriramNaN\"]","service":"search","time":"2020-08-13T06:37:43-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"tes\",path=\"/iplant/home/sriramNaN\"]","service":"search","time":"2020-08-13T06:37:43-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"tes\",path=\"/iplant/home/sriramNaN\"]","service":"search","time":"2020-08-13T06:40:53-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriramNaN\"]","service":"search","time":"2020-08-13T06:40:54-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"tes\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T06:42:48-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T06:42:48-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"tes\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T06:47:12-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T06:47:13-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"tes\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T06:58:08-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T06:58:09-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"tes\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T06:58:14-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T06:58:14-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"tes\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T06:58:27-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T06:58:27-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"tes\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T10:39:03-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T10:39:03-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T10:39:04-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\"]","service":"search","time":"2020-08-13T10:45:02-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T10:45:04-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\"]","service":"search","time":"2020-08-13T10:46:05-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T10:46:09-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T10:47:40-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T10:47:49-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T10:47:55-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\"]","service":"search","time":"2020-08-13T11:05:57-07:00"}
{"art-id":"search","group":"org.cyverse","level":"info","msg":"Got a request from user sriram: All:[label~\"test\",path=\"/iplant/home/sriram/\"]","service":"search","time":"2020-08-13T11:06:05-07:00"}
```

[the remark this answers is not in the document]
sri2k1us: Ahh I see what is the confusion..I miss understood that you were talking about the whole `path` clause for some reason. I see you are just talking about the `negated` key ? Is that right ?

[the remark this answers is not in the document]
ianmcorvidae: Yeah, that's correct. Just the `negated` key in each clause should be unnecessary, since that key isn't used -- the rest of the clauses should stick around.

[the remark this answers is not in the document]
sri2k1us: Makes sense. Sorry about the confusion.

src/components/search/GlobalSearchField.js
```diff
@@ -3,139 +3,495 @@
...
+                }}
+                open={open}
+                debug={false}
+                onOpen={() => {
+                    setOpen(true);
+                }}
+                onClose={() => {
+                    setOpen(false);
+                }}
+                id="search"
```
psarando, line 460: Should this come from an `ids` constant?

src/components/search/dataSearchQueryBuilder.js
```diff
@@ -0,0 +1,61 @@
...
+        };
+    } else {
+        return {
+            query: {
+                all: [
+                    {
+                        type: "label",
+                        args: {
+                            exact: false,
+                            negated: false,
```
psarando, line 45: It seems like the only difference between these objects is an extra "path" object in the `all` array? Except this "label" object has a `negated` arg, but I thought that wasn't needed anymore?

If they should be the same, then it might be easier to see this if this query in the `else` section was refactored out if the conditional, and into something like `const searchQuery = { query: ...}`, and then `if (userHomeDir) searchQuery.query.all.push({type: "path" ...})`, and finally `return searchQuery`.

ianmcorvidae: Yeah, `negated` can go away here too. Just since I like thinking about different ways of organizing code, here's my thought:

```javascript
const searchClauses = [ { type: "label", args: { exact: false, label: searchTerm } } ]
return {
  query: {
    all: (userHomeDir ?
      searchClauses.concat({ type: "path", args: { prefix: userHomeDir } }) : 
      searchClauses),
  },
  size: rowsPerPage,
  from: offset,
  sort: [ { field: sortField, order: sortDir } ]
}
```

I don't know whether it's clearer inlining the if with a ternary operator or doing it your way, though.

src/components/search/GlobalSearchField.js
psarando: I'd agree that it seems fragile and risky to use translated strings in conditional checks like this, especially if we have the option to use a known constant instead.

For example, we could create this as an object instead of a simple string: `data.resultType = {label: t("data"), type: DATA}` and then here we could switch on `resultType.type`, and use `groupBy={(option) => option.resultType.label}` for the Autocomplete component. I haven't used that component yet, so I'm only guessing this could work, but I've seen similar logic checks using translated strings for an Autocomplete component before and it bothered me there as well.

Then even if someone made all these keys translate to "X" for some reason, it wouldn't break the logic here (even though it would make the UX less usable).

I think we should at least keep this in mind going forward.

sri2k1us:
Thanks for the reviews.

PR #161, opened 2020-08-15
CORE-779 Add Analysis Rename support to listing.

psarando:
This PR will add Analysis Rename support to the analyses listing.

A dialog will display with a form field for renaming the analysis.

When the user selects "OK" to confirm the rename, the field will become read-only and display a loading spinner at the end of the field (similar to the "create folder" dialog). The listing behind the dialog will also display a loading skeleton until the request is complete:
Rename loading

If an error is encountered during the request, the rename dialog will remain open so the user can try again:
Rename error announcer


src/components/analyses/listing/Listing.js
```diff
@@ -288,19 +305,19 @@ function Listing(props) {
...
-            setDetailsAnalysis(
+            setSelectedAnalysis(
                 data.analyses.find((item) => item.id === selectedId)
             );
         } else {
-            setDetailsAnalysis(null);
+            setSelectedAnalysis(null);
         }
-    }, [data, detailsOpen, selected]);
+    }, [data, selected]);
```
psarando, line 320: I started refactoring this `isSingleSelection` and `selectedAnalysis` state because I thought I would need them for enabling the rename menu item in the listing dot menu, but once I got to that part of the code I realized the other menu items are enabling and looking up analysis info in a different way.

I didn't bother undoing what I refactored so far, but I can still change it back.
I don't know, the code might be ever so slightly cleaner this way 🤷‍♂️

src/components/analyses/toolbar/AnalysesDotMenu.js
```diff
@@ -184,26 +202,25 @@ function DotMenuItems(props) {
...
-    } = props;
-
-    const isSingleSelection = selected?.length === 1;
+function AnalysesDotMenu({
+    ButtonProps,
+    username,
+    getSelectedAnalyses,
+    ...props
+}) {
+    const { baseId, isSingleSelection } = props;
```
psarando, line 211: These `baseId` and `isSingleSelection` props need to be spread down into the `DotMenuItems` component below, but the props above do not.

This was also an attempt at removing the `Function components cannot be given refs` warning, but now I'm not sure it's related to the `AnalysesDotMenu` props 🤔

stories/analyses/Listing.stories.js
```diff
@@ -55,10 +55,34 @@ function ListingTest(props) {
     );
 }
 
+const errorResponse = {
+    error_code: "ERR_GOOD_NEWS_EVERYONE",
+    reason: "This error will only occur once! Please try again...",
+};
```
psarando, line 61: :farnsworth:

src/components/analyses/RenameAnalysisDialog.js
```diff
@@ -0,0 +1,117 @@
...
+    InputAdornment,
+} from "@material-ui/core";
+
+function RenameAnalysisDialog(props) {
+    const { open, isLoading, onClose, handleRename, selectedAnalysis } = props;
+
+    // Using separate hooks so that analyses and data message keys
+    // don't require a prefix.
+    const { t } = useTranslation(["analyses", "common"]);
+    const { t: dataI18N } = useTranslation("data");
```
sri2k1us, line 34: I have similar issue in my code. I can't use array syntax as It would only pick the first namespace in the array 😕. Did you encounter any issue as well ?

src/components/analyses/RenameAnalysisDialog.js
```diff
@@ -0,0 +1,117 @@
...
+                                    component={FormTextField}
+                                />
+                            </DialogContent>
+
+                            <DialogActions>
+                                <Button
+                                    id={build(baseId, UtilIds.DIALOG.CANCEL)}
+                                    onClick={onClose}
+                                >
+                                    {t("common:cancel")}
```
sri2k1us, line 98: 👍 . May be I should try this syntax.

src/components/analyses/listing/Listing.js
```diff
@@ -129,6 +133,19 @@ function Listing(props) {
         },
     });
 
+    const [renameAnalysesMutation, { isLoading: renameLoading }] = useMutation(
+        renameAnalyses,
+        {
+            onSuccess: () => {
+                queryCache.invalidateQueries(ANALYSES_LISTING_QUERY_KEY);
+                setRenameDialogOpen(false);
```
sri2k1us, line 141: I wonder if we can move away from reloading the view on mutations like rename, delete etc...since react-query gives us option to do optimistic update. See - https://react-query.tanstack.com/docs/guides/updates-from-mutation-responses
https://react-query.tanstack.com/docs/guides/optimistic-updates
But this can be tried in another PR.

src/components/analyses/listing/Listing.js
sri2k1us: `singleSelection` makes sense to me. We can refactor the rest of the code when encounter more changes / updates to `dot menu`.

src/components/analyses/listing/Listing.js
psarando: Ah yes, good idea 👍 
I forgot about that for rename mutations; but I didn't want to do this for deletes, since the listing can be paged, and removing 1 or more analyses from the listing will change the view for all pages, so I thought it was best to reload the listing in that case.

psarando:
> Can you post a screenshot of the error announcer with the rename dialog for `iphone 5`? The reason I ask is, for `rename` in `Data`, I suggested we inline the errors in the dialog itself since it was difficult to read the message in announcer over the dialog on mobile screens.

Sure. It looked OK to me, so I didn't include it in the first place:
Rename error announcer in mobile

The announcer looks fine to me, although the error dialog info looks a little cramped on mobile (opened by clicking `details`), but maybe that could be improved in the error dialog itself (in a separate PR).

I did notice the inline error message in the create folder dialog, but I thought showing this announcer was better since it gives the `details` button where the user can view the actual error from the service if needed (in case they do try again and keep getting an error, for example).

src/components/analyses/RenameAnalysisDialog.js
psarando: Yep, unfortunately, according to the documentation it will only look up the key in the first namespace of the array if the key doesn't have a prefix:
https://react.i18next.com/latest/usetranslation-hook#usetranslation-params

That's why below I have to use `t("common:cancel")` and `t("common:ok")`. The messages used in `validateDiskResourceName` can't use the `data:` namespace yet, since apps is still using `getMessage` for now, but once that's converted with `useTranslation`, then I think the `data` namespace can be used in that function and also included in this array.

src/components/analyses/listing/Listing.js
ash3rz: That's a good point.  Wouldn't the same be true for renames as well then? If I have 50 pages and rename `zzzFile` to `aaaFile`, for example, then maybe a reload is still best in that case

ianmcorvidae: It seems to me like we should do both things most of the time. For a synchronous rename/delete, we can either update or remove the entry (possibly also remove with a rename or with some renames -- e.g. for the zzzFile->aaaFile thing, if we know that aaaFile wouldn't be on the same page) so that the user gets immediate feedback and initiate a background update of the query that re-updates the display when finished; for async stuff, the same, but based on the 'complete' notification (except if we ever want to do some sort of "task in progress" indicator in listings, which would presumably get put in place immediately for async stuff).

I actually think this might be easier for deletes, because if you're deleting from a page you can largely assume that any newly-visible rows will be at the bottom of the page (page of a, b, c, d; four more after that, if you delete b and d you can show a page of a, c, and two loading skeletons until the query figures out that it should be showing a, c, e, f). With a rename you'd need to reproduce the sorting and make a guess about whether the new name will come before the page (meaning a first-row loading skeleton, probably), after the page (last-row skeleton), or elsewhere within the same page such that it can be updated to that place in the page. (renames are easy if the sorting isn't by name at all, though)

sri2k1us:
> > Can you post a screenshot of the error announcer with the rename dialog for `iphone 5`? The reason I ask is, for `rename` in `Data`, I suggested we inline the errors in the dialog itself since it was difficult to read the message in announcer over the dialog on mobile screens.
> 
> Sure. It looked OK to me, so I didn't include it in the first place:
> Rename error announcer in mobile
> 
> The announcer looks fine to me, although the error dialog info looks a little cramped on mobile (opened by clicking `details`), but maybe that could be improved in the error dialog itself (in a separate PR).
> 
> I did notice the inline error message in the create folder dialog, but I thought showing this announcer was better since it gives the `details` button where the user can view the actual error from the service if needed (in case they do try again and keep getting an error, for example).

Ahh that is `Create Folder` dialog not rename. But I remember now why I suggested inlining it for Create Folder dialog. There were `validation messages` for invalid folder names being displayed inline. So I thought we could keep it consistent and display all the error messages in 1 place. Remember we could use `ErrorTypography` to inline error message with `Details` button.

src/components/analyses/listing/Listing.js
psarando: For renames I figured it was OK to just update the row in-place, since I don't think the user would be confused if that one row is out of place in the sort, because they were just updating it, and they can reload or navigate to a new page to get the true sort order?

For deletes, where the user could remove half the rows or more, I think it would look worse to leave a bunch of empty space, unless we did something more complicated as described above.

I figured out how to update the listing for a rename, leaving the updated row in-place regardless of the sort order, without refreshing the entire listing. I can push that change to this PR before merging if that's the direction we want to go for now.

psarando:
> Ahh that is `Create Folder` dialog not rename. But I remember now why I suggested inlining it for Create Folder dialog. There were `validation messages` for invalid folder names being displayed inline. So I thought we could keep it consistent and display all the error messages in 1 place. Remember we could use `ErrorTypography` to inline error message with `Details` button.

Here is a service error message inlined with the form field, where validations messages are displayed, for mobile:

Rename service error - mobile view

---

And here is mobile view with a normal validation error:

Rename validation error - mobile view

The long message plus the button (with `size="small"`) takes up more space than the regular validation messages, but I don't know if that's a big deal.

It might be easier to see the difference by opening these images in their own tabs and switching between them.

Here's the same error messages for desktop, where the difference is more subtle, except when you're interacting with the dialog (by adding and removing an invalid character for example):

Rename service error - desktop

---

Rename validation error - desktop

---

Also notice that I'm not setting the service error as a validation error, so the field itself isn't getting highlighted as invalid. This also means the service error message (and the `Details` button) will stick around until the submission is tried again (even if the user makes and fixes validation errors after a service error).

src/components/analyses/listing/Listing.js
ash3rz: Hm... not sure.  Seems reasonable to me to keep the renamed value in place, though I'm not sure I've seen that kind of behavior elsewhere. Maybe others have additional thoughts.

As for async, I keep forgetting about that 😖 I did the simplest thing for the DE - I think it refreshes and shows an announcer that the actual action will complete in the background - and then copied it over for deleting files/folders in  Sonora. That's definitely not at all set in stone or ideal, we can make it better in Sonora.

ash3rz:
OK, cool, this seems like the best of both worlds to me at least.  The Details button seems a little strange - maybe because it's bigger than the rest of the text now?  But that's probably not a change for this PR.

PR #162, opened 2020-08-17
Cart Support

johnworth:
Adds a dialog and set of custom hooks for adding and removing items from a bag and populating the sharing dialog from the bag.

## Limitations
Downloads do not work yet, so its button has been disabled. Adding items to the bag has not been wired into any other components yet.

## App Bar with Icon and Badge
Screen Shot 2020-10-01 at 10 05 00 AM

## Add To Bag in Data Toolbar
It only shows up when one or more items are selected.

image

## Add To Bag in Apps Toolbar
image

## Add To Bag in Analyses Toolbar
image

## Unfiltered Bag
image

## Downloadable Items Only
image

## Shareable Items Only
image

## Mobile Version
image

## Tablet Version
image

## Download Links Dialog Populated from Bag
image

## Download Links Mobile
image

## Download Links Tablet
image

## Sharing Dialog Populated from Bag
image


src/components/Bag/index.js
```diff
@@ -0,0 +1,540 @@
...
+        setBagDlgOpen(false);
+    };
+
+    const handleDownloadClick = (event) => {
+        event.preventDefault();
+        event.stopPropagation();
+
+        setDownloadDlgOpen(true);
+        setBagDlgOpen(false);
+    };
```
sri2k1us, line 360: We should check if there are items for download before opening the download link dialog.

src/components/Bag/downloads/index.js
```diff
@@ -0,0 +1,146 @@
...
+                                        path.lastIndexOf("/") + 1
+                                    )}
+                                    href={`${getHost()}/api/download?path=${path}`}
+                                >
+                                    {path}
+                                </a>
+                            </ListItem>
+                        );
+                    })}
+                </List>
```
sri2k1us, line 109: It will nice if we could provide a way to copy these directly from here. Copied links could then be used with a script or a download manger.

sri2k1us (approved):
LGTM 👍. I checkout the PR locally and everything seems to work although the UX seems have some rough edges. I have mentioned some of them in the comments. We can address this after we get feedback from phase II testing.

public/static/locales/en/bags.json
```diff
@@ -0,0 +1,19 @@
...
+    "bagHelp": "Perform sharing or downloading operations on a set of items.",
+    "close" : "Close",
+    "download": "Download",
+    "downloadLinks" : "Download Links",
+    "downloadLinksDescription" : "Click on the links to download each file through your browser or through a download manager.",
+    "share": "Share",
+    "clearBag": "Clear",
+    "fetchBagError": "Error fetching bag contents",
+    "removeAllItemsError": "Error clearing bag contents",
+    "removeItemError": "Error remove item from bag",
```
ash3rz, line 13: Should this be "Error **removing** item from bag" ?

src/components/Bag/classes.js
```diff
@@ -0,0 +1,120 @@
...
+    icon(t) {
+        return <Label />;
+    }
+
+    get label() {
+        return (
+            this.item?.label ||
+            this.item?.name ||
+            this.item?.description ||
+            "override label() please"
```
ash3rz, line 19: Is this message for debugging or will this actually display to the user?

src/components/Bag/classes.js
```diff
@@ -0,0 +1,120 @@
...
+    get shareable() {
+        return true;
+    }
+
+    get downloadable() {
+        return true;
+    }
+
+    get name() {
+        return this.item.path.substring(this.item.path.lastIndexOf("/") + 1);
```
ash3rz, line 50: In `/src/components/data/utils`, we have a function `parseNameFromPath` that does something similar.  It'd be good if we used the same function here, as well as for `FolderBagItem`

src/components/Bag/classes.js
```diff
@@ -0,0 +1,120 @@
...
+}
+
+export const FOLDER_TYPE = "folder";
+export const FILE_TYPE = "file";
+export const ANALYSIS_TYPE = "analysis";
+export const APP_TYPE = "app";
+
+export const createNewBagItem = (item) => {
+    if (!item.id) {
+        item.id = uuidv4();
```
ash3rz, line 102: If the item type is anything other than data, and (for some reason, I'm not sure how this would actually happen) it doesn't have an id, I don't think creating an ID will work 🤔 I'm just thinking of sharing, because in order to share apps, analyses, or tools, the sharing endpoints require the id.

src/components/Bag/downloads/index.js
```diff
@@ -0,0 +1,146 @@
...
+
+        onClose();
+    };
+
+    return (
+        <Dialog
+            fullScreen={fullScreen}
+            open={open}
+            onClose={handleClose}
+            classes={{ paper: classes.paper }}
```
ash3rz, line 63: It'd be good if the dialog itself had its own static ID as well, looks like it should be `dialogID` from what I can tell

src/components/data/toolbar/DataDotMenu.js
```diff
@@ -135,6 +140,21 @@ function DataDotMenu(props) {
...
+                                      onClick={() => {
+                                          onClose();
+                                          onAddToBagSelected();
+                                      }}
+                                  >
+                                      <ListItemIcon>
+                                          <AddToBagIcon fontSize="small" />
+                                      </ListItemIcon>
+                                      <ListItemText primary={t("addToBag")} />
+                                  </MenuItem>
```
ash3rz, line 156: There is also now the `RowDotMenu` which appears within each row in the data's table view.  You could also add this same menu item there.

src/serviceFacades/bags.js
```diff
@@ -0,0 +1,222 @@
+import { useMutation, useQuery, useQueryCache } from "react-query";
+
+import callAPI from "../common/callApi";
+
+export const DEFAULT_BAG_QUERY_KEY = "getDefaultBag";
+export const DEFAULT_BAG_MUTATION_KEY = "mutateDefaultBag";
+export const DEFAULT_BAG_DELETE_KEY = "deleteDefaultBag";
```
ash3rz, line 7: I may have missed it, and this is minor, but I don't think these keys are used outside of this file so they may not need to be exported.  Also, I'm not sure the mutation or delete key are used.  IIRC, the mutation queries don't require a key.

src/components/Bag/index.js
```diff
@@ -0,0 +1,540 @@
...
+        minWidth: theme.spacing(20),
+    },
+    itemText: {
+        textOverflow: "ellipsis",
+        overflow: "hidden",
+    },
+}));
+
+const BagSkeleton = () => (
+    <Skeleton variant="rect" animation="wave" height={100} width="100%" />
```
ash3rz, line 98: A static ID on the loading skeletons is usually pretty helpful for testing

src/components/Bag/downloads/index.js
johnworth: A copy-to-clipboard button could be added. I tested it by highlighting the links and using the right-click menu to send them to a download manager extension I use.

src/components/Bag/classes.js
```diff
@@ -0,0 +1,120 @@
...
+    icon(t) {
+        return <img src="/analyses_selected.png" alt={t("common:analyses")} />;
+    }
+
+    get shareable() {
+        return true;
+    }
+}
+
+class AppBagItem extends BagItem {
```
ash3rz, line 85: I may have missed it, but at least for sharing, app items also need to contain the `system_id`

src/components/Bag/index.js
johnworth: Should I disable the button if there aren't any items to download? I know that we've been trying to avoid that, but it seems less annoying than having an error announcer pop up or just having nothing happen.

src/components/Bag/classes.js
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    get name() {
+        return this.item.path.substring(this.item.path.lastIndexOf("/") + 1);
+    }
+    get path() {
+        return this.item.path;
+    }
+}
+
```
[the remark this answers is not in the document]
johnworth: I may have to tackle that after the user test. Some of the paths get very long and I'd need to find a way to display them that isn't just a tooltip (large tooltips get annoying, imo, especially if they disappear automatically).

ash3rz (approved):
LGTM 🎉 I only had a few nits to pick.

As for the screenshots, my only question is whether the icons can still fit perhaps for mobile?  Otherwise, I'm a little worried if maybe two different resources with different resource types (for example, maybe an app and a folder) have the same name, if the user will be able to figure out which is which.  At the same time, the bag concept is so new, it may be fine to wait and get feedback to see what all comes out so changes can be done with everything in mind (if anything).

src/components/Bag/index.js
sri2k1us: Download button should not appear when there are no items that can be downloaded.

johnworth: This is fixed now.

src/components/Bag/index.js
```diff
@@ -0,0 +1,540 @@
...
+    useEffect(() => {
+        setBadgeCount(allItems.length);
+    }, [allItems, setBadgeCount]);
+
+    const handleSharingClick = (event) => {
+        event.preventDefault();
+        event.stopPropagation();
+
+        setBagDlgOpen(false);
+        setSharingOpen(true);
```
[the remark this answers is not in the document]
johnworth: I'll make it so the buttons to open the dialogs do not appear if there's nothing in the bag.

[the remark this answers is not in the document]
johnworth: Okay, this is done.

src/components/Bag/classes.js
johnworth: I removed it and replaced it with an empty string.

src/components/Bag/classes.js
johnworth: The nested item attribute in the AppBagItem contains the unaltered object from the app listing it originated from, and that includes the system_id. The item object gets pulled up as the main object before getting handed off to the sharing dialog.

src/components/Bag/index.js
johnworth: Oops, yup. I always forget that.

johnworth: Working on a fix in a separate PR.

src/serviceFacades/bags.js
johnworth: Thanks for pointing this out. Originally I was just using bare useQuery() and useMutation() calls and then changed to the current setup. Forgot to unexport stuff and remove uneeded keys.

src/components/analyses/toolbar/AnalysesDotMenu.js
```diff
@@ -306,6 +309,21 @@ function DotMenuItems(props) {
                 <ListItemText primary={t("filterLbl")} />
             </MenuItem>
         ),
+        hasSelection && (
```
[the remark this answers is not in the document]
johnworth: Working on a fix for this in another PR.

src/components/Bag/downloads/index.js
johnworth: This is going into a separate PR.

src/components/Bag/classes.js
```diff
@@ -0,0 +1,101 @@
...
+        return <img src="/apps_selected.png" alt={t("common:apps")} />;
+    }
+
+    get shareable() {
+        return true;
+    }
+}
+
+export const FOLDER_TYPE = "dir";
+export const FILE_TYPE = "file";
```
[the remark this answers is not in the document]
johnworth: Fixed in a new PR.

PR #164, opened 2020-08-19
fix 404 immediately after logging in

slr71:


src/server/index.js
```diff
@@ -75,7 +75,9 @@ app.prepare()
         server.use(keycloakClient.middleware());
 
         logger.info("adding the /login handler");
-        server.get("/login", keycloakClient.protect());
+        server.get("/login", keycloakClient.protect(), (req, res) => {
+            res.redirect("/");
+        });
```
slr71, line 80: Apparently we were never hitting the `/login` URL directly before. It must have always been `/login/dashboard` or something like that.

psarando:
> :shipit: immediately

Just a play on the PR title... not actually issuing orders 😜

slr71:
Thanks for the review!

slr71:
Er...reviews!

PR #169, opened 2020-08-29
CORE-776 Add cancel support to analyses listing.

psarando:
This PR will add multi-select cancel support to the analyses listing.

Separate error announcer messages are displayed if a cancel request returns an error and only 1 analysis is selected vs. many selected analyses.


src/serviceFacades/analyses.js
```diff
@@ -76,7 +76,28 @@ function getAnalysisRelaunchInfo(key, { id }) {
...
+        analysisIds &&
+        analysisIds.length > 0 &&
+        new Promise((resolve, reject) =>
+            Promise.all(analysisIds.map((id) => cancelAnalysis({ id })))
+                .then((values) => resolve(values))
+                .catch((error) => reject(error))
+        )
+    );
+}
+
```
sri2k1us, line 97: 👍

psarando:
Thanks for the reviews! 👍

PR #171, opened 2020-09-03
CORE-775 Add Complete and Save Outputs action to analyses listing

psarando:
This PR will add the "Complete and Save Outputs" action to the analyses listing.

It also includes a bug fix to prevent the "Cancel" action from displaying in the dot menu when no analyses are selected, and improved success and error announcer messages for stopped analyses.


ash3rz (approved):
LGTM :shipit:

psarando:
Thanks for the review 👍

PR #172, opened 2020-09-09
Use absolute imports, without aliasing (for data listing component)

ianmcorvidae:
this reduces the number of places in our bundles the data Listing component is included

as discussed in slack!

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -1,11 +1,5 @@
 {
-    "compilerOptions": {
-      "baseUrl": "./src",
-      "paths": {
-        "@/components/*": ["components/*"],
-        "@/contexts/*": ["contexts/*"],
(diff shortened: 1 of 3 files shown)


jsconfig.json
```diff
@@ -1,11 +1,5 @@
 {
-    "compilerOptions": {
-      "baseUrl": "./src",
-      "paths": {
-        "@/components/*": ["components/*"],
-        "@/contexts/*": ["contexts/*"],
-        "@/common/*": ["common/*"],
-        "@/serviceFacades/*":["@/serviceFacades/*"]
```
ianmcorvidae, line 8: We aren't using these anywhere and since they break storybook it seems like they're better removed.

sri2k1us:
LGTM :shipit:

PR #173, opened 2020-09-10
CORE-1138 - File Viewers 1/ n

sri2k1us:
- Text Viewer with syntax highlighting
image
- Image Viewer 
image
- Image Viewer in iphone 5 /SE 
image
- CSV / Structured Text Viewer
image
- CSV / Structured Text Viewer -iPAD
image
- PDF Viewer
image

Notes:
Only features ✅ are done
- Text Viewer with syntax highlighting ✅ 
-  PDF, Image viewer ✅ 
- CSV and other structured text viewer ✅ 
- Allow users to download file ✅ 
- Toolbar to view line numbers, details, download and set headers. ✅ 
- Send to Genome browser 
- HT input and Multi-Input file viewer with editing


src/components/data/viewers/FileViewer.js
```diff
@@ -0,0 +1,271 @@
...
+            return viewerConstants.COMMA_DELIMITER;
+        } else if (
+            infoTypes.TSV === infoType ||
+            infoTypes.VCF === infoType ||
+            infoTypes.GFF === infoType ||
+            infoTypes.BED === infoType ||
+            infoTypes.GTF === infoType ||
+            infoTypes.BOWTIE === infoType ||
+            infoTypes.HT_ANALYSIS_PATH_LIST === infoType ||
+            infoTypes.MULTI_INPUT_PATH_LIST === infoType
```
ianmcorvidae, line 104: It might be nice to change this to use [Array.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) if we can. Babel will probably swap it out for he sake of the few browsers that don't support it, but it makes this code a lot smaller/easier to read, I suspect.

src/server/api/data.js
```diff
@@ -149,5 +149,41 @@ export default function dataRouter() {
         })
     );
 
+    logger.info("adding the GET /api/filesystem/file/manifest handler");
+    api.get(
+        "/filesystem/file/manifest",
```
ianmcorvidae, line 154: This is somewhat secondary to this PR, but it would be worth checking the performance of the manifest endpoint, particularly whether we're returning anything we don't need/that might be a performance drain. `content-type` is often less than great, performance-wise, for example.

longer-term conversation, anyway, but I thought I'd mention it; no changes needed in this PR anyway

ianmcorvidae (approved):
some miscellaneous comments, nothing that seems blocking or anything like that though

johnworth (approved):
LGTM and the code runs locally and I can open files after dragging and dropping them into a folder. :shipit:

slr71 (approved):
I didn't catch any problems. :shipit:

sri2k1us:
Thanks for the reviews.

PR #177, opened 2020-09-16
CORE-1139 Only validate non-empty app launch fields when not required.

psarando:
Regex validators could display an error for blank fields, even when they were not marked as required.

This PR will update the app launch forms to only validate non-empty fields when not required.


sri2k1us (approved):
:shipit:

psarando:
Thanks for the review 👍

PR #178, opened 2020-09-16
VICE Admin: Searching or setting job limit for a non-existant user displays error

johnworth:
Previously, the job limit setting component would make it appear that the limit was set successfully if you put in the name of a non-existant user. Now an announcer will pop up displaying an error message telling you that the user does not exist.

Additionally, the user field is not cleared when you search for a user, only when the new limit is set successfully after hitting the Set button.

There are no pictures because there are no visible changes aside from the contents of the error message announcer.

The workspaces endpoint was added to Sonora to allow us to determine if the user exists in our system.


src/serviceFacades/vice/admin.js
```diff
@@ -51,23 +53,56 @@ export const downloadInputFiles = ({ analysisID }) =>
...
+
+    if (!username.endsWith(suffix)) {
+        username = `${username}${suffix}`;
+    }
+
+    return callApi({
+        endpoint: `/api/workspaces?username=${username}`,
+        method: "GET",
+    }).then((data) => data.workspaces.length > 0);
+};
```
sri2k1us, line 100: I think we can still use react-query for this logic: https://react-query.tanstack.com/docs/guides/queries#dependent-queries

sri2k1us (approved):
LGTM. :shipit:

ash3rz (approved):
LGTM :shipit:

PR #180, opened 2020-09-19
CORE-733 Add sharing dialog

ash3rz:
This ended up being a massive PR and the sharing logic ended up being a bit dense 😬 I'm heavily using `reduce` for just about everything, but it matches the old UI logic for the most part.

Known Bug: For an iPhone 5 size, there's supposed to be a divider between the user list and the resource list, but it disappears for some reason.

Still to-do: wire up into sonora.  For now, this is all storybook-based.

Desktop:

image

Mobile:
image

image


src/components/sharing/SubjectSearchField.js
```diff
@@ -0,0 +1,117 @@
...
+    const [searchTerm, setSearchTerm] = useState(null);
+    const [options, setOptions] = useState([]);
+    const theme = useTheme();
+    const isMobile = useMediaQuery(theme.breakpoints.down("sm"));
+
+    const { status: subjectSearchStatus } = useQuery({
+        queryKey: { searchTerm },
+        queryFn: searchSubjects,
+        config: {
+            enabled: !!searchTerm,
```
sri2k1us, line 31: I think this should be just `!searchTerm`. `!!` doesn't do logical `NOT`

src/components/sharing/util.js
```diff
@@ -0,0 +1,431 @@
...
+    return addDisplayPermissions(userMap, resourceTotal);
+};
+
+// Takes a user and examines its paths, apps, analyses, and tools keys to compare
+// against the resources list to find missing resources.  Adds the missing resources.
+export const addMissingResourcesToUser = (user, resources) => {
+    const currentResources = getUserResourceList(user);
+
+    return Object.keys(resources).reduce(
+        (acc, type) => {
```
sri2k1us, line 269: It is not very obvious what `acc` stands for ? `user account` ? if so can we rename this variable for clarity ?

sri2k1us (commented):
This is great progress on Sharing overall 👍 🎉 . I found only minor issues but I am holding it just short of approval mainly for 2 UX things. I want to see if these things can be fixed.

- Dialog scroll behavior
- Chips vs List

src/components/sharing/util.js
```diff
@@ -0,0 +1,431 @@
...
+
+// Takes a list of permission values (i.e. read, write, own) and returns
+// what to display as their overall permission
+const getDisplayPermission = (permissionList, resourceTotal) => {
+    if (permissionList.length !== resourceTotal) {
+        return SharingPermissions.VARIES;
+    }
+    const values = getUniqueList(permissionList);
+    return values.length === 1 ? values[0] : SharingPermissions.VARIES;
+};
```
slr71, line 28: I like the idea of having separate functions for these. 👍

src/components/sharing/util.js
slr71: `acc` is  a common variable name used in reduction functions that refers to the accumulated result.

src/components/sharing/util.js
```diff
@@ -0,0 +1,431 @@
...
+ *
+ * @param responses - permission-lister responses
+ * @param userInfos
+ * @param resourceTotal
+ * @returns {{}}
+ */
+export const getUserMap = (responses, userInfos, resourceTotal) => {
+    const userMap = responses.reduce((acc, resp) => {
+        const type = getResponseType(resp);
+        const { permListFn, userIdFn, fmtSharingFn } = getSharingFns(type);
```
slr71, line 225: I like the use of higher-order functions in this file. It makes the code a _lot_ more readable.

slr71 (approved):
I just had some general comments. I didn't catch any issues that @sri2k1us didn't already catch.

src/components/sharing/util.js
```diff
@@ -0,0 +1,431 @@
...
+        return resp[type].reduce(
+            (shares, sharing) => {
+                const permissions = permListFn(sharing);
+
+                return permissions.reduce(
+                    (userMap, permission) => {
+                        const userId = userIdFn(permission);
+                        const currentUser = {
+                            ...(userMap[userId] || userInfos[userId]),
+                        };
```
psarando, line 235: No need to copy here since `currentUser` will be copied below.
```suggestion
                        const currentUser = (userMap[userId] || userInfos[userId]);
```

src/components/sharing/util.js
```diff
@@ -0,0 +1,431 @@
...
+                        const updatedUser = {
+                            ...currentUser,
+                            [type]: [
+                                ...(currentUser[type] || []),
+                                resourceWithPermission,
+                            ],
+                        };
+                        return { ...userMap, [userId]: updatedUser };
+                    },
+                    { ...shares }
```
psarando, line 254: I don't think this copy is necessary either.

src/components/sharing/util.js
```diff
@@ -0,0 +1,431 @@
...
+                            ],
+                        };
+                    } else {
+                        return { ...resources };
+                    }
+                },
+                { ...acc }
+            );
+        },
+        { ...user }
```
psarando, line 294: This was a little tricky to follow, but I'm not sure what can be done about it: but it seems that the `user` is copied into the first reducer, which is then named `acc` and copied into the inner reducer, which now refers to it as `resources`.

In the end, since the inner reducer is always returning a copy of `user`/`resources`, then I don't think `user` and `acc` need to be copied initially.

src/components/sharing/util.js
```diff
@@ -0,0 +1,431 @@
...
+                    } else {
+                        return { ...acc };
+                    }
+                },
+                { ...acc }
+            );
+        } else {
+            return { ...acc };
+        }
+    }, {});
```
psarando, line 384: Since something is being returned inside these if-checks, I don't think the `else` blocks are necessary:
```javascript
if (check) {
  return ...;
}
return acc;
```
Again, I don't think the `{ ...acc }` copies are required, especially since it's being created by this function in the first place as an initial `{}` in the top-level reducer.

src/components/sharing/util.js
```diff
@@ -0,0 +1,431 @@
...
```
psarando, line 430: Same comments as `getSharingUpdates`.

psarando (approved):
Nice work! 🎉 

This was a little dense to go through, but I found it fairly easy to follow, especially with the tests and the comments/docs (but I'm also familiar with the current DE code).

I just had a few code-style suggestions, but this LGTM 👍

src/components/sharing/util.js
ianmcorvidae: Reading this myself since seeing the comment intrigued me :sweat_smile:, I do think that the copies aren't needed too, and I also think that having the innermost reduce using the name `resources` is a likely source of confusion, since that's also an argument to the outermost function. I think there's some other stuff that might be more nicely done without the `...` operator too to make the whole thing shorter, which would I think help with reading it, despite the added code density

Perhaps something like:

```javascript
const addMissingResourcesToUser = (user, resources) => {
  const currentResources = getUserResourceList(user);

  return Object.keys(resources).reduce(
    (newUser, type) => {
      const { idFn, fmtSharingFn } = getSharingFns(type)
      return resources[type].reduce(
        (newResources, resource) => (
          currentResources.some((perm) => idFn(perm) === idFn(resource)) ?
          { ...newResources } :
          { ...newResources, [type]: (newResources[type] || []).concat(fmtSharingFn(resource)) }
        ),
        newUser
      )
      }
    },
    user
  )
}
```

(untested, but hopefully the ideas in terms of non-reuse of names + compactness come through anyway)

psarando: Looks right to me 👍 
Although I'm fine with `...` spread/copy in place of `concat` as well:
```javascript
{ ...newResources, [type]: [...(newResources[type] || []), fmtSharingFn(resource)] }
```

src/serviceFacades/sharing.js
```diff
@@ -22,10 +27,192 @@ import callApi from "../common/callApi";
...
+    }
+
+    return new Promise((resolve, reject) => {
+        Promise.all(promises)
+            .then((values) => {
+                resolve(values);
+            })
+            .catch((error) => {
+                reject(error);
+            });
```
ianmcorvidae, line 154: I think these lines are unneeded, and the `return new Promise((resolve, reject) => { })` wrapper. next.js uses [bluebird](http://bluebirdjs.com) for the Promise implementation, not that it matters much since it's based on a standard, and what [`Promise.all`](http://bluebirdjs.com/docs/api/promise.all.html) already does is return a promise that... resolves with an array of values or rejects with an error. So this whole thing can just be `return Promise.all(promises)`

sri2k1us (approved):
Looks like we all are in agreement with the proposed UX changes. :shipit:

src/components/sharing/index.js
```diff
@@ -0,0 +1,386 @@
...
+        config: {
+            onSuccess: (results) => {
+                setPermissions(results);
+                setUserIdList(getUserSet(results));
+            },
+            onError: (error) => {
+                showErrorAnnouncer(tSharing("fetchPermissionsError"), error);
+            },
+        },
+    });
```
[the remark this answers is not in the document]
ash3rz: Yes, good catch - I have this fixed in the next PR when sharing gets added to sonora

src/components/sharing/index.js
```diff
@@ -0,0 +1,386 @@
...
+            : userExist
+            ? announce({
+                  text: tSharing("userAlreadyAdded"),
+                  variant: AnnouncerConstants.INFO,
+              })
+            : isGroup(user) && hasData
+            ? announce({
+                  text: tSharing("dataGroupSharingDisabled"),
+                  variant: AnnouncerConstants.INFO,
+              })
```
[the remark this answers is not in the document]
ash3rz: Yes, good point - I'll address this in the second UI PR for this

src/components/sharing/SharedItem.js
```diff
@@ -0,0 +1,75 @@
...
+                idFn: (resource) => resource.id,
+                labelFn: (resource) => resource.name,
+                icon: <Apps />,
+            };
+        }
+        case TYPE.ANALYSES: {
+            return {
+                idFn: (resource) => resource.id,
+                labelFn: (resource) => resource.name,
+                icon: <AnalysisIcon />,
```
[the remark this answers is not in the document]
ash3rz: Will do - will fix in the second UI PR

src/__tests__/sharing.js
```diff
@@ -0,0 +1,522 @@
...
+    };
+
+    const expectedUnsharing = {
+        paths: [
+            {
+                paths: ["/iplant/home/aramsey/CORE-9077-path.list"],
+                user: "alfred",
+            },
+        ],
+    };
```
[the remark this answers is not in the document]
ash3rz: Ah, this might not be super readable/recognizable, but on line 335 we're making a call to the `getUnsharingUpdates` function and we're sending an original `userMap` that contained alfred's permissions (I guess the confusing part is that it says we're updating the permission from `read` to `own`) and then a new `userMap` which is just an empty object `{}`. 

```const unsharingResult = getUnsharingUpdates(user, {});```

The idea is that originally alfred had some permissions, we don't really care which, and now, according to the second `userMap` which is empty, that user has been removed.  It was a quick and dirty way I could think of to test unsharing with the same user object I'm using for sharing.

File: src/__tests__/sharing.js [hunk 15]
```diff
@@ -0,0 +1,522 @@
...
+    const expected = [];
+
+    expect(result).toStrictEqual(expected);
+});
+
+/**
+ * Tests for a user with permissions on only 1 resource, a user with varied
+ * permissions, and users with uniform permissions on all resources
+ */
+it("tests userMap given all permission-lister responses", () => {
```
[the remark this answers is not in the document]
ash3rz: Good catch - I'll update this for a bit more clarity

src/components/sharing/index.js
```diff
@@ -0,0 +1,386 @@
...
+        const userId = user.id;
+
+        // Cannot update permissions for groups if data is included
+        if (isGroup(user) && hasData) {
+            announce({
+                text: tSharing("dataGroupSharingDisabled"),
+                variant: AnnouncerConstants.INFO,
+            });
+            return;
+        }
```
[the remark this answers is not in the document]
ash3rz: Oops!  Great catch - I'll include analyses in this check

src/__tests__/sharing.js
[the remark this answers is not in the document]
psarando: Ah, OK. I guess I didn't read this test carefully enough the first time.

I see now that it's really testing 2 things, with `getUnsharingUpdates` being tested 2nd by `expect(unsharingResult).toStrictEqual(expectedUnsharing)`.

src/components/sharing/util.js
ash3rz: 😄 I just love making copies I guess 🤷‍♀️

ash3rz:
I believe I've addressed all the non-UI comments at this point.  Thanks so much for the really helpful reviews and suggestions!  I should have 2 more PRs incoming - one for wiring this up into sonora and one for the UI changes.

PR #184, opened 2020-09-23
Cache page size in local storage

sri2k1us:
- SSIA
Note: Removed 25, 50 from size options.

### DIFF
--- a/public/static/locales/en/util.json
+++ b/public/static/locales/en/util.json
@@ -1,4 +1,5 @@
 {
     "details": "Details",
+    "selectPageSize": "Select page size",
     "viewDetails": "View Details"
 }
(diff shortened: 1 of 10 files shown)


public/static/locales/en/util.json
```diff
@@ -1,4 +1,5 @@
 {
     "details": "Details",
+    "selectPageSize": "Select page size",
```
ash3rz, line 3: With the addition of this, does this mean the data/apps/analyses pages also have to have `util` added to their `namespacesRequired` array? (Which reminds me, I probably need to do the same for sharing)

ash3rz (approved):
LGTM, just had a question about i18n stuff :shipit:

public/static/locales/en/util.json
sri2k1us: I don't see specific warning about including `util` when I checked for analyses now but I do see a warning about `app.js` not having `namespacesRequired` array declared. So Let me add that.

sri2k1us:
Thanks for the review.

PR #186, opened 2020-09-28
CORE-733 Add sharing to Sonora

ash3rz:
NOTE: All the UI changes discussed in #180 will be in the next PR. This is just for integrating the dialog into Sonora.

I've also updated the dot menus to be more consistent and only show additional options instead of repeating options that are already visible in the toolbar.

Data on desktop:
image

Data on mobile with dot menu expanded:
image

Data's new row dot menu:
image

Apps on Desktop:
image

Apps on mobile with dot menu expanded:
image

Analyses on Desktop:
image

Analyses on Desktop with dot menu expanded:
image

Analyses on mobile with dot menu expanded:
image

### DIFF
--- a/src/components/analyses/toolbar/Toolbar.js
+++ b/src/components/analyses/toolbar/Toolbar.js
@@ -32,9 +32,9 @@ import {
 import Autocomplete from "@material-ui/lab/Autocomplete";
 
 import { Info, FilterList as FilterListIcon } from "@material-ui/icons";
-import SharingButton from "../../sharing/SharingButton";
-import Sharing from "../../sharing";
-import { formatSharedAnalyses } from "../../sharing/util";
+import SharingButton from "components/sharing/SharingButton";
+import Sharing from "components/sharing";
+import { formatSharedAnalyses } from "components/sharing/util";
 
 const useStyles = makeStyles((theme) => ({
     menuButton: {
--- a/src/components/analyses/toolbar/Toolbar.js
+++ b/src/components/analyses/toolbar/Toolbar.js
@@ -32,6 +32,9 @@ import {
 import Autocomplete from "@material-ui/lab/Autocomplete";
 
 import { Info, FilterList as FilterListIcon } from "@material-ui/icons";
+import SharingButton from "../../sharing/SharingButton";
+import Sharing from "../../sharing";
+import { formatSharedAnalyses } from "../../sharing/util";
 
 const useStyles = makeStyles((theme) => ({
     menuButton: {
@@ -174,11 +177,16 @@ function AnalysesToolbar(props) {
         handleRename,
         handleSaveAndComplete,
         handleBatchIconClick,
+        canShare,
     } = props;
     const classes = useStyles();
     const { t } = useTranslation("analyses");
     const analysesNavId = build(baseId, ids.ANALYSES_NAVIGATION);
     const [openFilterDialog, setOpenFilterDialog] = useState(false);
+    const [sharingDlgOpen, setSharingDlgOpen] = useState(false);
+
+    const sharingAnalyses = formatSharedAnalyses(getSelectedAnalyses());
+
     return (
         <>
             <Toolbar variant="dense" id={analysesNavId}>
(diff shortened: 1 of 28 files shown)


sri2k1us (approved):
LGTM 👍 . Nice refactors and components re-use ✨ 🎉 . I do have few concerns which I discovered when I ran the PR locally. 

1. Selection Drawer:
I think I brought this up in #ui but here it is again. The buttons overflow in the selection drawer and I think we should hide them.
image

2. Dialog sizing:
The dialog size varies a lot even within one share / unshare action. For example, when am sharing 1 item, the sharing dialog was at a reasonable size but after I clicked the `Done` button, the loading skeleton almost fills the entire screen in desktop. Is there a way make the dialog have fixed width ?

3. Hidden dot menu on mdUp in Data:
I see the dot menu is hidden for large screens. Does it mean we will displaying all operations (metadata, Url Sharing etc...) as button(s) in toolbar or is this a temporary thing ?

src/components/apps/toolbar/Toolbar.js
```diff
@@ -99,13 +102,18 @@ function AppsToolbar(props) {
         toggleDisplay,
         detailsEnabled,
         onDetailsSelected,
+        canShare,
+        selectedApps,
         intl,
```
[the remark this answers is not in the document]
ash3rz: Good catch - I'll remove it

ash3rz:
1. I'll take a look at selection drawer - I didn't even think to check
2. Yes, I noticed the sizing was off as well 😄 I think I copied the code from the data listing where 25 rows made a lot more sense.  I'll scale it down to 5 probably.
3. Hiding dot menu is temporary - I just wanted to hide the dot menu if there was nothing in there, so it wouldn't be confusing as to why it's there. As we add more things, we can add it back.

src/components/sharing/util.js
```diff
@@ -407,3 +409,41 @@ export const getUnsharingUpdates = (originalUsers, userMap) => {
...
+
+export const formatSharedData = (selectedData) => {
+    const formattedData =
+        selectedData?.map((data) => {
+            return {
+                label: data.label,
+                path: data.path,
+                type: data.type,
+            };
+        }) || [];
```
psarando, line 421: Destructuring could make this a little more condensed:
```suggestion
        selectedData?.map(({ label, path, type }) => ({ label, path, type })) || [];
```

src/components/sharing/util.js
```diff
@@ -407,3 +409,41 @@ export const getUnsharingUpdates = (originalUsers, userMap) => {
...
+
+export const formatSharedApps = (selectedApps) => {
+    const formattedApps =
+        selectedApps?.map((app) => {
+            return {
+                name: app.name,
+                id: app.id,
+                system_id: app.system_id,
+            };
+        }) || [];
```
psarando, line 434: Destructuring could make this a little more condensed too.

src/components/sharing/util.js
```diff
@@ -407,3 +409,41 @@ export const getUnsharingUpdates = (originalUsers, userMap) => {
...
+};
+
+export const formatSharedAnalyses = (selectedAnalyses) => {
+    const formattedAnalyses =
+        selectedAnalyses?.map((analysis) => {
+            return {
+                name: analysis.name,
+                id: analysis.id,
+            };
+        }) || [];
```
psarando, line 446:
```suggestion
        selectedAnalyses?.map(({ name, id }) => ({ name, id })) || [];
```

psarando (approved):
LGTM 👍
I also agree: nice refactors in this PR ✨ 

I just had a few style suggestions, and I also spotted a GitHub Actions warning about unused imports.

src/components/analyses/toolbar/Toolbar.js
```diff
@@ -174,11 +177,19 @@ function AnalysesToolbar(props) {
...
     const classes = useStyles();
     const { t } = useTranslation("analyses");
     const analysesNavId = build(baseId, ids.ANALYSES_NAVIGATION);
     const [openFilterDialog, setOpenFilterDialog] = useState(false);
+    const [sharingDlgOpen, setSharingDlgOpen] = useState(false);
+
+    const hasSelection = getSelectedAnalyses
+        ? getSelectedAnalyses().length > 0
+        : false;
+    const sharingAnalyses = formatSharedAnalyses(getSelectedAnalyses());
```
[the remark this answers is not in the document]
ash3rz: The `formatSharedAnalyses` function checks for null as well, so I think it's safe to omit it here

ash3rz:
Thanks for the reviews!

PR #187, opened 2020-09-29
CORE-1136 Add ShareWithSupportDialog to analyses listing

psarando:
This PR will add the `ShareWithSupportDialog` to the Analyses listing.

It also adds an `ExternalLink` util component, which is just a wrapper around the `target="_blank" rel="noopener noreferrer"` boilerplate for links that open external sites in a new tab.

Submitted Status Help Dialog for a DE app:
Submitted Status Help Dialog - DE

Submitted Status Help Dialog for an HPC app:
Submitted Status Help Dialog - HPC

Running Status Help Dialog for a DE app:
Running Status Help Dialog - DE

Running Status Help Dialog for an HPC app:
Running Status Help Dialog - HPC

Completed Status Help Dialog for no outputs:
Completed Status Help Dialog - No Output

Completed Status Help Dialog for unexpected outputs:
Completed Status Help Dialog - Unexpected Output

Failed Status Help Dialog:
Failed Status Help Dialog

---

Once the user taps "I still need help" the dialog displays a summary of their analysis that will be emailed to support. The user must check the share agreement disclaimer in order to enable the `Submit` button.
Analysis Share Summary and Disclaimer Dialog

When the user submits the dialog, the analysis is shared with the support user and an email is sent via the `support-email` endpoint containing a JSON summary of the analysis, with some user info included.


src/components/analyses/ShareWithSupportDialog.js
```diff
@@ -0,0 +1,577 @@
...
+                </Grid>
+                <Grid item xs={12}>
+                    {t("user")} : {analysis.username}
+                </Grid>
+                <Grid item xs={12}>
+                    {t("name")} : {name}
+                </Grid>
+                <Grid item xs={12}>
+                    {t("email")} : {email}
+                </Grid>
```
sri2k1us, line 80: We can also use `utils/GridLabelValue.js` (for all the `label: value` type display) here for consistency.

src/components/analyses/ShareWithSupportDialog.js
```diff
@@ -0,0 +1,577 @@
...
+                                <Link
+                                    id={build(
+                                        baseId,
+                                        UtilIds.CONTACT_SUPPORT_BUTTON
+                                    )}
+                                    component="button"
+                                    onClick={(event) => {
+                                        // prevent form submission
+                                        event.preventDefault();
+                                        intercomShow();
```
sri2k1us, line 334: 👍

sri2k1us (approved):
LGTM 👍. I had minor nit-picks and few comments.

src/components/analyses/ShareWithSupportDialog.js
```diff
@@ -0,0 +1,577 @@
...
+                    <React.Fragment>
+                        <AnalysisInfo
+                            analysis={analysis}
+                            name={name}
+                            email={email}
+                        />
+                        <TextField
+                            id={build(baseDebugID, ids.COMMENTS)}
+                            placeholder={t("comments")}
+                            style={{ margin: 8 }}
```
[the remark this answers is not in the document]
psarando: Good call. This was [copied from the `ui` code](https://github.com/cyverse-de/ui/blob/b7226c32ece70df7077beeae2e247491fb297125/react-components/src/analysis/view/dialogs/ShareWithSupportDialog.js#L233-L236), but I can update it here for this PR.

src/components/analyses/ShareWithSupportDialog.js
```diff
@@ -0,0 +1,577 @@
...
+    const { t } = useTranslation("analyses");
+
+    return (
+        <FormControl component="fieldset">
+            <FormLabel
+                component="legend"
+                style={{ padding: 5, fontWeight: "bold" }}
+            >
+                {t("statusHelpOutputConditionHeader")}
+            </FormLabel>
```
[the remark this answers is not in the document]
psarando: Also [copied from the `ui` code](https://github.com/cyverse-de/ui/blob/b7226c32ece70df7077beeae2e247491fb297125/react-components/src/analysis/view/dialogs/ShareWithSupportDialog.js#L81-L87), but I'll check it out 👍

src/components/analyses/ShareWithSupportDialog.js
```diff
@@ -0,0 +1,577 @@
...
+                        i18nKey={
+                            analysis.system_id === "de"
+                                ? "statusHelpCondorSubmitted"
+                                : "statusHelpAgaveSubmitted"
+                        }
+                        components={{
+                            xsede: (
+                                <ExternalLink
+                                    href={constants.HELP_DOCS.HPC_APPS}
+                                />
```
ash3rz, line 147: 👍

src/components/analyses/ShareWithSupportDialog.js
```diff
@@ -0,0 +1,577 @@
...
+                ) : (
+                    <Button
+                        variant="contained"
+                        color="primary"
+                        onClick={() => {
+                            setShareWithSupport(true);
+                        }}
+                    >
+                        {t("needHelp")}
+                    </Button>
```
ash3rz, line 563: Since this is mostly static content in a dialog, I'm not sure how much we really want to test this, but in the very least it would be helpful to have static IDs on these dialog action buttons if we want to make sure the dialog at least can open, shows something, and then close it.

ash3rz (approved):
LGTM :shipit:

src/components/analyses/ShareWithSupportDialog.js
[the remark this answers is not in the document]
psarando: I think it actually looks better without the margins, so I'll just remove this style:
Comments - no margin

src/components/analyses/ShareWithSupportDialog.js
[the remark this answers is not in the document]
psarando: I also think this "Select Output condition" header looks better without the styling.

I don't think it needs the padding at all, but I think it could also do without the `bold`:
Select Output condition style

src/components/analyses/ShareWithSupportDialog.js
psarando: Looks much better with `GridLabelValue`:
GridLabelValue

src/components/analyses/ShareWithSupportDialog.js
```diff
@@ -50,36 +51,32 @@ function AnalysisInfo(props) {
     if (analysis) {
         return (
             <Grid container spacing={3} style={{ marginTop: 2, fontSize: 12 }}>
```
sri2k1us, line 53: Sorry missed this one. `marginTop` should use `theme.spacing()`. `fontSize` should not be set.

psarando: I think it looks good without the `margin` as well, so I'll just remove the custom styling here too:
Analysis Summary - no margin top

ash3rz: Just a quick note - `GridLabelValue` doesn't assume that the value, i.e. children passed in, will always be text like the `label` does, so you might need to wrap the values in a `Typography`.  Looks like the font is different between the two right now.

psarando: Good catch. I never seem to notice the font difference 😖 
(probably because I generally prefer serif fonts)

psarando:
Looks like I addressed all comments now, so thanks for the reviews! 👍

PR #190, opened 2020-10-05
CORE-733 Sharing cosmetic changes

ash3rz:
* Switches to using our custom app and analyses icons
* Switches from Chip to List for resource list
* Switches to using a sortable table for the user list
* Error messages are inlined into the dialog now
* Reduces some padding to maximize space

To finish another time:
Setting up independent scrolling for the two columns (user list, resources list). I've tried getting this to work, but keep running into buggy-ness.

Desktop with sample inlined error:
image

Mobile with scrolling:
image

image


src/components/sharing/SharedItem.js
```diff
@@ -1,25 +1,27 @@
...
 import {
-    Apps,
-    Assessment as AnalysisIcon,
     Build as ToolIcon,
     Folder as FolderIcon,
     InsertDriveFileOutlined as FileIcon,
 } from "@material-ui/icons";
+import { makeStyles } from "@material-ui/styles";
 import ResourceTypes from "../models/ResourceTypes";
+import { useTranslation } from "../../i18n";
```
sri2k1us, line 12: Should this be just `import { useTranslation } from "i18n";` ?

src/components/sharing/UserTable.js
```diff
@@ -0,0 +1,140 @@
...
+    ListItem,
+    ListItemText,
+    Table,
+    TableBody,
+    TableCell,
+    TableHead,
+    TableRow,
+    TableSortLabel,
+} from "@material-ui/core";
+import { Group } from "@material-ui/icons";
```
sri2k1us, line 20: Unused import. Not sure why github actions are not catching these anymore ? 🤔

src/components/sharing/UserTable.js
```diff
@@ -0,0 +1,140 @@
...
+                            onPermissionChange={(updatedPermission) =>
+                                onPermissionChange(user, updatedPermission)
+                            }
+                            onRemoveSelected={() => onRemoveUser(user)}
+                        />
+                    );
+                },
+            },
+        ],
+        [baseId, t, onPermissionChange, onRemoveUser]
```
sri2k1us, line 83: Another lint issue.

src/components/sharing/index.js
```diff
@@ -45,31 +43,33 @@ import isQueryLoading from "../utils/isQueryLoading";
...
 import {
     doSharingUpdates,
     GET_PERMISSIONS_QUERY_KEY,
     getPermissions,
 } from "../../serviceFacades/sharing";
-import withErrorAnnouncer from "components/utils/error/withErrorAnnouncer";
 import styles from "./styles";
+import UserTable from "./UserTable";
+import ErrorTypography from "../utils/error/ErrorTypography";
+import DEErrorDialog from "../utils/error/DEErrorDialog";
```
sri2k1us, line 58: This can be updated to `components/utils/error/...` for consistency. Think we agreed to move towards absolute imports for some performance improvements 🤔.

sri2k1us (approved):
LGTM. I had some minor comments. Another minor thing from the screen shot above: It seems like there is a lot of white space on the right side of the dialog. Can it be reduced ? That way `Cancel` and `Done` can be brought directly under the shared item list.

ash3rz:
I pushed some changes to fix the permissions panel bug that Ian found where doi-curators and data-curators weren't displaying properly at all - it's an issue where these entities don't return anything from user-info. I found this bug in the sharing dialog, but forgot to update it here. I also updated the text that is displayed to match the sharing dialog - email and description/institution only, no names.

image

ash3rz:
Thanks for the review!

PR #191, opened 2020-10-07
Fix type query parameter in data page.

sri2k1us:


src/pages/data/ds/[...pathItems].js
```diff
@@ -53,8 +53,8 @@ export default function DataStore() {
                 );
             } else {
                 router.push(
-                    `${baseRoutingPath}${dynamicPathName}?file=true&resourceId=${id}`,
-                    `${baseRoutingPath}${encodedPath}?file=true&resourceId=${id}`
+                    `${baseRoutingPath}${dynamicPathName}?type=${resourceType}&resourceId=${id}`,
+                    `${baseRoutingPath}${encodedPath}?type=${resourceType}&resourceId=${id}`
```
Comment (ash3rz) [inline src/pages/data/ds/[...pathItems].js:57]: Minor, but I don't think the `resourceType` in this `else` will ever be anything other than the file type?  In other places in this file we have `type=${ResourceTypes.FILE}` which I think would be fine here.

ash3rz (approved):
Minor suggestion, but LGTM 👍

src/pages/data/ds/[...pathItems].js
Comment (psarando) [reply src/pages/data/ds/[...pathItems].js:57]: The `if` check suggests it could even be `null` or `undefined`, so maybe another reason to explicitly set `type=${ResourceTypes.FILE}` here, I think.

Comment (sri2k1us) [reply src/pages/data/ds/[...pathItems].js]: Updated to `ResourceTypes.FILE`. Thanks for the reviews.

PR #193, opened 2020-10-07
Fix when data dot menu shows

ash3rz:
On mobile, it's always available.  On desktop, it only displays when something is selected so the Delete option is available.

### DIFF
--- a/src/components/data/toolbar/Toolbar.js
+++ b/src/components/data/toolbar/Toolbar.js
@@ -24,7 +24,14 @@ import DisplayTypeSelector from "../../utils/DisplayTypeSelector";
 
 import { build } from "@cyverse-de/ui-lib";
 
-import { Button, Hidden, makeStyles, Toolbar } from "@material-ui/core";
+import {
+    Button,
+    Hidden,
+    makeStyles,
+    Toolbar,
+    useMediaQuery,
+    useTheme,
+} from "@material-ui/core";
 
 import { CreateNewFolder, Info } from "@material-ui/icons";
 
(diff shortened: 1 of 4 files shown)


src/components/data/toolbar/Toolbar.js
```diff
@@ -61,6 +68,10 @@ function DataToolbar(props) {
     const onCreateFolderClicked = () => setCreateFolderDlgOpen(true);
     const selectedResources = getSelectedResources();
     const canShare = isOwner(selectedResources);
+    const theme = useTheme();
+    const isMobile = useMediaQuery(theme.breakpoints.down("sm"));
```
sri2k1us, line 72: I always use `xs` down for checking mobile screen sizes. I tested this PR locally and it seems to render stuff correctly but I would like to standardize on breakpoints to provide a uniform UX. Does using `xs` cause issues in iPads like screen sizes ?

sri2k1us (approved):
LGTM. I had one comment.

src/components/data/toolbar/Toolbar.js
ash3rz: I could rename this to `isSmall` if that's clearer - I agree that `xs` is the real mobile size so the variable name is probably confusing.  

This is set to `sm` because the buttons in the toolbar were already set to hide with `<Hidden smDown...`, so the `isMobile` is used to match that for the dot menu. If I switch to `xs` then things get cramped before they hide into the dot menu:
image

sri2k1us: Renaming the variable sounds good to me 👍. One last thing - Setting the toolbar button size to `small` might also help with real estate.

ash3rz: Looks like `small` helps a little bit, but still is cramped within the `sm` range:
image

sri2k1us: Thanks for giving it shot. So let's settle for `sm` break point. But can we still keep the Button size set `small`. I think it definitely will help breadcrumbs (large paths) getting little bit more space.

PR #194, opened 2020-10-10
NotificationMenu

flynn1999:
Still needs I18N so that is not yet implemented. Warnings need to be removed as well. Tidying and  documentation left.

Styling and functionality done.


sri2k1us (changes requested):
So far so good. I had few comments but the main thing right now is when I tried get this working in my machine, the notification menu won't pop up.

ash3rz (commented):
This is a good start 🎉 Just needs a few more small changes in addition to what Sriram already commented.

src/components/notifications/NotificationsMenu.js
```diff
@@ -0,0 +1,147 @@
...
+                    Notifications
+                </Typography>
+                <Divider/>
+
+                {isFetching && (
+                    <NotificationLoading/>
+                )}
+                {!isFetching && (
+                    <>
+                        {notifications.length > 0 && notifications.map((n) => (
```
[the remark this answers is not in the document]
flynn1999: Yeah I had fixed this and all the styling changes yesterday but could not commit due to the `keepMounted` issue. 😅

src/components/notifications/NotificationStyles.js
```diff
@@ -0,0 +1,43 @@
...
+            marginTop: theme.spacing(1),
+            marginLeft: theme.spacing(2),
+        },
+
+        footer: {
+            color: theme.palette.primary.main,
+            textAlign: "left",
+            float: "left",
+            display: "block",
+            width: "100%",
```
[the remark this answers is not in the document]
flynn1999: I tried getting rid of all the stylings but I could not get the text to align to the left. I could not find anything for aligning under Button api and I also tried using a typography child component and its display and textAlign as well but it was still aligned to the centre. I've removed the `width: 100%` but the textAlign is still present.

flynn1999:
@sri2k1us It was not opening under the app bar because the mockAxios was missing. It should run fine now on dev, notificationsPreviewTest and App Bar

sri2k1us:
> @sri2k1us It was not opening under the app bar because the mockAxios was missing. It should run fine now on dev, notificationsPreviewTest and App Bar

Ok..I will check it out but Just to be clear I wasn't talking about `storybook` (my understanding is `mockAxios` should only be used with `storybook`) but I tried running the whole Sonora on my machine.

src/components/layout/CyVerseAppBar.js
```diff
@@ -230,6 +232,10 @@ function CyverseAppBar(props) {
     const [bootstrapQueryEnabled, setBootstrapQueryEnabled] = useState(false);
     const [profileRefetchInterval, setProfileRefetchInterval] = useState(null);
     const setPreferences = usePreferences()[1];
+    mockAxios
```
ash3rz, line 235: You added mockAxios directly to the CyverseAppBar component - this is incorrect.  mockAxios stuff is for `storybook` only.  mockAxios fakes api calls so `storybook` can simulate real api calls. Adding this fake stuff to a real component (*.js file instead of *.story file) breaks the real component outside of `storybook`.

sri2k1us: Yup. Am trying to run this in local machine and everything is failing.

src/components/notifications/messages.js
```diff
@@ -0,0 +1,14 @@
...
+        startDate: "Start Date",
+        notifications: "Notifications",
+        endDate: "End Date",
+        applyFilter: "Apply Filter",
+        userName: "User Name",
+        errorMsg: 'There was an error fetching the latest notifications. Please try again!',
+    },
+};
+
+export default intlData;
```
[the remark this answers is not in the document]
ash3rz: This file still needs to be removed

ash3rz (commented):
This is looking good 🎉 and the checks are passing now.  However, I would encourage you to try this out in dev again and verify all the functionality (by dev, I mean `npm run dev`, not storybook).

* Log in
* Open the notifications drop down and see how the last 10 notifications look
* Do something in sonora that generates a new notification (rename a file, delete a file, run an analysis, etc.)
* Wait for the announcer to pop up with the new notification
* View the drop down again
* The dropdown should contain the new notification but still only have 10 max

src/components/notifications/NotificationsMenu.js
```diff
@@ -0,0 +1,139 @@
...
+                variant="h6"
+            >
+                {t("notifications")}
+            </Typography>
+            <Divider />
+
+            {isFetching && <Skeleton variant="rect" width={350} height={400} />}
+            {!isFetching &&
+                notifications.length > 0 &&
+                notifications.slice(0, 10).map((n, index) => (
```
ash3rz, line 88: Ah, I missed this part in my review.  With what you have now, if the user gets 1,000 notifications (just as an extreme example), your component would be holding 1,000 notifications in state and growing with every new notification but only displaying 10.  

Instead of this, let's make this component more efficient and update state to only hold 10 notifications.

src/components/layout/Notifications.js
```diff
@@ -147,18 +149,40 @@ function Notifications(props) {
...
+                    onClick={handleClick}
+                >
+                    <Badge badgeContent={unSeenCount} color="error">
+                        <NotificationsIcon />
+                    </Badge>
+                </IconButton>
+            </Tooltip>
+            <>
+                <NotificationsMenu
+                    unSeenCount={unSeenCount}
```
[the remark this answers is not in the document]
sri2k1us: Also related. I don't see `unSeenCount` being set initially when user logs in and last 10 notifications are fetched.

src/components/notifications/NotificationsMenu.js
```diff
@@ -0,0 +1,139 @@
...
+
+const useStyles = makeStyles(NotificationStyles);
+
+function getDisplayMessage(notification) {
+    if (notification) {
+        return notification.type === "data"
+            ? notification.subject
+            : notification.message.text;
+    }
+}
```
sri2k1us, line 31: This seems like a dupe of the function in `Notifications.js`

src/components/notifications/NotificationsMenu.js
```diff
@@ -0,0 +1,139 @@
...
+
+    const handleClose = () => {
+        props.setAnchorEl();
+    };
+
+    useEffect(() => {
+        if (notificationMssg != null) {
+            setNotifications([notificationMssg, ...notifications].reverse());
+        }
+    }, [notifications, notificationMssg]);
```
sri2k1us, line 55: This will cause infinite re-render (I double checked this by running this PR locally). We should only update the state with the notification message if it is not already in the array.

```suggestion
          const found = notifications?.find(
            (msg) => msg.message.id === notificationMssg?.message.id
        );
        if ((found === null  || found === undefined) && notificationMssg) {
            setNotifications([notificationMssg, ...notifications].slice(0,10));
        }
    }, [notifications, notificationMssg]);
```

sri2k1us: Also note, my suggestion includes @ash3rz suggestion to limit the state to hold only last 10 notifications at any point-in-time.

src/components/notifications/NotificationsMenu.js
```diff
@@ -0,0 +1,139 @@
...
+
+    const { isFetching } = useQuery({
+        queryKey: NOTIFICATIONS_KEY,
+        queryFn: getNotifications,
+        config: {
+            onSuccess: (results) => setNotifications(results?.messages),
+        },
```
sri2k1us, line 62: I noticed in line `53` you were reversing the array to show the latest first. I think that happen here.

src/components/notifications/NotificationsMenu.js
sri2k1us: The `slice` here can removed once you include my suggested commit.

src/components/notifications/NotificationsMenu.js
sri2k1us: Also you should update the `unSeenCount` from here so that the `badge` will be shown to user as soon as they login.

sri2k1us (changes requested):
Looks lot closer to getting this merged but we should fix the infinite re-render issue that happens after a new notification appears.

psarando (commented):
This is coming along nicely ✨ 

I didn't have much to add, and I only found a few minor nits to pick beyond @ash3rz's and @sri2k1us' comments.

Also, I can wait and rebase my PR #199 on top of this one after it's merged, so you don't have to worry about renaming the service facade key and get-function in this PR, since I can rename them in my PR once it's rebased.

src/components/notifications/styles.js
```diff
@@ -0,0 +1,20 @@
...
+        marginBottom: theme.spacing(1),
+        marginTop: theme.spacing(1),
+        marginLeft: theme.spacing(2),
+    },
+
+    footer: {
+        color: theme.palette.primary.main,
+        textAlign: "left",
+        display: "block",
+    },
```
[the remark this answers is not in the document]
flynn1999: Thanks for the review and suggestions!

src/components/notifications/NotificationsMenu.js
```diff
@@ -0,0 +1,156 @@
...
+                            )}
+                            variant="caption"
+                        >
+                            {getTimeStamp(n.message?.timestamp)}
+                        </Typography>
+                    </MenuItem>
+                ))}
+
+            <Divider light />
+            <>
```
ash3rz, line 126: I see in the browser console that this fragment is causing a bunch of warnings about how the Menu cannot accept a fragment as a child.  I don't think you need to wrap these buttons in anything. I think you can remove this fragment and it should still render fine.

src/serviceFacades/notifications.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * @author flynn
+ *
+ */
+import callApi from "../common/callApi";
+
+const NOTIFICATIONS_LAST_TEN_KEY = "fetchNotificatons";
```
ash3rz, line 7: This variable has been renamed, which is good, but I think Paul also intended for this string to be more descriptively named.  The string serves as a key for `react-query` to recognize different queries being run and `fetchNotifications` could accidentally be used as a key for getting all notifications instead of just the last 10, for example

ash3rz (approved):
I found 2 last nits to pick, but I tried this out locally and things seem to be working 🎉 
I'm approving because I believe all the comments have been addressed, but others can feel free to chime in again if I missed something either in the code or running it locally.

Nicely done!  First sonora component!

ash3rz:
Discussed with the team and we're going to merge this now so Paul can work off of what you have so far.  Great job!

PR #196, opened 2020-10-12
Analyses Pagination Bar bug fix

sri2k1us:
- Fix bug that was preventing pagination bar from being displayed after an incoming notification update.

### DIFF
--- a/src/components/analyses/listing/Listing.js
+++ b/src/components/analyses/listing/Listing.js
@@ -396,7 +396,7 @@ function Listing(props) {
                         if (analysisStatus !== found.status) {
                             found.status = analysisStatus;
                             found.enddate = message.payload.enddate;
-                            setData({ analyses: [...data.analyses] });
+                            setData({ analyses: [...data.analyses], ...data });
                         }
                     } else {
                         //add a new analysis record and remove the last record from the page
@@ -408,13 +408,15 @@ function Listing(props) {
                             );
                             setData({
                                 analyses: [message.payload, ...newPage],
+                                ...data,
                             });
                         } else if (data.analyses?.length === 0) {
                             //if page is empty...
-                            setData({ analyses: [message.payload] });
+                            setData({ analyses: [message.payload], total: 1 });
                         } else {
                             setData({
                                 analyses: [message.payload, ...data.analyses],
+                                ...data,
                             });
                         }
                     }
--- a/src/components/analyses/listing/Listing.js
+++ b/src/components/analyses/listing/Listing.js
@@ -394,9 +394,16 @@ function Listing(props) {
 
                     if (found) {
                         if (analysisStatus !== found.status) {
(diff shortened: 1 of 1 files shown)


src/components/analyses/listing/Listing.js
```diff
@@ -396,7 +396,7 @@ function Listing(props) {
                         if (analysisStatus !== found.status) {
                             found.status = analysisStatus;
                             found.enddate = message.payload.enddate;
-                            setData({ analyses: [...data.analyses] });
+                            setData({ analyses: [...data.analyses], ...data });
```
psarando, line 399: I think the copy of `...data` needs to come before the new value of `analyses`, otherwise the old values merged from `data` will clobber any new values.

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals

I think this is technically working now because the `found` value from state is getting mutated directly, which we shouldn't be doing.

Instead, `found` should be copied with the new values set in the copy:
```javascript
const newAnalyses = data.analyses.map((analysis) => analysis.id === message.payload.id
    ? { ...analysis, status: analysisStatus,  enddate: message.payload.enddate }
    : analysis);

setData({ ...data, analyses: newAnalyses });
```

src/components/analyses/listing/Listing.js
```diff
@@ -408,13 +408,15 @@ function Listing(props) {
                             );
                             setData({
                                 analyses: [message.payload, ...newPage],
+                                ...data,
                             });
```
psarando, line 412:
```suggestion
                            setData({
                                ...data,
                                analyses: [message.payload, ...newPage],
                            });
```

src/components/analyses/listing/Listing.js
```diff
@@ -408,13 +408,15 @@ function Listing(props) {
...
                         } else if (data.analyses?.length === 0) {
                             //if page is empty...
-                            setData({ analyses: [message.payload] });
+                            setData({ analyses: [message.payload], total: 1 });
                         } else {
                             setData({
                                 analyses: [message.payload, ...data.analyses],
+                                ...data,
                             });
```
psarando, line 420:
```suggestion
                            setData({
                                ...data,
                                analyses: [message.payload, ...data.analyses],
                            });
```

psarando (changes requested):

src/components/analyses/listing/Listing.js
sri2k1us: Makes sense. Will update it.

psarando (approved):
LGTM :shipit: 👍

psarando:
Looks like Chromatic can't build the Data Header story, starting with this PR for some reason 😕

sri2k1us:
Thanks for the review.

PR #199, opened 2020-10-14
CORE-1140 Add Notifications listing with table view.

psarando:
This PR will import the Notifications listing table view from the `cyverse-de/ui` repo, with updates for displaying and using the API service facades of this repo.

### Changes from the original made for this PR

* The `Mark as Seen` button will hide if any messages are selected that have already been read, or if none are selected.
* The `Delete` button will hide when no messages are selected.
* The entire row is now highlighted for unread messages.
* The table header, toolbar, filter field, and button styles have been standardized with the rest of Sonora.

### TODO

* A next.js page integrating this view will be added in the next PR.
* All message selection handling still needs to be integrated.
  * Including things like navigating to the data view, and opening dialogs for team actions and request status or history.
  * This will probably need to be added in a common util namespace that can be reused by the Notifications Menu.

---

Notifications Listing on Desktop:
Notifications Listing - Desktop view

---

Mobile view, but the entire page (no scrolling):
Notifications Listing - Mobile full view

---

Mobile view on iPhone SE (requires scrolling):
Notifications Listing - Mobile actual view


src/components/models/notificationCategory.js
```diff
@@ -0,0 +1,12 @@
...
+    new: "New",
+    analysis: "Analysis",
+    data: "Data",
+    tool_request: "Tool Request",
+    apps: "Apps",
+    permanent_id_request: "Permanent ID Request",
+    team: "Team",
+    community: "Community",
+};
+export default notificationCategory;
```
sri2k1us, line 12: 👍

src/components/notifications/listing/TableView.js
```diff
@@ -0,0 +1,241 @@
...
+                colSpan={3}
+                component="div"
+                count={total || 0}
+                rowsPerPage={rowsPerPage}
+                page={page}
+                onChangePage={handleChangePage}
+                onChangeRowsPerPage={handleChangeRowsPerPage}
+                ActionsComponent={TablePaginationActions}
+                rowsPerPageOptions={[25, 100, 500, 1000]}
+            />
```
sri2k1us, line 236: This should be updated to use `components/utils/DEPagination`;

src/components/notifications/listing/index.js
```diff
@@ -0,0 +1,178 @@
...
+                    messages: notifications.messages.map((n) =>
+                        selected.includes(n.message.id)
+                            ? { ...n, seen: true }
+                            : n
+                    ),
+                };
+
+                setMarkAsSeenEnabled(false);
+                setNotifications(newPage);
+                queryCache.setQueryData(notificationsKey, newPage);
```
sri2k1us, line 92: 👍

src/components/notifications/listing/index.js
```diff
@@ -0,0 +1,178 @@
...
+                markAsSeenEnabled={hasSelection && markAsSeenEnabled}
+                deleteEnabled={hasSelection}
+                onMarkSeenClicked={handleMarkSeenClick}
+                onDeleteClicked={handleDeleteClick}
+            />
+            <TableView
+                baseId={baseId}
+                data={notifications?.messages}
+                error={error}
+                loading={isFetching || markSeenLoading || deleteLoading}
```
sri2k1us, line 161: Minor suggestion - Since `mark as seen` is not really changing other that row highlighting from users perspective, we don't need a loading mask or skeleton. This could be silent background operation.

stories/notifications/Listing.stories.js
```diff
@@ -0,0 +1,45 @@
...
+import { mockAxios } from "../axiosMock";
+
+import { mockNotificationsListing } from "./NotificationMocks";
+
+import NotificationsListing from "components/notifications/listing";
+
+const errorResponse = {
+    error_code: "ERR_FALSE_ALARM",
+    reason: "Nothing to see here... Please try again!",
+};
```
sri2k1us, line 12: 😄

src/components/notifications/listing/TableView.js
```diff
@@ -0,0 +1,241 @@
...
+        }
+    };
+
+    const handleRequestSort = (event, property) => {
+        let newOrder = "desc";
+
+        if (orderBy === property && order === "desc") {
+            newOrder = "asc";
+        }
+
```
sri2k1us, line 135: `desc` and `asc` are now available from  `src/constants`.

sri2k1us (approved):
I had minor comments. LGTM 👍

src/components/notifications/listing/index.js
psarando: Good point.
I guess if the service returns an error, I can reset the highlight on those rows (assuming the user hasn't navigated away already) 🤔

sri2k1us: And I do see logic to display an error message 👍

psarando: The `Mark as Seen` action will no longer mask the listing with a loading skeleton, and will instantly mark the selected rows as seen. If an error is returned from the service, then the listing will now reload.

I also added logic so the `Mark as Seen` enabled flag will now update whenever the listing or the selection changes. That way if an error occurs and the user hasn't changed their selection or the page they are viewing, then the `Mark as Seen` button should be re-enabled so they can try again.

src/components/notifications/Toolbar.js
```diff
@@ -0,0 +1,104 @@
...
+                    notificationCategory.analysis,
+                    notificationCategory.data,
+                    notificationCategory.tool_request,
+                    notificationCategory.apps,
+                    notificationCategory.permanent_id_request,
+                    notificationCategory.team,
+                ].map((category) => (
+                    <MenuItem key={category} value={category}>
+                        {category}
+                    </MenuItem>
```
ash3rz, line 70: Can these MenuItems also have ids?

src/components/notifications/ids.js
```diff
@@ -0,0 +1,23 @@
...
+    APPROVE_BTN: "approveBtn",
+    DENY_BTN: "denyBtn",
+    SET_PRIVILEGE_DLG: "setPrivilegeDlg",
+    OK_BTN: "okBtn",
+    TOOLBAR: ".toolbar",
+    REFRESH_BTN: "refreshBtn",
+    MARK_ALL_SEEN_BTN: "markAllSeenBtn",
+    DELETE_BTN: "deleteBtn",
+    FILTER: "categoryCbo",
+    NOTIFICATION_VIEW: ".notificationView",
```
ash3rz, line 14: Minor, but I'm assuming we don't need these `.` in the ids anymore

src/components/notifications/listing/TableView.js
```diff
@@ -0,0 +1,245 @@
...
+                                        <Message
+                                            className={classnames(
+                                                classes.notification,
+                                                className
+                                            )}
+                                            message={n.message}
+                                            onMessageClicked={onMessageClicked}
+                                        />
+                                        <TableCell className={className}>
+                                            {formatDate(n.message.timestamp)}
```
ash3rz, line 214: This date cell and the category cell should probably also contain a `Typography` since the message cell does.  It's hard to tell from the screenshots, but I do think the message cell font looks slightly different from these two and it may be due to this?

ash3rz (approved):
LGTM :shipit: Just had some minor comments

psarando:
Thanks for the reviews! 👍 

I think I've addressed everything, and I've rebased on top of #194 which was just merged into master, so I'll go ahead and merge this one now.

Additional reviews still welcome 🙇

PR #200, opened 2020-10-14
Google analytics

sri2k1us:
image

### DIFF
--- a/src/gtag.js
+++ b/src/gtag.js
@@ -1,0 +1,11 @@
+export const GA_TRACKING_ID = "G-YXTRXCKR7T";
+
+//https://github.com/vercel/next.js/blob/canary/examples
+//https://developers.google.com/analytics/devguides/collection/gtagjs/pages
+export const pageview = (url) => {
+    if (window.gtag) {
+        window.gtag("config", GA_TRACKING_ID, {
+            page_path: url,
+        });
+    }
+};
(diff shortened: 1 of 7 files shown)


src/gtag.js
```diff
@@ -0,0 +1,11 @@
+export const GA_TRACKING_ID = "G-YXTRXCKR7T";
```
ash3rz, line 1: Should this also be in the config?

src/pages/_app.js
```diff
@@ -90,6 +92,16 @@ function MyApp({ Component, pageProps }) {
         queries: { refetchOnWindowFocus: false, retry: false },
     };
 
+    useEffect(() => {
+        const handleRouteChange = (url) => {
+            gtag.pageview(url);
+        };
+        router.events.on("routeChangeComplete", handleRouteChange);
+        return () => {
+            router.events.off("routeChangeComplete", handleRouteChange);
```
ash3rz, line 101: 👍 I always forget about the return function with useEffect

ash3rz (approved):
LGTM :shipit:

ianmcorvidae (approved):
Looks good as far as I can tell! I'm not totally sure about what GA provides us as far as looking at this data, but it seems like we're following best practices anyway.

sri2k1us:
Thanks for reviews.

PR #201, opened 2020-10-14
Move default data page ordering to listing

ash3rz:
This should fix the errors the data selection drawer was getting.  I just moved the default values that were in the page over to the listing directly to handle (since both the page and the selection drawer need those values).


sri2k1us (approved):
LGTM 👍

ash3rz:
Thanks for the review!

PR #202, opened 2020-10-15
Add static IDs for testing

ash3rz:
Just adding some IDs that would make testing easier


ash3rz:
Since this is just adding some IDs, I think I'll go ahead and merge.  Reviews still welcome though!

PR #203, opened 2020-10-15
Fix when dot menu items are displayed.

sri2k1us:
- iPhone 5
Note: When nothing is selected and browsing a folder thats is not writable, `dot menu` wouldn't show at all.
image

- iPad
image

- Desktop
image

- Desktop 
image

### DIFF
--- a/src/components/data/toolbar/DataDotMenu.js
+++ b/src/components/data/toolbar/DataDotMenu.js
@@ -64,23 +64,7 @@ function DataDotMenu(props) {
                 render={(onClose) => [
                     isSmall
                         ? [
-                              detailsEnabled && (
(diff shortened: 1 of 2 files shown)


ash3rz (approved):
LGTM :shipit: 

My head is starting to spin from trying to remember all the scenarios 😄 Thanks for looking into this 👍

sri2k1us:
> LGTM :shipit:
> 
> My head is starting to spin from trying to remember all the scenarios 😄 Thanks for looking into this 👍

No problem. Yes it is getting little complicated for Data view at least. We have to factor in a combination of screen size, permissions and selection count.

sri2k1us:
Thanks for the review.

PR #204, opened 2020-10-15
CORE-1143: partial wiring of app listings on the dashboard.

slr71:
Not everything is done here, but I wanted to submit a PR now just because I'm inexperienced when it comes to UI work, and I want to catch any errors early. The dashboard cards for apps now look like this:

image

Clicking on the `Favorite` icon when a user is not logged in will display a message indicating that an error occurred. attempting to obtain more details will display the usual login required dialog:

image

Clicking an app's `Launch` icon will now navigate to the launch screen for that app.

The `AppFavorite` component from the app details panel is being reused, so the icon color has changed to the primary color. Should the other icons be modified to use the primary color as well?

### DIFF
--- a/src/components/Dashboard/DashboardItem/AppItem.js
+++ b/src/components/Dashboard/DashboardItem/AppItem.js
@@ -1,14 +1,20 @@
-import React from "react";
+import React, { useState } from "react";
 
-import { Launch, Info, Favorite, People, Apps } from "@material-ui/icons";
+import { useMutation } from "react-query";
+
+import { Launch, Info, People, Apps } from "@material-ui/icons";
 import { IconButton, MenuItem } from "@material-ui/core";
+
+import FavoriteIcon from "@material-ui/icons/Favorite";
+import UnFavoriteIcon from "@material-ui/icons/FavoriteBorderOutlined";
 
 import { formatDate } from "@cyverse-de/ui-lib";
 
 import NavigationConstants from "common/NavigationConstants";
 
-import * as constants from "../constants";
+import { appFavorite } from "serviceFacades/apps";
 
+import * as constants from "../constants";
 import ItemBase, { ItemAction } from "./ItemBase";
 import { useTranslation } from "i18n";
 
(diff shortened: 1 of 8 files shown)


src/components/Dashboard/DashboardItem/AppItem.js
```diff
@@ -23,33 +28,69 @@ class AppItem extends ItemBase {
...
                 </ItemAction>,
                 <ItemAction
                     ariaLabel={t("launchAria")}
-                    key={`${constants.KIND_APPS}-${props.content.id}-launch`}
+                    key={buildKey("launch")}
                     tooltipKey="launchAction"
                 >
-                    <IconButton>
+                    <IconButton href={buildRef("launch")}>
                         <Launch />
```
slr71, line 85: I think I'm going to change this to the primary color, and I need to set the `IconButton` size to match the size of the `AppFavorite` icon. I'll do that in the next PR, though.

src/components/Dashboard/DashboardItem/AppItem.js
```diff
@@ -23,33 +28,69 @@ class AppItem extends ItemBase {
...
                 <ItemAction
-                    arialLabel={t("shareAria")}
-                    key={`${constants.KIND_APPS}-${props.content.id}-share`}
+                    ariaLabel={t("shareAria")}
+                    key={buildKey("share")}
                     tooltipKey="shareAction"
                 >
-                    <IconButton>
+                    <IconButton href={buildRef("share")}>
                         <People />
```
slr71, line 94: I'm going to change this to the primary color and change the size as well. Once again, this will happen in a subsequent PR.

src/components/Dashboard/DashboardItem/AppItem.js
```diff
@@ -23,33 +28,69 @@ class AppItem extends ItemBase {
...
-                    <IconButton>
-                        <Favorite />
-                    </IconButton>
+                    <AppFavorite
+                        isFavorite={isFavorite}
+                        isExternal={false}
+                        onFavoriteClick={onFavoriteClick}
+                        baseId={buildKey("favorite")}
+                    />
                 </ItemAction>,
```
slr71, line 78: I just realized that the `ItemAction` around `AppFavorite` only adds a (now redundant) tool tip. I'll remove it in the next PR.

ash3rz (approved):
Looking good so far to me! 👍

slr71:
Thanks for the review. I've got another PR coming up, so I'm going to merge this one now.

src/components/Dashboard/DashboardItem/AppItem.js
```diff
@@ -23,33 +28,69 @@ class AppItem extends ItemBase {
...
+        // Extract app details. Note: dashboard-aggregator only queries the DE database.
+        const app = props.content;
+
+        // State variables.
+        const [isFavorite, setIsFavorite] = useState(app.is_favorite);
+
+        // Functions to build keys and links.
+        const baseId = `${constants.KIND_APPS}-${app.system_id}-${app.id}`;
+        const buildKey = (keyType) => `${baseId}-${keyType}`;
+        const buildRef = (refType) =>
```
psarando, line 43: Minor nitpick, but I might name this `buildHRef` or `buildPath` or something, only because React already uses the `ref` term for DOM element refs, so I was confused for an instant (see https://reactjs.org/docs/refs-and-the-dom.html for example).

src/components/Dashboard/index.js
```diff
@@ -131,4 +135,4 @@ const Dashboard = () => {
     );
 };
 
-export default Dashboard;
+export default withErrorAnnouncer(Dashboard);
```
psarando, line 138: Rather than wrapping `Dashboard` and passing `showErrorAnnouncer` down through `DashboardSection` to every item/component, could only `AppItem` be wrapped, since it's the only item that needs this for now?

I haven't tried wrapping a class component yet, but I don't see why it shouldn't work. So the last line of `src/components/Dashboard/DashboardItem/AppItem.js` could be updated to this:
```jsx
export default withErrorAnnouncer(AppItem);
```

psarando (commented):
Looks good 👍 ✨ 

I just had a couple of minor suggestions for a future PR.

src/components/Dashboard/index.js
psarando: Ah 👌 
I suspected there was a reason, so thanks for clarifying.

Maybe another way to prevent passing this prop around would be to create a helper component in `AppItem` that wraps `AppFavorite` and the mutation query and all, which could be wrapped in `withErrorAnnouncer`, and used in `AppItem` where `AppFavorite` is used now.

Maybe something like:
```jsx
const AppItemFavoriteIcon = withErrorAnnouncer((props) => {
        const { baseId, app, showErrorAnnouncer } = props;

        const [isFavorite, setIsFavorite] = useState(app.is_favorite);

        const [favorite] = useMutation(appFavorite, {
            onSuccess: () => {
                setIsFavorite(!isFavorite);
            },
            onError: (e) => {
                showErrorAnnouncer(t("favoritesUpdateError", { error: e }), e);
            },
        });

        const onFavoriteClick = () => {
            favorite({
                isFav: !app.is_favorite,
                appId: app.id,
                systemId: app.system_id,
            });
        };

        return (
                    <AppFavorite
                        isFavorite={isFavorite}
                        isExternal={false}
                        onFavoriteClick={onFavoriteClick}
                        baseId={baseId}
                    />
        );
});
```

This is just some copy-paste from this PR and it's not tested, but just to give an idea 😅

slr71: I could give that a try. 👍

PR #206, opened 2020-10-16
Bag Fixes and Improvements

johnworth:
Various fixes for the bag feature that couldn't make it into the first PR without endangering its inclusion in the user testing period.

Work includes:

- Fixing typos.
- Only allowing shareable analyses to get added.
- Only allowing owned apps to get added to the bag.
- Add more static IDs.
- Add a button to copy links to the clipboard to the download links dialog.


src/components/Bag/downloads/index.js
```diff
@@ -55,12 +61,38 @@ const DownloadLinksDialog = ({
...
+            const fullLink = `${getHost()}/api/download?path=${curr}`;
+            if (acc !== "") {
+                acc = `${acc}\n${fullLink}`;
+            } else {
+                acc = fullLink;
+            }
+            return acc;
+        }, "");
+
+        navigator.clipboard
```
ash3rz, line 78: For anyone else, I was curious about the difference between the copy method we're using [here](https://github.com/cyverse-de/ui-lib/blob/master/src/components/copy/clipboardFunctions.js) vs. this one.  Seems it's totally up to you which way you want to go with according to [this](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard)

ash3rz (approved):
LGTM :shipit:

sri2k1us (changes requested):
Looks good but I found couple of nagging issues.
1. I don't see a `Share` button anymore when I add file(s) I own. Seems to work right in `prod` though.
image

2. Icon for `Clear` should match what it is mobile version. 
i.e It should be `import ClearAllIcon from '@material-ui/icons/ClearAll';`

3. Rename Bag folder to lowercase `bag`

johnworth:
This is old and out of date. I don't even remember working on it at this point. 😄 

Closing this out for now.

PR #208, opened 2020-10-16
CORE-1143: dashboard app details.

slr71:
This change enables app details to be displayed on the dashboard and adds some cosmetic changes to the dashboard.


src/components/dashboard/index.js
```diff
@@ -78,6 +79,9 @@ const Dashboard = (props) => {
     const isLoading = status === "loading";
     const hasErrored = status === "error";
 
+    // State variables.
+    const [detailsApp, setDetailsApp] = useState(null);
+
     // TODO: Unify error handling across components, somehow.
     if (hasErrored) {
         console.log(error.message);
```
sri2k1us, line 87: Console msgs can be removed ?

sri2k1us (changes requested):
So I checked this PR in my local machine and found few issues
1. Apps thats are favorited are not showing as favorited in Dashboard
image

2. I know app launch from Dashboard is not part of this PR but launching from dashboard causes whole app refresh. The launch button should be wrapped with `next/Link`

3. Launch icon should be `import PlayArrowIcon from '@material-ui/icons/PlayArrow';` and Share Icon should be `import ShareIcon from '@material-ui/icons/Share';`

4. Minor but after `Details` is selected from vertical menu, the menu stays open. Menu is expected to close after an option is selected.

I think at least 1 and 2 should be fixed before the testing.

src/components/apps/AppFavorite.js
```diff
@@ -7,7 +7,13 @@ import FavoriteIcon from "@material-ui/icons/Favorite";
 import UnFavoriteIcon from "@material-ui/icons/FavoriteBorderOutlined";
 
 export default function AppFavorite(props) {
-    const { isFavorite, isExternal, onFavoriteClick, baseId } = props;
+    const {
+        isFavorite,
+        isExternal,
+        onFavoriteClick,
+        baseId,
+        size = "small",
```
psarando, line 15: 👍

src/components/dashboard/dashboardItem/AppItem.js
```diff
@@ -28,7 +28,7 @@ class AppItem extends ItemBase {
 
     static create(props) {
         const item = new AppItem(props);
-        const { showErrorAnnouncer } = props;
+        const { showErrorAnnouncer, setDetailsApp } = props;
         const { t } = useTranslation(["dashboard", "apps"]);
```
psarando, line 32: Since `getFavoriteActionKey` was removed, then I think this can now be reverted:
```suggestion
        const { t } = useTranslation("dashboard");
```

src/components/dashboard/index.js
```diff
@@ -123,13 +128,21 @@ const Dashboard = (props) => {
...
-            ref={dashboardEl}
-            id={fns.makeID(ids.ROOT)}
-            className={classes.gridRoot}
-        >
+        <div ref={dashboardEl} id={baseId} className={classes.gridRoot}>
             {isLoading ? <DashboardSkeleton /> : componentContent}
+            {detailsApp !== null && (
+                <AppDetailsDrawer
+                    appId={detailsApp?.id}
+                    systemId={detailsApp?.system_id}
```
psarando, line 140: This could be shorter (and cover more falsey cases):
```suggestion
            {detailsApp && (
                <AppDetailsDrawer
                    appId={detailsApp.id}
                    systemId={detailsApp.system_id}
```
Also no need to check for falsey `detailsApp` again when getting its IDs.

psarando (commented):
LGTM 👍 

Just had a couple of code style/cleanup suggestions ✨

ash3rz (commented):
I didn't catch anything that wasn't already mentioned 👍 Seems to be coming along nicely! 🎉

src/components/dashboard/index.js
slr71: Ah, I hadn't noticed that console message.

1. I added an optional callback function to the details panel. If the callback function is specified, it will be called with the new favorite flag value.

I'll make the suggested changes for the rest of the items.

slr71:
I think I've addressed all of the issues. I'm a real novice at UI work, though, so I'd appreciate some reviews if possible.

sri2k1us:
LGTM 👍. Some nice to fix, minor nit picks but not necessary for this PR 
 - The spacing between the icons in App Item seems to be messed up.
- App Fav Icon doesn't show `pointer` cursor instead it shows default one.

Also - Don't forget to push the related `terrain` changes to prod. 👍

sri2k1us (approved):

slr71:
Thanks for the reviews!

PR #210, opened 2020-10-21
CORE-1140 Add range select support to Notifications listing

psarando:
This PR will add range selection support to the Notifications listing table, which probably should have been added in #199.

This will also standardize selection behavior so that when a row is selected (without the shift-key) all other selections are cleared.

Selection using the row's checkbox works as before, toggling a single row's selection, or with range selection support when using the shift-key.

Selecting a notification's message selects only that row, ignoring the shift-key, since message selection will execute some
other action (navigating to an app or data folder, for example).

### DIFF
--- a/src/components/notifications/listing/TableView.js
+++ b/src/components/notifications/listing/TableView.js
@@ -98,6 +98,7 @@ const TableView = (props) => {
     } = props;
 
     const [page, setPage] = React.useState(0);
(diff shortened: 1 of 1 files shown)


ash3rz (approved):
LGTM :shipit:

sri2k1us (approved):
LGTM 👍

psarando:
Thanks for the reviews! 👍

PR #211, opened 2020-10-21
Search UX improvements.

sri2k1us:
- Set filter according to the page viewed.
- Navigate to app launch when selecting an app search result.


src/components/layout/CyVerseAppBar.js
```diff
@@ -241,8 +241,19 @@ function CyverseAppBar(props) {
...
 
+    if (activeView === NavigationConstants.APPS) {
+        filter = searchConstants.APPS;
+    } else if (activeView === NavigationConstants.ANALYSES) {
+        filter = searchConstants.ANALYSES;
+    } else if (activeView === NavigationConstants.DATA) {
+        filter = searchConstants.DATA;
+    } else {
+        filter = searchConstants.ALL;
+    }
```
ianmcorvidae, line 255: So, I don't want to be a pest here, but I'm honestly a bit worried that this approach limits the usefulness of global search for our users, and I'm wondering if we can do better.

Specifically, doing it this way means that for the majority of pages in our app, a filter will already be applied; discovering that the search can search across everything at once requires either:

a.) noticing the filter dropdown, using it, noticing that there's an "All" option, and trying it. Some folks will do this, but more novice users, who are often our target, aren't that likely to get past the first or second step.
b.) trying the search on one of the small number of pages that doesn't have a filter in place. The dashboard is the most likely one of these, but it still seems like a small funnel for people to know about the feature.

All that being said, I think it's obviously valuable to prioritize searching within the kind of thing you're already looking at. I don't think this needs to be changed immediately or anything, but I wonder if we should add another pathway to the global search, perhaps another entry in the results similar to the "view all <data/apps/analyses> results for <search term>" that instead points users towards the global search, maybe something like (in the data view) "View Apps and Analyses for <search term>" (or similarly for the other filters). This would be especially valuable when there's no results for the given filter, so a user can with one click go from a "no results" search within one subtype, to a broader global search that may give them the results they want.

Anyway, food for thought.

src/components/layout/CyVerseAppBar.js
```diff
@@ -289,6 +300,8 @@ function CyverseAppBar(props) {
         }
     }, [userProfile]);
 
+    useEffect(() => {}, [activeView]);
```
ianmcorvidae, line 303: I think I'm probably missing something. What does this do?

ianmcorvidae (commented):
One UX concern, one question. Things look good to me otherwise, but I'm not sure I'm qualified to approve or anything either, so I just marked this as a comment.

src/components/layout/CyVerseAppBar.js
sri2k1us: This does nothing cos this code should have been deleted and I didn't 🤦

src/components/layout/CyVerseAppBar.js
sri2k1us: I like this idea. I can easily implement it too. I will wait for few others to comment to get a broader feedback.

ash3rz: Good points.  I think another thing could be if you're in the Apps view, for example, then the app search results could be on top as opposed to always having Data results on top.  That might also help prioritize without hiding stuff with a filter?  Either way, I do like the idea of adding another entry to the results that signals to users that it can be a global search 👍

psarando: I like the idea of a `view all` or `search all` option when searching under a filter 👍

ash3rz (approved):
LGTM :shipit:

src/components/layout/CyVerseAppBar.js
sri2k1us: @ash3rz - Only problem is right now we can't control which search (app / data / analyses) results comes back first. We can still do some thing to makes sure the results are properly ordered but that will take changes / refactor to get it right.

sri2k1us: I will try to implement @ianmcorvidae suggestion for now. Let see what our users think.

sri2k1us: Ok I have implemented @ianmcorvidae suggestion. Can I get a review again ? Thanks.

psarando (commented):
I think I caught a copy/paste error, otherwise LGTM.

src/components/search/GlobalSearchField.js
```diff
@@ -375,12 +463,17 @@ function GlobalSearchField(props) {
...
-                    resultType: { type: t("data"), id: searchConstants.DATA },
-                };
-                setOptions([...options, ...data, viewAll]);
+                if (filter === searchConstants.DATA) {
+                    setOptions([
+                        ...options,
+                        ...data,
+                        viewAllDataOptions,
+                        viewAllAppOptions,
+                        viewAllDataOptions,
```
ash3rz, line 472: This last one is a dupe - guessing it should probably be `viewAllAnalysesOptions`?

ash3rz (approved):
I may have caught one bug, but otherwise LGTM :shipit:

sri2k1us:
Thanks for the reviews. I fixed the issues identified. Also found a bug with tab selection. Fixed that too.

PR #214, opened 2020-10-28
CORE-1150 Add Notification message navigation links

psarando:
This PR will refactor the Notification listing so that analysis, app, and data notification messages are now links that navigate the user to the appropriate page.

Other messages are no longer clickable and are not displayed as a link.

Other refactors are also included in this PR, such as standardizing the `NotificationCategory` model and migrating the `getDisplayMessage` function into a `notifications/utils` namespace.


src/components/utils/NameLink.js
```diff
@@ -38,7 +38,7 @@ const NameLink = React.forwardRef((props, ref) => {
                 }
             }}
         >
-            <Highlighter search={searchTerm}>{name}</Highlighter>
+            <Highlighter search={searchTerm || ""}>{name}</Highlighter>
```
psarando, line 41: Without this change a warning was issued if the `searchTerm` prop was not set in `NameLink`, although `Highlighter` appears to display its child components just fine (like a no-op) with an empty `search` prop.

src/components/data/utils.js
```diff
@@ -29,7 +29,7 @@ function getEncodedPath(path) {
  * @param {string} folderPath The data store path to the folder.
  */
 const getFolderPage = (folderPath) =>
-    `${constants.PATH_SEPARATOR}${NavigationConstants.DATA}${constants.PATH_SEPARATOR}ds${folderPath}`;
+    `${constants.PATH_SEPARATOR}${NavigationConstants.DATA}${constants.PATH_SEPARATOR}${constants.DATA_STORE_STORAGE_ID}${folderPath}`;
```
sri2k1us, line 33: 👍

src/components/layout/Notifications.js
```diff
@@ -114,7 +112,7 @@ function Notifications(props) {
         (notification, category) => {
             let analysisStatus =
                 category.toLowerCase() ===
-                    constants.NOTIFICATION_CATEGORY.ANALYSIS.toLowerCase() ||
+                    NotificationCategory.ANALYSIS.toLowerCase() ||
```
sri2k1us, line 115: 👍

src/components/models/notificationCategory.js
```diff
@@ -1,12 +0,0 @@
...
-    new: "New",
-    analysis: "Analysis",
-    data: "Data",
-    tool_request: "Tool Request",
-    apps: "Apps",
-    permanent_id_request: "Permanent ID Request",
-    team: "Team",
-    community: "Community",
-};
-export default notificationCategory;
```
sri2k1us, line 12: I thought we will are using file names starting with uppercase only for Components 🤔 . Regardless standardizing this will be good.

src/components/notifications/Message.js
```diff
@@ -0,0 +1,137 @@
...
+import NameLink from "components/utils/NameLink";
+
+import { Typography } from "@material-ui/core";
+
+function MessageLink(props) {
+    const { message, href, as } = props;
+
+    return (
+        <Link href={href} as={as} passHref>
+            <NameLink name={message} />
```
sri2k1us, line 28: Now that we are using `NameLink` for other things as well should this be renamed to something link `DELink` ?

src/components/notifications/Message.js
```diff
@@ -0,0 +1,137 @@
...
+                id: app.category_id,
+            })
+        );
+
+        return <MessageLink message={message} href={href} />;
+    }
+
+    return message;
+}
+
```
sri2k1us, line 83: When I tried this locally, clicking on an `App` Sharing notification caused an `400` error. The address in URL bar was 
`http://localhost:3000/apps?selectedOrder=null&selectedOrderBy=null&selectedPage=null&selectedRowsPerPage=null&selectedFilter=null&selectedCategory={%22system_id%22:%22de%22,%22id%22:%22eeeeeeee-eeee-eeee-eeee-eeeeeeeeeeee%22}`
image



I think Clicking on an app notification should take users to the individual app listing using this following `href` and `as`

const href = `/${NavigationConstants.APPS}/[systemId]/[appId]`;	
    const as = `/${NavigationConstants.APPS}/${selectedApp?.system_id}/${selectedApp?.id}`;

sri2k1us (changes requested):
LGTM but I may have found a bug which needs to investigated / fixed.

src/components/models/notificationCategory.js
ash3rz: I don't think there's any standard or convention really out there, but for some reason my mind seems to think stuff in the models directory seem better suited as PascalCase. I guess because they're supposed to represent some aspect of a data object model, and object = PascalCase to me?  

Our `index` files are also components but they're not PascalCase so 😄 🤷‍♀️ (not saying they should be PascalCase, just an interesting exception)

ash3rz (commented):
I didn't catch anything, only had some minor comments 👍

src/components/models/notificationCategory.js
sri2k1us: Agreed. Having models starting with uppercase seems reasonable to me. WRT `index` files, everyone seems to use lowercase for that. I guess that convention is beyond `react` comes from the `js` community but 🤷‍♂️

psarando: I guess I was also thinking JSON model = object = PascalCase.

I think the `index.js` naming convention goes back to traditional webpage directories, where a URL like `my.site.example.com/some/thing/` would retrieve the document at `my.site.example.com/some/thing/index.html` by default, so an `import SomeThing from "some/thing";` will look for the file `some/thing/index.js` by default. And URLs were traditionally case-insensitive, so I'm used to seeing all lowercase HTML docs and I can't recall ever seeing an `Index.html` file. The `index.js` naming convention is probably also following the same tradition.

So if everyone agrees, I like standardizing on components and models using PascalCase file names.
And things like `util` files, which are just a collection of functions, seem better suited as all lowercase file names.

But I also have no objection to standardizing on lowercase model file names either, and reserving PascalCase strictly for components, if anyone has a strong opinion on that.

sri2k1us: Lets settle for PascalCase for models.

src/components/notifications/Toolbar.js
```diff
@@ -58,21 +58,21 @@ const NotificationToolbar = (props) => {
...
-                    "permanent_id_request",
-                    "team",
+                    "NEW",
+                    "ALL",
+                    "ANALYSIS",
+                    "DATA",
+                    "TOOL_REQUEST",
+                    "APPS",
+                    "PERMANENT_ID_REQUEST",
+                    "TEAM",
```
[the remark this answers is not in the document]
psarando: Ah, I had considered it, but that would change the order of the menu items, and I thought they were ordered this way for a specific reason (with `New` on top, followed by `All`, etc.).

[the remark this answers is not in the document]
ash3rz: 🤔 I'd have to test it, but I don't think `Object.keys()` reorders the list of keys. So if you have the `NotificationCategory` object set up with the keys in the order you want to view them in as MenuItems, I think that should still work?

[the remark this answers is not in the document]
ash3rz: (if not, no big deal either)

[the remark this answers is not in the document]
psarando: Yes, it does appear to preserve the order of the keys in this model object anyway, but I think it was setup with `All` first, followed by `New`.

I'll go with that and then we can reorder the keys of the category model as needed.

src/components/notifications/Message.js
psarando: So I was attempting to use `components/apps/utils/getListingPath` by passing `null` for all params except the `category` param, hoping that the code that built or parsed the URL would strip params with `null` values, but it looks like this function is currently building a raw string for the URL and these nulls become the string "null" when parsed by the target page, and I didn't test this in dev mode before opening this PR 😞 

I agree that this link should take the user to the specific app in the listing, so I've made the suggested change, but first I was curious and tried refactoring `getListingPath` to return a URL `{pathname, query}` object compatible with `next/link` and `Router.push`, but it appears `next/link` also preserves `null` values set in that URL's `query` object 😞

psarando:
Thanks for the reviews! 👍 
I think I've addressed all comments and made the suggested changes, so I'll go ahead and merge this PR now, but additional reviews still welcome 🙇

PR #216, opened 2020-10-30
Minor sharing updates

ash3rz:
This should add the user's CyVerse ID to the sharing dialog's user table.  It didn't make sense that you could search for user ID but not see the user ID in the table.

image

This will also fix the issue where you could make some modifications to the sharing dialog (add or remove some users, for example), then hit cancel, reopen the sharing dialog, and see all the modifications again as if they had saved.

### DIFF
--- a/public/static/locales/en/sharing.json
+++ b/public/static/locales/en/sharing.json
@@ -1,6 +1,5 @@
 {
     "cannotAddSelf": "Cannot add yourself for sharing",
-    "cyverseId": "CyVerse ID: {{id}}",
     "dataGroupSharingDisabled": "Data and/or analyses resources cannot be shared with groups at this time",
     "fetchPermissionsError": "Failed to fetch resource permissions",
     "fetchUserInfoError": "Failed to fetch user information",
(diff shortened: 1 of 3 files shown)


public/static/locales/en/sharing.json
```diff
@@ -1,5 +1,6 @@
 {
     "cannotAddSelf": "Cannot add yourself for sharing",
+    "cyverseId": "CyVerse ID: {{id}}",
```
ash3rz, line 3: I went with `CyVerse ID:` but perhaps `ID:` or `Username:` would work too?

psarando: `Username` makes sense.

psarando (approved):
LGTM :shipit:

public/static/locales/en/sharing.json
sri2k1us: But the search field prompt says `CyVerse ID`. So if we are changing, we should be consistent.

sri2k1us (approved):
LGTM :shipit:

ash3rz:
Thanks for the reviews!

PR #217, opened 2020-11-02
Instant Launch Proof-of-Concept

johnworth:
Proof of concept project to show that instant launches can be built on top of existing APIs. Even if the overall PoC doesn't work out, additional APIs and work on Sonora will get finished.


package.json
```diff
@@ -77,6 +77,7 @@
         "js-yaml": "^3.13.1",
         "jsonpath-plus": "^4.0.0",
         "keycloak-connect": "^7.0.1",
+        "micromatch": "^4.0.2",
```
johnworth, line 80: I'm using this for the glob matching support in the instant launch mapping. infoType matching is also supported, and regex matching should be easy to add.

src/components/data/listing/Listing.js
```diff
@@ -47,6 +47,11 @@ import {
     INFO_TYPES_QUERY_KEY,
 } from "serviceFacades/filesystem";
 
+import {
+    getDefaultsMapping,
+    DEFAULTS_MAPPING_QUERY_KEY,
+} from "serviceFacades/instantlaunches";
```
johnworth, line 53: I put this here because I wasn't sure that this needed to be a context, and I read that contexts should be used sparingly.

src/components/data/listing/Listing.js
```diff
@@ -452,7 +472,11 @@ function Listing(props) {
         }
     };
 
-    const isLoading = isQueryLoading([isFetching, removeResourceStatus]);
+    const isLoading = isQueryLoading([
+        isFetching,
+        removeResourceStatus,
+        isFetchingDefaultsMapping,
+    ]);
```
johnworth, line 479: Updated to account for the new API call to fetch the defaults mapping. Still not 100% that I like having the additional API call here, but I'm unsure about where else to put it so that it will get reloaded periodically.

src/components/data/listing/TableView.js
```diff
@@ -135,8 +140,9 @@ function TableView(props) {
         onMetadataSelected,
         onPublicLinksSelected,
         rowDotMenuVisibility,
-        onDownloadSelected,
```
johnworth, line 138: This was raising warnings in my IDE about it not being used, so I nuked it.

src/components/data/listing/TableView.js
```diff
@@ -361,7 +372,7 @@ function TableView(props) {
                                                 )}
                                                 onClick={() => {
                                                     handlePathChange(
-                                                        `${path}/${resource.label}`,
+                                                        resourcePath,
```
johnworth, line 375: Only changed because we're now using the resourcePath in a couple of locations.

src/components/data/listing/TableView.js
```diff
@@ -286,10 +292,15 @@ function TableView(props) {
...
                                 const resourceId = resource.id;
+                                const resourcePath = `${path}/${resource.label}`;
                                 const isSelected =
                                     selected.indexOf(resourceId) !== -1;
                                 const isInvalid =
                                     isSelected && isInvalidSelection(resource);
+                                const [instantLaunch] = defaultInstantLaunch(
+                                    instantLaunchDefaultsMapping,
+                                    resource
+                                );
```
johnworth, line 303: This returns the instant launch object that is marked as the default for the resource, or null if there isn't one. If we ever want to support having multiple instant launches possible for a file, this is one of the calls that will end up changing.

src/components/instantlaunches/index.js
```diff
@@ -0,0 +1,176 @@
...
+    resource,
+    showErrorAnnouncer,
+}) => {
+    const { t } = useTranslation(["instantlaunches", "common"]);
+    return (
+        <IconButton
+            variant="contained"
+            onClick={async (event) => {
+                event.stopPropagation();
+                event.preventDefault();
```
johnworth, line 155: The above two lines prevent a click on the instant launch button from selecting the resource row in the data listing. That got a little confusing and could imply that we support sending multiple files to a single instantly launched app, which isn't the case (yet).

src/pages/apps/[systemId]/[appId]/launch.js
```diff
@@ -69,7 +69,7 @@ export default function Launch() {
         },
     });
 
-    const { status: qLuanchStatus } = useQuery({
+    const { status: qLaunchStatus } = useQuery({
```
Comment (johnworth) [inline src/pages/apps/[systemId]/[appId]/launch.js:72]: Fixed a typo I ran across.

src/pages/apps/[systemId]/[appId]/launch.js
```diff
@@ -80,7 +80,7 @@ export default function Launch() {
     });
 
     const loading =
-        appStatus === constants.LOADING || qLuanchStatus === constants.LOADING;
+        appStatus === constants.LOADING || qLaunchStatus === constants.LOADING;
```
Comment (johnworth) [inline src/pages/apps/[systemId]/[appId]/launch.js:83]: Same typo fix.

src/serviceFacades/analyses.js
```diff
@@ -17,6 +17,14 @@ function getAnalyses(key, { rowsPerPage, orderBy, order, page, filter }) {
...
 }
 
+function getAnalysis(id) {
+    const filter = JSON.stringify({ field: "id", value: id });
+    return callApi({
+        endpoint: `/api/analyses?filter=[${filter}]`,
+        method: "GET",
+    });
+}
+
```
johnworth, line 27: Similar to the existing call, but simplifies the calling code a bit.

stories/data/DataMocks.js
```diff
@@ -637,6 +637,297 @@ export const pathListChunk = {
     ],
 };
 
+export const listingSuccessResp = {
```
johnworth, line 640: I found myself adding to the mocks for the data listing quite a bit, so I moved it into the DataMock.js file to clean up the data listing Story a bit. Hopefully that's okay.

stories/data/listing/Listing.stories.js
```diff
@@ -0,0 +1,73 @@
...
+        .reply(200, instantLaunchGlobalQuickLaunches);
+
+    mockAxios
+        .onGet(/\/api\/quicklaunches\/a4b1f851-80c0-415d-ba3c-6663432e4f7e.*/)
+        .reply(200, instantLaunchQuickLaunch);
+
+    mockAxios
+        .onPost(/\/api\/analyses.*/)
+        .reply(200, instantLaunchSubmissionResponse);
+
```
johnworth, line 71: I'm tempted to the instant launch related API mocks into a separate file and call them here, since it might be confusing to have them all dumped here like this.

PR #218, opened 2020-11-04
Fix Chrome crashing while loading data

ash3rz:
This should fix an extraneous call to userinfo happening, but more importantly fix an issue in Chrome that occurs when navigating to the Data page.  In the console, you get this warning over and over again:
```
Throttling navigation to prevent the browser from hanging. See https://crbug.com/882238. Command line switch --disable-ipc-flooding-protection can be used to bypass the protection
```

Now we're basically defining a default path with default order, orderBy, etc. to minimize the number of redirects that happen. 

Without this change, we have `/data` -> `/data/ds` ->`/data/ds/somePath` -> `/data/ds/somePath?someParams`.  
With this change we have `/data` ->`/data/ds/somePath?someParams`
Or `/data/ds` -> `/data/ds/somePath?someParams`

### DIFF
--- a/src/components/data/dataFields.js
+++ b/src/components/data/dataFields.js
@@ -1,7 +1,9 @@
+import { DEFAULT_PAGE_SETTINGS } from "./utils";
+
 const dataFields = (dataI18n) => ({
     NAME: {
         fieldName: dataI18n("name"),
-        key: "name",
+        key: DEFAULT_PAGE_SETTINGS.orderBy,
     },
     LAST_MODIFIED: {
         fieldName: dataI18n("modified"),
--- a/src/components/data/dataFields.js
+++ b/src/components/data/dataFields.js
@@ -1,9 +1,7 @@
-import { DEFAULT_PAGE_SETTINGS } from "./utils";
-
 const dataFields = (dataI18n) => ({
     NAME: {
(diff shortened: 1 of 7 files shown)


src/components/data/dataFields.js
```diff
@@ -1,7 +1,9 @@
+import { DEFAULT_PAGE_SETTINGS } from "./utils";
+
 const dataFields = (dataI18n) => ({
     NAME: {
         fieldName: dataI18n("name"),
-        key: "name",
+        key: DEFAULT_PAGE_SETTINGS.orderBy,
```
ianmcorvidae, line 6: This feels weird to me since it's a key-value thing that's under a hardcoded `NAME` -- I guess I'd expect this to stay hardcoded, not use the default page settings map.

ianmcorvidae (approved):
One thing that feels weird to me, but looks good otherwise! Who knows if I might've missed something, of course.

src/components/data/dataFields.js
ash3rz: Looking at this now, I agree - I'll switch it back

sri2k1us (approved):
LGTM :shipit:

ash3rz:
Thanks for the reviews!

PR #220, opened 2020-11-12
Public Links (aka Send to Genome Browser)

sri2k1us:
- Users can select multiple files at once to create public links
- Created links can be Copied, Saved to a local file or it can be saved to file in Data Store.
     
image

image

-iPhone 5
image


sri2k1us:
So Nirav wants to add an option to directly copy link to a file (single selection). Should that be a separate option in addition to `Public Link(s)` menu item at the row level or the row level menu item should only provide option to directly copy the link?

src/components/data/PublicLinks.js
```diff
@@ -0,0 +1,247 @@
...
+    );
+
+    useEffect(() => {
+        if (download) {
+            window.open(
+                `${getHost()}/api/downloadText?text=${encodeURIComponent(
+                    links
+                )}`,
+                "_blank"
+            );
```
ash3rz, line 121: I may be missing something, but it seems like instead of using the `download` state and useEffect, this could just be a normal function that gets called when the `Download` button gets clicked?

src/components/utils/DEDialog.js
```diff
@@ -0,0 +1,62 @@
...
+    const { t } = useTranslation("common");
+    const theme = useTheme();
+    const fullScreen = useMediaQuery(theme.breakpoints.down("sm"));
+    const dialogTitleId = build(baseId, ids.DIALOG.TITLE);
+    return (
+        <Dialog
+            open={open}
+            onClose={onClose}
+            maxWidth="sm"
+            fullWidth
```
ash3rz, line 37: Not necessary for this PR, but might be nice to make `maxWidth` and `fullWidth` configurable with props but also have these as default values perhaps

src/components/utils/DEDialog.js
```diff
@@ -0,0 +1,62 @@
...
+            fullScreen={fullScreen}
+            aria-labelledby={dialogTitleId}
+        >
+            <DialogTitle>
```
ash3rz, line 41: I see `aria-labelledby` being set in the `Dialog`, but I think this `DialogTitle` is then missing the `dialogTitleId` being set

src/components/utils/DEDialog.js
```diff
@@ -0,0 +1,62 @@
...
+                        position: "absolute",
+                        right: theme.spacing(0.5),
+                        top: theme.spacing(0.5),
+                        margin: 0,
+                    }}
+                >
+                    <CloseIcon />
+                </IconButton>
+            </DialogTitle>
+            <DialogContent>{content}</DialogContent>
```
ash3rz, line 56: This works - I also personally wouldn't mind seeing the dialog content come from `children` instead of a `content` prop, but it's a small thing

src/server/api/fileio.js
```diff
@@ -64,5 +64,13 @@ export default function fileIORouter() {
     logger.info("adding the GET /api/download handler");
     api.get("/download", auth.authnTokenMiddleware, downloadHandler);
 
+    logger.info("adding the GET /api/downloadText handler");
+    api.get("/downloadText", function (req, res) {
+        res.set({
+            "Content-Disposition": 'attachment; filename="de-untitled.txt"',
```
ash3rz, line 70: If I'm understanding right, the `filename` property is a default name that gets auto-filled in for the `Save As` dialog that pops up for the user?  Maybe `cyverse-download.txt` or something similar might be a default that users in a hurry won't have to change too frequently?

ash3rz (approved):
LGTM :shipit: Only had a few minor comments.

As for the single selection "copy link to a file" option - could that be something added to the data details drawer?  We already have an option to copy the iRods path, we could add an option to copy the full URL too perhaps.  Other than that, I'd say we could add a new menu item, but keep in mind that we might end up having a lot of menu items (once sonora has feature parity with the current DE) and might reconsider at that point if it should remain a menu item.

src/components/data/PublicLinks.js
sri2k1us: `window` can be `undefined` when use outside `useEffect` since Nextjs does SSR. `useEffect` is guaranteed to run on the browser. Same logic is used in File Viewers `Toolbar.js` as well. Please see for https://dev.to/vvo/how-to-solve-window-is-not-defined-errors-in-react-and-next-js-5f97 for more info.

src/server/api/fileio.js
sri2k1us: This logic is not used in `Save As`. This used when users request downloading of `links` to the local machine. I could name it to something like `de-links.txt` but this function is pretty generic so  I don't want to do that. May be I can a filename in query param ? 🤔

src/components/data/PublicLinks.js
ash3rz: Got it - thanks for the link!

src/server/api/fileio.js
ash3rz: You're right - for some reason I thought when you download something, you have an opportunity to change the name of the downloaded file, but I guess not.  It would be nice if the downloaded filename could be the same as the original file name 👍

sri2k1us: Some correction - This is not a downloading a file from Datastore. But rather, we create a new file in memory in Express with all the links requested by the user and then that file is downloaded.

sri2k1us: `Copy Link(s)` - Will copy links to clipboard
`Download` - Will download the links as a file to users local computer
`Save to file` - Will save to links to a file in the Data store.

ash3rz: Got it - really my only hesitation was with the `untitled` part of the file name. If `de-links` is too specific then maybe `de-download` or something similar could work?

ianmcorvidae: I've been vaguely following this thread, and my feeling is: a query parameter is good if we want to have a suggested filename, but the default should just be `Content-Disposition: attachment` -- that will still set it to be downloaded, but without a filename specified. I don't think "de-untitled.txt" will help anyone, so it's better left off so the browser/client/user can decide what to name the file.

psarando (approved):
LGTM 👍 

I only had a couple of minor suggestions in addition to those already made.

src/server/api/fileio.js
sri2k1us: Query Params seems to be best options but if you don't set a default file name, chrome downloads it as `downloadText.html`(HTML file) and firefox opens it as `downloadText`. I still think setting a default file name is good idea and that file name should start with `de`.

src/components/utils/DEDialog.js
sri2k1us: I would have gone with Children if it weren't for `dialogActions`. Should `dialogActions` be a `prop` or should that be Children too ? 🤔

sri2k1us:
> LGTM :shipit: Only had a few minor comments.
> 
> As for the single selection "copy link to a file" option - could that be something added to the data details drawer? We already have an option to copy the iRods path, we could add an option to copy the full URL too perhaps. Other than that, I'd say we could add a new menu item, but keep in mind that we might end up having a lot of menu items (once sonora has feature parity with the current DE) and might reconsider at that point if it should remain a menu item.

I think Nirav wants a quicker way (less clicks) to get a link for single file. So putting it in `Details` won't really solve it 🤔. I am leaning towards changing the `Public Link(s)` row level menu item to `Copy Public Link` which when clicked will directly copy the link to clipboard.

src/components/utils/DEDialog.js
sri2k1us: Interesting, react docs refer to the way I have it currently - https://reactjs.org/docs/composition-vs-inheritance.html

src/server/api/fileio.js
ianmcorvidae: Ah, that makes sense. It's unfortunate that the browser defaults aren't nicer, and `de-download.txt` does seem like a better option than theirs.

src/components/utils/DEDialog.js
psarando: I wouldn't include `actions` in the `children` too. Then it we would have to use it like this:
```jsx
<DEDialog
    baseId={...}
    open={...}
    title={...}
    ...
>
    <DialogContent>{...}</DialogContent>
    <DialogActions>{...}</DialogActions>
</DEDialog>
```
To my mind it makes sense to have the content as the `children`, and optional `actions` as a prop:
```jsx
<DEDialog
    baseId={...}
    open={...}
    title={...}
    actions={[...]}
    ...
>
    {...}
</DEDialog>
```

sri2k1us:
Thanks for the reviews.

PR #221, opened 2020-11-12
Use react-player light mode to improve performance of dashboard page

ianmcorvidae:
Not a super huge amount to this. Mostly just changes the iframe to a div (which then adopts the width/height/etc), and inside it puts a `ReactPlayer` with the `light` option set, which uses noembed.com to fetch a thumbnail and uses that, superimposing a play button that loads the actual full player.

Here's how it looks:

Screenshot_2020-11-12 Discovery Environment

### DIFF
--- a/src/common/NavigationConstants.js
+++ b/src/common/NavigationConstants.js
@@ -17,5 +17,5 @@ export default {
     SEARCH: "search",
     SETTINGS: "settings",
     VICE: "vice",
-    YOUTUBE_EMBED_BASE: "https://www.youtube-nocookie.com/embed",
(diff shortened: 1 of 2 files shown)


ianmcorvidae:
This shows the video titles on hover, not sure if that's sufficient or if we need to figure out a way to show the titles more prominently.

sri2k1us (approved):
LGTM :shipit:

ash3rz (approved):
:shipit:

psarando (approved):
:shipit:

ianmcorvidae:
Thanks for the reviews!

PR #223, opened 2020-11-12
Make next dynamic imports work with jest, and dynamically load VideoViewer and TextViewer

ianmcorvidae:
This is another thing I decided to take a stab at while I was thinking about sonora bundle sizes and performance. Without this PR, every data listing and file view loads all the viewers, which includes several hundred kilobytes of `highlight.js` (and the smaller but still notable 45-ish kilobytes that are the `react-player` File player). Previously, this didn't work, because jest would get tripped up on the `next/dynamic` imports, but with the other additions here, dynamic imports can work correctly for jest too. The changes in `FileViewer.js` are the ones that improve performance, the rest of the files are changes to make dynamic imports work with jest.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -11,5 +11,12 @@
         }
       }
     ]
-  ]
+  ],
+  "env": {
+    "test": {
+      "plugins": [
+        "babel-plugin-dynamic-import-node"
+      ]
+    }
+  }
 }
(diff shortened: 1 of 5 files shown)


sri2k1us (approved):
I have approved #224. So this is just a formality.

PR #224, opened 2020-11-13
Remove react-table from initial loading dependencies.

ianmcorvidae:
Includes commits from #223 because this also uses `next/dynamic` and another test needed the appropriate treatment.

The core things here:

 * import SharingView dynamically in the Bag component. This is only used if the user is logged-in and it pulls in a whole bunch of stuff, so splitting it into a separate chunk seems warranted.
 * move some constants, `SEARCH_RESULTS_TABS`, into their own namespace. This way, their former namespace, `DetailedSearchResults`, doesn't need to all be brought into namespaces that require those constants, perhaps most notably the global search field that appears on every page.

With only the second of those (which I did first), the "shared by all pages" JS went from 350kB to 304kB. With both, it's down to 284kB. Ideally we could cut this down a lot more, but `next-i18next` and what it brings along is a bit more than 70k by itself, and next.js's main/normal stuff is also about 70k. A lot of the rest is material-ui stuff. The biggest remaining target would be getting rid of the old vestiges of `react-intl` and our `I18NWrapper`, I think.

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,6 +131,7 @@
         "fetch-mock": "^9.5.0",
         "husky": "^4.0.10",
         "jest": "^24.9.0",
+        "jest-next-dynamic": "^1.0.1",
         "jsdoc": "^3.6.3",
         "lint-staged": "^10.2.2",
         "nodemon": "^2.0.2",
--- a/src/components/Bag/index.js
+++ b/src/components/Bag/index.js
@@ -1,5 +1,7 @@
 import React, { useState, useEffect, useCallback } from "react";
 
+import dynamic from "next/dynamic";
+
 import {
     Badge,
     Dialog,
@@ -32,8 +34,6 @@ import {
     ShoppingBasket as ShoppingBasketIcon,
 } from "@material-ui/icons";
 
-import SharingView from "components/sharing";
-
 import { build as buildID } from "@cyverse-de/ui-lib";
 
 import constants from "./constants";
@@ -55,6 +55,8 @@ import {
 import { useTranslation } from "i18n";
 import { useTheme } from "@material-ui/styles";
 import { useUserProfile } from "contexts/userProfile";
+
+const SharingView = dynamic(() => import("components/sharing"));
 
 const useStyles = makeStyles((theme) => ({
     help: {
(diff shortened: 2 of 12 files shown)


src/components/Bag/index.js
```diff
@@ -56,6 +56,8 @@ import { useTranslation } from "i18n";
 import { useTheme } from "@material-ui/styles";
 import { useUserProfile } from "contexts/userProfile";
 
+const SharingView = dynamic(() => import("components/sharing"));
+
```
sri2k1us, line 60: 👍

package.json
```diff
@@ -131,6 +131,7 @@
         "fetch-mock": "^9.5.0",
         "husky": "^4.0.10",
         "jest": "^24.9.0",
+        "jest-next-dynamic": "^1.0.1",
```
sri2k1us, line 134: 👍

src/components/data/viewers/FileViewer.js
```diff
@@ -45,6 +44,10 @@ import {
     Typography,
 } from "@material-ui/core";
 
+// these are at the bottom so eslint doesn't complain
+const TextViewer = dynamic(() => import("./TextViewer"));
+const VideoViewer = dynamic(() => import("./VideoViewer"));
```
sri2k1us, line 49: I am curious if we could use the async build provided `react-syntax-highlighter` directly and still get the same performance improvement. https://github.com/react-syntax-highlighter/react-syntax-highlighter#async-build

sri2k1us (approved):
LGTM. I tried this locally and everything checks out as expected. :shipit:

src/components/data/viewers/FileViewer.js
ianmcorvidae: I admit I couldn't get it to work, but I could give it another shot. We could get additional benefits from it, at least, because with the async build we could make sure we only load the languages that we actually use. If I did that, what I might consider doing is removing the dynamic imports here, but dynamically importing `FileViewer` itself (along with `react-syntax-highlighter`, of course). That way, the file viewer code doesn't end up loaded when just viewing listings, but all the viewers get loaded at once (and then the syntax highlighter only when needed).

ianmcorvidae:
Thanks for the review! I'll merge them now so I don't lose track of it, but all of this can of course continue being refined.

PR #225, opened 2020-11-16
First load niceties

ianmcorvidae:
Basically, three (small) changes here I found as I was nosily poking around, as usual:

 * make it so `useTheme` is always imported from `@material-ui/core`. This makes a test work when an otherwise-useless `Hidden` is removed, which makes it so the Bag icon is server-side rendered. I also went through the rest of the code to import it consistently, though.
 * remove aforementioned `Hidden`
 * add a width and height to the DE logo `img`, so that there isn't a layout shift as it loads

### DIFF
--- a/src/components/Bag/index.js
+++ b/src/components/Bag/index.js
@@ -23,6 +23,7 @@ import {
     Typography,
     FormControl,
     InputLabel,
+    useTheme,
 } from "@material-ui/core";
 import {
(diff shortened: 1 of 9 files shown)


ash3rz (approved):
LGTM :shipit:

ianmcorvidae:
Cool, thanks for the lookover!

PR #226, opened 2020-11-16
Fix issue where logged in users still see anonymous's data roots

ash3rz:
Small change.  I think having the `staleTime: Infinity, cacheTime: Infinity,` was causing this bug, so now I'm just invalidating the results whenever the userprofile changes. Still not sure why this was only happening sometimes, but it seems to happen more consistently for me now that anonymous doesn't have a home directory and is returning an error.

### DIFF
--- a/src/components/data/toolbar/Navigation.js
+++ b/src/components/data/toolbar/Navigation.js
@@ -328,6 +328,7 @@ function Navigation(props) {
     const [userProfile] = useUserProfile();
     const [config] = useConfig();
     const irodsHomePath = config?.irods?.home_path;
+    const rootsQueryKeyArray = [DATA_ROOTS_QUERY_KEY, userProfile?.id];
 
     const preProcessData = (respData) => {
         if (respData) {
@@ -373,12 +374,8 @@ function Navigation(props) {
         }
     }, [dataRoots, handlePathChange, path]);
 
-    useEffect(() => {
-        queryCache.invalidateQueries(DATA_ROOTS_QUERY_KEY);
-    }, [userProfile]);
-
     const { error } = useQuery({
-        queryKey: DATA_ROOTS_QUERY_KEY,
+        queryKey: rootsQueryKeyArray,
         queryFn: getFilesystemRoots,
         config: {
             enabled: true,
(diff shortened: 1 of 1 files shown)


psarando (approved):
:shipit:

sri2k1us (approved):
LGTM 👍

ianmcorvidae:
So, this seems fine to me, but also, wouldn't it make more sense to include the username or some sort of user profile inforrmation in the react-query query key for this? Then, when it changes react-query should know to change it/refetch it anyway, I think

ash3rz:
> So, this seems fine to me, but also, wouldn't it make more sense to include the username or some sort of user profile inforrmation in the react-query query key for this? Then, when it changes react-query should know to change it/refetch it anyway, I think

Ah, good point!  Let me try that out because I'm not sure if it's possible that the `Infinity` caching clobbers that query key check, but it sounds right to me.

ianmcorvidae:
So, I got a debugger into this and it looks like what's happening is the roots endpoint preprocessing that adds the dummy "Home" value for anonymous is clearly putting that in, even though the rest are all getting correctly pulled from the roots endpoint response, and then later it tries to `dataRoots.findIndex` with `root.path === userHomePath` -- `userHomePath` comes from the `base-paths` key of the roots response, so it has the user's proper path, so it gets a -1 index, which pulls out an undefined when it tries to use it (this is the `g[m] is undefined` error). I guess this suggests to me that it's a race condition between the roots endpoint and the user profile, so hopefully putting user profile info into the query key will prevent it (or at least make it retry).

ianmcorvidae:
I guess another thing that we could do that could help is if `selectedIndex` is ever < 0, something's wrong, or at least, we shouldn't depend on it as an index into any arrays. But it's probably better if we can just prevent it from hitting this state, lol

ash3rz:
@ianmcorvidae that makes sense! Thanks for digging into this!  Let's try this for now since the query should re-run based on the user ID changing.  The QA runs have had the most luck in getting into the race condition so hopefully that's a good validation check.

ash3rz:
Thanks for the reviews!

src/components/data/toolbar/Navigation.js
```diff
@@ -328,6 +328,7 @@ function Navigation(props) {
     const [userProfile] = useUserProfile();
     const [config] = useConfig();
     const irodsHomePath = config?.irods?.home_path;
+    const rootsQueryKeyArray = [DATA_ROOTS_QUERY_KEY, userProfile?.id];
```
psarando, line 331: Should this also be set in a `useEffect` with `userProfile` as a dependency?

ash3rz: 🤔 thinking about it, yeah, I guess that's what we've done elsewhere.  It seemed to work fine locally, but I can refactor to be consistent.

PR #227, opened 2020-11-16
Add fetch-mock mocks for noembed.com thumbnail-fetching for dashboard

ianmcorvidae:
This is a bit ugly, recommendations welcome. But it does get the dashboard back to working.


ianmcorvidae:
PS the bodies I've put in these mocks are just from doing the actual requests with `curl`

src/components/dashboard/dashboardItem/ItemBase.js
```diff
@@ -212,7 +212,7 @@ export const DashboardVideoItem = ({ item }) => {
         >
             <ReactPlayer
                 url={item.getLinkTarget()}
-                light={true}
+                light={item.content.thumbnailUrl || true}
```
ianmcorvidae, line 215: This change means that if the data that comes from dashboard-aggregator includes a `thumbnailUrl` (as I added manually to one of the videos in the data.json below) then noembed won't be called at all.

ash3rz (approved):
LGTM 👍

ianmcorvidae:
Cool, I'll merge this for now. If I update dashboard-aggregator to provide the thumbnails, I can update the data file here to reflect that too (though I might leave one that uses noembed so that keeps being tested, little though it is).

PR #228, opened 2020-11-16
Attempt to only build Chromatic CI for PRs and pushes in main repo

psarando:
Hopefully this will prevent Chromatic builds for pushes in fork repos, except when a PR is open against our main repo.

I'm not sure how to test this except to merge it and see what happens 🤷‍♂️ 

Also, perhaps something to consider is their warning that "The pull-request event requires special consideration": https://www.chromatic.com/docs/github-actions#recommended-configuration-for-build-events.
I'm not exactly sure if that will apply to this change, since we will still include the `push` event.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,9 +1,13 @@
 name: Chromatic CI
 
-on: [push]
+on:
+  push
+  pull_request:
+    types: [synchronize]
 
 jobs:
   build:
+    if: github.repository == 'cyverse-de/sonora'
 
     runs-on: ubuntu-latest
 


sri2k1us (approved):
LGTM :shipit:   Go for merge.

psarando:
🆗 If it doesn't work we can just revert the change.

I'll make sure Chromatic doesn't build my fork once it syncs with this merge, then we'll just need to make sure that Chromatic still builds for the next PR that's opened, or for new commits to an already-opened PR.

psarando:
Looks like I missed that by default a workflow will run "when a `pull_request`'s activity type is `opened`, `synchronize`, or `reopened`."
https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request

We probably want `opened` to trigger a build as well, so I'll push a commit to leave it at the defaults for now.

PR #229, opened 2020-11-17
Revert Chromatic CI GitHub workflow changes

psarando:
This reverts #228 and subsequent commits to the workflow config.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,13 +1,9 @@
 name: Chromatic CI
 
-on:
-  push
-  pull_request:
-    types: [synchronize]
+on: [push]
 
 jobs:
   build:
-    if: github.repository == 'cyverse-de/sonora'
 
     runs-on: ubuntu-latest
 
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,6 +1,9 @@
 name: Chromatic CI
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+    types: [synchronize]
 
 jobs:
   build:
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,7 +1,7 @@
 name: Chromatic CI
 
 on:
-  push:
+  push
   pull_request:
     types: [synchronize]
 


sri2k1us (approved):
:shipit:

.github/workflows/chromatic.yml
```diff
@@ -1,10 +1,9 @@
 name: Chromatic CI
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
   build:
-    if: github.repository == 'cyverse-de/sonora'
```
psarando, line 7: One more suggestion based on @ianmcorvidae's idea:

We could leave this line in to limit Chromatic builds if we modify our PR workflow and only open PRs from branches in `cyverse-de/sonora`. So we could do our dev process as usual, and work in branches in our own forks, but once something is ready for a PR, we would have to create a branch in `cyverse-de/sonora` and push the commits from our fork's branch to the duplicate branch in `cyverse-de/sonora`, then open the PR from the `cyverse-de/sonora` branch.

If more commits or rebases are need, those could still happen in our fork's branches, but then they would have to be merged/pushed/force-pushed to the `cyverse-de/sonora` duplicate branch in order to show up in the PR.

This also has the benefit of allowing us to make a bunch of commits in our fork's branch without triggering a build for each one in Chromatic. Then when we feel the changes are ready for the PR, they can all be pushed at once to the `cyverse-de/sonora` duplicate branch, which will update the PR and trigger only 1 build in Chromatic for all the changes at once.

We don't have to decide on adopting this workflow for Sonora now, but it could be something to keep in mind.

ash3rz: 🤔 That's an interesting idea.  I do like that we could have more control over when the PRs actually get updated once they've been created.  I know I tend to be a bit neurotic about keeping my branches up-to-date with the main branch so I probably cause a ton of Chromatic re-builds. The downside I guess is a bit more branch management, but that's pretty minor.

ash3rz (approved):
:shipit:

.github/workflows/chromatic.yml
psarando: I'll go ahead and merge this now and we can ruminate on an alternative workflow some more and revisit this later.

But first, an example of how this workflow might work (assuming most of the team uses `origin` for their fork and `upstream` for the main cyverse-de repo):
```
$ git checkout -b feature-branch
$ git push -u origin feature-branch
$ git commit -a "my new feature"
$ git push
$ git push upstream feature-branch
# open PR using feature-branch in cyverse-de repo and get feedback
$ git fetch --all
$ git rebase upstream/master
$ git commit -a "Address PR comments"
$ git push --force
$ git push --force upstream feature-branch
# Merge PR and delete branches
```

PR #230, opened 2020-11-18
Sonora - Product Tour

sri2k1us:
- Tour Prompt
image
- Interactive. i.e Waits for users to click on the hamburger menu.
image
- Tour steps can be navigated back / next
image

Notes:
- This covers only tour for Overall DE. 
- We should / can add tour to specific pages like Data, Apps, Analyses etc... 
- The tour text is modified version of the one in the current prod DE. Will be reviewed by Tina.

Known Issues:
- Has a scrolling issue in iphone 5 for long text.


public/static/locales/en/intro.json
```diff
@@ -0,0 +1,24 @@
...
+    "apps": "<p>Use the Apps View to view and find apps to analyze your data.</p><p>In the Apps View, you can: <ul><li>Browse for all available apps in CyVerse.</li><li>Customize an existing app.</li><li>Create your own App.</li></ul></p>",
+    "data": "<p>Use the Data view to manage your data in our cloud-based, central repository called the CyVerse Data Store.<ul><li>You can transfer data in/out of CyVerse.</li><li>Access data you own, data shared with you, and public data.</li><li>Share your data with other CyVerse or external users.</li></ul></p>",
+    "notifications": "<p>Automated notifications let you track progress of actions, such as: <ul><li>The status of your analyses, data transfers, and any requests you have made.</li><li>Sharing of files and folders with you or by you.</li></ul></p>",
+    "preferences": "<p>Use the Preferences to customize settings for your DE experience.<ul><li>Customize notification preferences.</li><li>Change settings and modify keyboard shortcuts.</li></ul></p>",
```
ash3rz, line 17: I think "view got chopped off this sentence: "Use the Preferences `view`..."

public/static/locales/en/intro.json
```diff
@@ -0,0 +1,24 @@
...
+    "bag": "<p>Use a bag to perform bulk operations. <ul><li>Share Analyses, Apps, and Data from the bag.</li><li>Download multiple data sets from the bag.</li></ul></p>",
+    "dashboard": "<p>Use Dashboard to find information about: <ul><li>Recently launched analyses.</li><li>Recently used Apps.</li><li>New Apps published in the DE.</li><li>CyVerse New and Events.</li></ul></p>",
+    "search": "<p>Use Global Search to: <ul><li>Type 3 or more characters to quickly search and find Analyses, Apps and Data.</li><li>Filter search results by type.</li><li>View detailed results page by hitting the `Enter` key followed by the search term or selecting the `View All` from the results.</li></ul></p>",
```
ash3rz, line 20: To me, this sounds like you're supposed to hit Enter and then type out a search term.  Maybe `... by hitting the Enter key after the search term`? 
Also for `... or selecting the View All from ...` - I would either drop `the` from this sentence or keep `the` and add what `View All` is - such as `View All link` or `View All option`

public/static/locales/en/intro.json
```diff
@@ -0,0 +1,24 @@
...
+    "support": "<p>Use this in-app chat to <ul><li>Contact CyVerse support when you encounter issue(s) with the DE or any Cyverse products.</li><li>Ask a question about the DE or any Cyverse products.</li><li>Get notified about Cyverse planned maintenance and unplanned downtimes and events such as webinar.</li></ul></p>",
```
ash3rz, line 21: In this one, there are a couple occurrences of `Cyverse` instead of `CyVerse`

public/static/locales/en/intro.json
```diff
@@ -0,0 +1,24 @@
...
+    "tourPrompt": "Welcome to CyVerse. Do you like to take a tour of the Discovery Environment?",
```
ash3rz, line 22: I think `Would you like to take a tour...` might sound a little better here

src/components/layout/CyVerseAppBar.js
```diff
@@ -575,160 +621,200 @@ function CyverseAppBar(props) {
...
+                <main
+                    className={clsx(classes.content, {
+                        [classes.contentShift]: open,
+                    })}
+                >
+                    {children}
+                </main>
+                {newUser && (
+                    <ConfirmationDialog
+                        baseId={"de"}
```
ash3rz, line 781: This id should perhaps go in an IDs file but I would give it something more descriptive like `userTourDlg` or something like that.  
Maybe the `baseId` prop is misleading and the prop should be `id` instead so it's clear it's the ID for the dialog and not like a parent ID (not for this PR, just thinking out loud).

src/components/layout/CyVerseAppBar.js
```diff
@@ -575,160 +621,200 @@ function CyverseAppBar(props) {
...
+                            paper: clsx({
+                                [classes.drawerOpen]: open,
+                                [classes.drawerClose]: !open,
+                            }),
+                        }}
+                    >
+                        <div className={classes.toolbar}>
+                            {!open && (
+                                <IconButton
+                                    aria-label="open drawer"
```
ash3rz, line 704: I know this isn't a change in this PR, but this should probably go through i18n

ash3rz (approved):
LGTM 👍 
I'm sure Tina/MMS will have better feedback for the wording, but I added some of my own and mentioned a couple other minor things.

psarando (approved):
LGTM 🎉 

I only had a couple of minor suggestions 👍

src/components/layout/CyVerseAppBar.js
```diff
@@ -299,6 +306,8 @@ function CyverseAppBar(props) {
         bootstrapQueryEnabled,
         (respData) => {
             setPreferences(respData.preferences);
+            const workspace = respData["apps_info"].workspace;
+            setNewUser(workspace["new_workspace"] || false);
```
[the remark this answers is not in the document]
sri2k1us: what if `workspace["new_workspace"]` is undefined ? Then won't newUser be to set `undefined` ? 🤔

sri2k1us:
Thanks for the reviews.

src/components/layout/CyVerseAppBar.js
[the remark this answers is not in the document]
psarando: Yes, but isn't that effectively the same thing as far as this state is concerned?
If we only check `if (newUser)` later, then `undefined` would be the same as `false` in that check.
Unless I'm missing something with setting React state to `undefined`.

PR #231, opened 2020-11-19
Sharing improvements.

sri2k1us:
- Feedback when no matching users / group found
image
     We can provide a search field prompt label (above the search field) instead of this long feedback. 
- Custom renderer for options (group vs user)
image

### DIFF
--- a/public/static/locales/en/sharing.json
+++ b/public/static/locales/en/sharing.json
@@ -3,7 +3,7 @@
     "dataGroupSharingDisabled": "Data and/or analyses resources cannot be shared with groups at this time",
     "fetchPermissionsError": "Failed to fetch resource permissions",
     "fetchUserInfoError": "Failed to fetch user information",
-    "noUsersFound": "No users found",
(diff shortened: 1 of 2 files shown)


public/static/locales/en/sharing.json
```diff
@@ -3,7 +3,7 @@
     "dataGroupSharingDisabled": "Data and/or analyses resources cannot be shared with groups at this time",
     "fetchPermissionsError": "Failed to fetch resource permissions",
     "fetchUserInfoError": "Failed to fetch user information",
-    "noUsersFound": "No users found",
+    "noUsersFound": "No user(s) found. Enter the full CyVerse username or email of the user you wish to share.",
```
ash3rz, line 6: 🤔 You mentioned this in the PR description and now I'm wondering if having this as some text above the search would be nicer.  This looks good for now though, maybe we can wait to see how any additional feedback goes.

ash3rz (approved):
LGTM :shipit:

sri2k1us:
Thanks for the review.

PR #232, opened 2020-11-19
Login menu

sri2k1us:
User Menu shows when logged in avatar is clicked in large screen. Shows by default in Mobile.

-iPhone 5
image

- Laptop
image


public/static/locales/en/common.json
```diff
@@ -30,6 +33,7 @@
     "noOptions": "No results to display.",
     "openDrawer": "Open Drawer",
     "ok": "OK",
+    "polices": "Polices",
```
ash3rz, line 36: 😄 I didn't notice this when reviewing the screenshots, but this should be `Policies`

src/components/layout/UserMenu.js
```diff
@@ -0,0 +1,188 @@
...
+            color: theme.palette.info.contrastText,
+        },
+    },
+}));
+export default function UserMenu(props) {
+    const {
+        profile: {
+            id,
+            attributes: { email, name },
+        },
```
ash3rz, line 63: Instead of having `userProfile` passed in as a prop, couldn't this component use the `useUserProfile` hook for itself?

src/components/layout/CyVerseAppBar.js
```diff
@@ -721,6 +741,28 @@ function CyverseAppBar(props) {
...
+                    vertical: "top",
+                    horizontal: "left",
+                }}
+            >
+                <UserMenu
+                    baseId={build(ids.APP_BAR_BASE, ids.ACCOUNT_MI)}
+                    profile={userProfile}
+                    onLogoutClick={onLogoutClick}
+                    onManageAccountClick={() =>
+                        window.open("https://user.cyverse.org", "_blank")
```
ash3rz, line 762: The CyVerse policy and about URLs are in a constant file, maybe the user portal URL should also be in one?

src/components/layout/UserMenu.js
```diff
@@ -0,0 +1,188 @@
...
+                        <Typography
+                            variant="caption"
+                            className={classes.typography}
+                        >
+                            {t("polices")}
+                        </Typography>
+                    </ExternalLink>
+                </Grid>
+                <Grid item>
+                    <Typography variant="caption">.</Typography>
```
ash3rz, line 170: I think, if we're trying to replicate Google's user menu, this is actually a • dot (Opt+8 on mac) and not a regular . period.  Switching in the dot should help it look a little more vertically centered

ash3rz (approved):
LGTM :shipit:

public/static/locales/en/common.json
sri2k1us: `i` catching!

src/components/layout/UserMenu.js
sri2k1us: We can do that for sure. Only downside I can think of is that it will involve checking profile to see if  it null or not which is already happening in `CyverseAppBar` in several places. Otherwise I can't think of any reason either in favor or against.

src/components/layout/CyVerseAppBar.js
sri2k1us: Yup..missed this one.

src/components/layout/UserMenu.js
sri2k1us: Lol. GTK that it is (opt + 8) cos I was searching for a bigger dot than a period in my keyboard 😄

sri2k1us:
Gonna merge this one but reviews still welcome.

PR #233, opened 2020-11-19
Add Mark All as Seen functionality

flynn1999:
serviceFacades has the API function call added, notifications menu sets unSeenCount as 0 by making POST request, resets badge value and sets seen value of all notifications as true to change CSS after re-render.

I needed advice on how to get rid of es-lint error of having the unused variable isMarked from line 136. As of have a `console.log` to overcome it. So it will need changing before merge.

### DIFF
--- a/src/components/notifications/NotificationsMenu.js
+++ b/src/components/notifications/NotificationsMenu.js
@@ -7,9 +7,8 @@ import classnames from "classnames";
 import {
     getNotifications,
     markAllSeen,
-    NOTIFICATIONS_MARK_ALL_SEEN_KEY,
     NOTIFICATIONS_MESSAGES_QUERY_KEY,
-} from "serviceFacades/notifications";
+} from "../../serviceFacades/notifications";
 import { useTranslation } from "../../i18n";
 import ids from "./ids";
 import NotificationStyles from "./styles";
@@ -133,7 +132,6 @@ function NotificationsMenu(props) {
 
     const handleMarkAllAsSeenClick = () => {
         markAllSeenMutation();
-        setAllNotificationsSeen();
         handleClose();
     };
 
@@ -170,23 +168,19 @@ function NotificationsMenu(props) {
         },
     });
 
-    const [markAllSeenMutation] = useMutation({
-        queryKey: NOTIFICATIONS_MARK_ALL_SEEN_KEY,
-        queryFn: markAllSeen,
-        config: {
-            onSuccess: () => {
-                setAllNotificationsSeen();
-            },
-            onError: (error) => {
-                console.log("Error marking all notifications as Seen");
-                // *** Will add this after understanding HOC better. ***
-                // showErrorAnnouncer(
-                //     t("errorMarkAsSeen", {
-                //         count: notifications.length,
-                //     }),
-                //     error
-                // );
-            },
+    const [markAllSeenMutation] = useMutation(markAllSeen, {
+        onSuccess: () => {
+            setAllNotificationsSeen();
+        },
+        onError: (error) => {
+            console.log("Error marking all notifications as Seen");
+            // *** Will add this after understanding HOC better. ***
+            // showErrorAnnouncer(
+            //     t("errorMarkAsSeen", {
+            //         count: notifications.length,
+            //     }),
+            //     error
+            // );
         },
     });
 
--- a/src/components/notifications/NotificationsMenu.js
+++ b/src/components/notifications/NotificationsMenu.js
@@ -118,7 +118,13 @@ function InteractiveAnalysisUrl(props) {
 }
 
 function NotificationsMenu(props) {
-    const { setUnSeenCount, notificationMssg, setAnchorEl, anchorEl } = props;
+    const {
+        setUnSeenCount,
+        notificationMssg,
+        setAnchorEl,
+        anchorEl,
+        showErrorAnnouncer,
+    } = props;
     const [notifications, setNotifications] = useState([]);
     const [error, setError] = useState();
     const classes = useStyles();
@@ -173,14 +179,7 @@ function NotificationsMenu(props) {
             setAllNotificationsSeen();
         },
         onError: (error) => {
-            console.log("Error marking all notifications as Seen");
-            // *** Will add this after understanding HOC better. ***
-            // showErrorAnnouncer(
-            //     t("errorMarkAsSeen", {
-            //         count: notifications.length,
-            //     }),
-            //     error
-            // );
+            showErrorAnnouncer(t("errorMarkAsSeen"), error);
         },
     });
 
--- a/src/components/notifications/NotificationsMenu.js
+++ b/src/components/notifications/NotificationsMenu.js
@@ -18,6 +18,7 @@ import NavigationConstants from "common/NavigationConstants";
 
 import ExternalLink from "components/utils/ExternalLink";
 import ErrorTypographyWithDialog from "components/utils/error/ErrorTypographyWithDialog";
+import withErrorAnnouncer from "../utils/error/withErrorAnnouncer";
 
 import { build } from "@cyverse-de/ui-lib";
 import {
@@ -349,4 +350,4 @@ function NotificationsMenu(props) {
     );
 }
 
-export default NotificationsMenu;
+export default withErrorAnnouncer(NotificationsMenu);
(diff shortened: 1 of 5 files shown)


src/components/notifications/NotificationsMenu.js
```diff
@@ -149,6 +133,57 @@ function NotificationsMenu(props) {
...
+        );
+    };
+
+    const handleClose = () => {
+        setAnchorEl();
+    };
+
+    const handleMarkAllAsSeenClick = () => {
+        // set badge value as 0.
+        setUnSeenCount(0);
```
sri2k1us, line 173: We should reset this count only after calling the service and getting a success response. Right now I get `404` for `/api/notifications/mark-all-seen` cos the corresponding handler in `/src/server/api/notifications.js` is missing.

src/components/notifications/NotificationsMenu.js
```diff
@@ -149,6 +133,57 @@ function NotificationsMenu(props) {
...
+                // showErrorAnnouncer(
+                //     t("errorMarkAsSeen", {
+                //         count: notifications.length,
+                //     }),
+                //     error
+                // );
+            },
+        },
+    });
+
```
sri2k1us, line 155: 2 issues:
- I don't think you can de-structure `isMarked` or a random variable from useQuery. Please refer https://react-query.tanstack.com/docs/api to find out what values are returned by the `useQuery`.
- This query is now running (although failing) any time sonora is viewed. This should only run when a user click `Mark All as Read` button. To do this, you should enable this query from `handleMarkAllAsSeenClick` function. Please refer to the `api` page I linked above more info.

sri2k1us (changes requested):
Good Start but as of now this code doesn't work. I suggest you test the functionality you are developing in Storybook and Dev. environment. As of the lint issue, `isMarked` is an unused variable so you get warning but refer to my comments on what variable you can de-structure from `useQuery`. `isMarked` is NOT returned by `useQuery`.

src/components/notifications/NotificationsMenu.js
ash3rz: I see based on what you've commented out that you've been referencing some of Paul's code for marking a selection of notifications as seen:
https://github.com/cyverse-de/sonora/blob/9293b3c2de6f39d2ed96fa107afc2055ac10b60e/src/components/notifications/listing/index.js#L95-L109

The really big difference that you're missing (it's easy to miss, the differences are slight, and I also missed this during our Zoom chat, sorry) is that you're using `useQuery` and if you look carefully you'll see Paul is using `useMutation`.  `useQuery` is appropriate only for `GET` requests where you are only fetching or reading information.  
In Paul's case, and in your case as well, you're not just reading information, you're modifying data with a `POST` request.  For these types of requests, you must use `useMutation` which operates a little differently than `useQuery`. See docs [here](https://react-query.tanstack.com/docs/guides/mutations).

I think you'll find if you review the docs carefully and Paul's code, you'll be able to accomplish what Sriram mentioned above where this call needs to only trigger after an action (clicking on a button) occurs.

src/components/notifications/NotificationsMenu.js
```diff
@@ -149,6 +133,57 @@ function NotificationsMenu(props) {
...
+    const setAllNotificationsSeen = () => {
+        console.log(isMarked);
+        setNotifications(
+            notifications?.map((item) => {
+                const tempItem = item;
+                tempItem.seen = true;
+                return tempItem;
```
ash3rz, line 162: It looks like `tempItem` is meant to be a copy, and this isn't quite right.  I'm sure in your CS classes you've come across or will very soon come across the idea of copying by value or copying by reference.  Javascript will copy an object, like `item` in this case, by reference.  

[Here](https://attacomsian.com/blog/javascript-clone-objects) is an article I found real quick that will explain why this is incorrect and some methods for doing it correctly.  We prefer the `spread` method, which is mentioned in that article, and I would suggest you use that here. Understanding destructuring and the spread syntax will be key in understanding a lot of our code base.

src/components/notifications/NotificationsMenu.js
ash3rz: I think I may have helped cause some confusion here 😖 .  The other day he asked me if an API call failing was expected and in the screenshot was the error page for `/api/bootstrap`.  I said that the bootstrap endpoint is expected to fail because iRods was currently down because I thought that's what he was asking me and I think he took my response as his API call is expected to fail. Sorry Flynn!

ash3rz (changes requested):
Good start, and sorry about confusion/miscommunication on some of the things here, but you're really close and only need a few more changes 👍

src/components/notifications/NotificationsMenu.js
sri2k1us: Good catch. This should indeed use `useMutation`.

flynn1999:
> Good start, and sorry about confusion/miscommunication on some of the things here, but you're really close and only need a few more changes 👍

It's fine. I should have tested it on dev more but since it was down was down, I thought it was working. I'll make sure to set aside dedicated time for testing. 
These documentations are helpful. Thank you Ashley and Sriram.

flynn1999:
Still needs testing but I've made the necessary changes. Will update on Tuesday.

src/components/notifications/NotificationsMenu.js
ash3rz: I think you said you hadn't tested this yet, but if it doesn't work, I believe you're still missing a few refactors for switching over to `useMutation`.  The [api reference](https://react-query.tanstack.com/docs/api#usemutation) might make these differences a little more clear than the guide docs that I linked earlier.  You can also review Paul's code.

src/components/notifications/NotificationsMenu.js
```diff
@@ -149,6 +133,56 @@ function NotificationsMenu(props) {
...
+        setUnSeenCount(0);
+    };
+
+    const handleClose = () => {
+        setAnchorEl();
+    };
+
+    const handleMarkAllAsSeenClick = () => {
+        markAllSeenMutation();
+        setAllNotificationsSeen();
```
ash3rz, line 173: Having `setAllNotificationsSeen()` get called here when the user clicks the `Mark All As Seen` button means that this will trigger even if the API call to set all notifications as seen fails.  I don't think you need this here - having the call in `onSuccess` like you already have should be sufficient.

ash3rz:
I did a quick checkout of your branch, ran it locally, and I believe all the features requested are in place 🎉 
One last thing that'll help us get the green light on your PR is to fix your branch.  For some reason, it says you have 100+ commits and 100+ files changed.  There may have been something that went wrong during a rebase?

src/components/notifications/NotificationsMenu.js
```diff
@@ -163,6 +174,26 @@ function NotificationsMenu(props) {
         },
     });
 
+    const [markAllSeenMutation] = useMutation(markAllSeen, {
+        onSuccess: () => {
+            setAllNotificationsSeen();
+        },
+        onError: (error) => {
+            showErrorAnnouncer(t("errorMarkAsSeen"), error);
```
ash3rz, line 182: You've got the right idea that `showErrorAnnouncer` is a prop that gets passed to your component. The one thing you're missing is how your component gets that prop. Reviewing how HOCs work, the [code for our error HOC](https://github.com/cyverse-de/sonora/blob/master/src/components/utils/error/withErrorAnnouncer.js), and how other components are using that `withErrorAnnouncer` code should help.

flynn1999: I think it should be correct now. I tested it by substituting the `onError` part with `onSuccess` and commenting out the original `onSuccess` and clicking Mark All As Seen in dev.

image

ash3rz: Excellent!  

This is one way to test in dev sort of on the fly.  Another way you could test that might be faster and wouldn't require you to switch code back and forth is by using Storybook.  

If you recall, you're able to mock API calls in Storybook using [axios-mock-adapter](https://github.com/ctimmerm/axios-mock-adapter).  So far, we've mocked our API calls to always be successful by returning HTTP 200 and then providing a response with mock data.  

You can also test out your component by returning an unsuccessful response back.  There are examples in that mock axios repo and [here's](https://github.com/cyverse-de/sonora/blob/master/stories/notifications/Listing.stories.js) an example in our repo that Paul did for his portion of Notifications.  In this case, this story wanted an error to always show up the first time you try to mark a notification as seen, but then succeed after that, so when you run the story you can see both branches of pass/fail logic.

ash3rz: Also, minor thing - since the text on the button says `Mark All as Read`, I would change the language here in the error message to match that: `Unable to mark notifications as read`

ash3rz (approved):
LGTM :shipit: 
Nice work! 🎉

sri2k1us (approved):
LGTM :shipit: 🎉

PR #234, opened 2020-11-20
Fix errors for logged out users viewing Data

ash3rz:
NOTE: Depends on change from https://github.com/cyverse-de/terrain/pull/180

As discussed in Slack, we're moving towards not hitting the data roots endpoint anymore and having Sonora create the roots for the anonymous user.

Since the QA env is down, I went ahead and tested this directly against prod.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -17,6 +17,7 @@ module.exports = withBundleAnalyzer({
         ANALYSIS_SUPPORT_USER: config.get("analysis.support.user"),
         ANALYSIS_SUPPORT_SOURCE_ID: config.get("analysis.support.source_id"),
         IRODS_HOME_PATH: config.get("irods.home_path"),
+        IRODS_TRASH_PATH: config.get("irods.trash_path"),
         SESSION_POLL_INTERVAL_MS: config.has("sessions.poll_interval_ms")
             ? config.get("sessions.poll_interval_ms")
             : 5000,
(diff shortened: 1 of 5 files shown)


next.config.js
```diff
@@ -17,6 +17,7 @@ module.exports = withBundleAnalyzer({
         ANALYSIS_SUPPORT_USER: config.get("analysis.support.user"),
         ANALYSIS_SUPPORT_SOURCE_ID: config.get("analysis.support.source_id"),
         IRODS_HOME_PATH: config.get("irods.home_path"),
+        IRODS_TRASH_PATH: config.get("irods.trash_path"),
```
ash3rz, line 20: This trash path config value was already made available server side, but I never added it client side, so this includes those changes.

src/serviceFacades/filesystem.js
```diff
@@ -88,10 +89,36 @@ export const getPagedListing = (
...
+                      label: constants.TRASH,
+                  },
+              ],
+              "base-paths": {
+                  user_home_path: `${homePath}/${constants.ANONYMOUS_USER}`,
+                  user_trash_path: `${trashPath}/${constants.ANONYMOUS_USER}`,
+                  base_trash_path: trashPath,
+              },
+          });
 };
```
sri2k1us, line 123: 👍

sri2k1us (approved):
LGTM :shipit: Details still throw an error when accessed as logged out user. Is that something that still needs work ?

psarando (approved):
LGTM :shipit:

ash3rz:
Thanks for the reviews!
And yes - I forgot about details, the `/stats` endpoint.  Those changes will be coming in a terrain PR shortly.

PR #235, opened 2020-11-20
Fix card content overflow and analysis status.

sri2k1us:
- Fix overflowing card content.
- Display analysis status as secondary text. (Long analysis name was truncating analysis status).

### DIFF
--- a/src/components/dashboard/dashboardItem/ItemBase.js
+++ b/src/components/dashboard/dashboardItem/ItemBase.js
@@ -127,8 +127,7 @@ const DashboardItem = ({ item }) => {
                 <Typography
                     color="textSecondary"
                     variant="body2"
-                    component="p"
-                    style={{ overflow: "ellipsis", height: "4.2em" }}
+                    style={{ overflow: "ellipsis", height: "4.0em" }}
                 >
                     {description || t("noDescriptionProvided")}
                 </Typography>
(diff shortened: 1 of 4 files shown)


src/components/dashboard/dashboardItem/ItemBase.js
```diff
@@ -123,46 +124,17 @@ const DashboardItem = ({ item }) => {
                     root: classes.root,
                 }}
             >
-                <Typography color="textSecondary" variant="body2" component="p">
+                <Typography
+                    color="textSecondary"
+                    variant="body2"
+                    component="p"
+                    style={{ overflow: "ellipsis", height: "4.2em" }}
```
ash3rz, line 131: I think this height can be removed.  Otherwise, it looks like the action buttons start to get cramped as the card resizes:
image

ash3rz (approved):
Found a small bug with a `height` style, but otherwise LGTM :shipit:

src/components/dashboard/dashboardItem/ItemBase.js
sri2k1us: Removing height will cause the text to `overflow`. It is required for `ellipsis` to work properly. I have adjusted the `height` and `DESC_MAX_LENGTH` to mitigate this problem. From my testing this happens only at a particular screen size and it should be fixed now.

sri2k1us:
Thanks for the review.

PR #236, opened 2020-11-23
Allow logged out users to view public data details

ash3rz:
NOTE: Requires https://github.com/cyverse-de/terrain/pull/182 to be merged first

Now users should be able to view the data details drawer within Community Data without logging in.  Since we don't have/support public tags, I effectively disabled that feature if no user is logged in and updated the error message to indicate that you need to be signed in to use it.

I also hid the `read`, `write`, and `own` labels in the Permissions panel since those really should only show up for users with `own` permission anyway. Showing `own (0), write (0), and read (0)` without this change was a bit misleading otherwise.

image

image

### DIFF
--- a/src/components/data/details/PermissionsPanel.js
+++ b/src/components/data/details/PermissionsPanel.js
@@ -267,7 +267,7 @@ function PermissionsTabPanel(props) {
             <Typography id={build(baseId, ids.SELF_PERMISSION)}>
                 {t("selfPermission", { permission: selfPermission })}
             </Typography>
-            {fetchResourcePermissionsKey && (
+            {selfPermission === Permissions.OWN && (
                 <List
                     className={classes.permissionsList}
                     id={build(baseId, ids.PERMISSION_LIST)}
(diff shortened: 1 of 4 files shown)


sri2k1us (approved):
LGTM :shipit:

src/components/data/details/PermissionsPanel.js
```diff
@@ -267,7 +267,7 @@ function PermissionsTabPanel(props) {
             <Typography id={build(baseId, ids.SELF_PERMISSION)}>
                 {t("selfPermission", { permission: selfPermission })}
             </Typography>
-            {fetchResourcePermissionsKey && (
+            {selfPermission && selfPermission === Permissions.OWN && (
```
psarando, line 270:
```suggestion
            {selfPermission === Permissions.OWN && (
```
This should already result in `false` if `selfPermission` is falsey.

psarando (approved):
LGTM 👍 

Just had a minor style suggestion.

ash3rz:
Thanks for the reviews!

PR #238, opened 2020-11-25
CORE-1165 Import Metadata listing and AVU forms

psarando:
This PR will import the Metadata listing and AVU forms from the `cyverse-de/ui` repo, but with some modifications for Sonora.

The metadata listing view is now intended for full-page display and no longer contained in a dialog, so the form is reset and a success announcer is displayed when metadata is saved successfully.

Edit Metadata listing

Note that the `Save` button has been renamed `Apply`, which displays as disabled when there are no changes to the metadata, when the submission is in progress, and after a successful save (so only enabled when there are changes not yet saved).

All other toolbar actions have been moved into the dot-menu, which is more mobile-friendly.

Also the add AVU `+` icon button is no longer a FAB.

---

Dot Menu:
User Metadata Dot Menu

If the user is not logged in and they select the "Save Metadata to File" action, then a "Sign In" dialog is displayed (same as the dialog that displays when an endpoint returns a 401 error), otherwise the Save-As dialog is displayed.

---

iRODS metadata listing:
iRODS Metadata Listing

---

iRODS Metadata Dot Menu:
iRODS Metadata Dot Menu

Note that the "Import" menu item is hidden until at least 1 row in the iRODS listing is selected.

---

The AVU edit form is still in a Dialog, but without an AppBar heading.

AVU Edit Dialog

---

Read-only metadata listing:

Read-only Metadata Listing view

---

Read-only AVU dialog:

Read-only AVU Dialog

---

Mobile view listing:
Edit Metadata Listing - mobile view

---

Mobile view AVU Dialog (not sure why, but Firefox Responsive Dev Mode still displays a horizontal scrollbar for the listing under the dialog when emulating this mobile view):
AVU Edit Dialog - mobile view

### TODO

* The Metadata Template viewer/editor still needs to be imported, so the view-in-template action will display a "Coming Soon" message for now.

PR #241, opened 2020-12-01
CORE-1175 Pathlist automation

sri2k1us:
-Menu:
image

- Path list automation form:
image

- Form validation:
image

- Service Error handling - no matching paths:
image

- Service Error handling - duplicate file name:
image

-iPhone 5:
image


src/components/data/SaveAsDialog.js
```diff
@@ -46,7 +47,13 @@ function SaveAsDialog(props) {
...
         setSaveNewFileError,
     } = props;
-    const [dest, setDest] = useState(path);
+
+    const [userProfile] = useUserProfile();
+    const [config] = useConfig();
+    const username = userProfile?.id;
+    const irodsHomePath = config?.irods?.home_path;
+
+    const [dest, setDest] = useState(path || `${irodsHomePath}/${username}`);
```
sri2k1us, line 56: I saw some of these changes in #238 as well.

src/components/data/SaveAsField.js
```diff
@@ -0,0 +1,126 @@
...
+                    {...props}
+                />
+            </Grid>
+        </Grid>
+    );
+};
+
+export { SaveAsButton };
+
+export default withErrorAnnouncer(SaveAsField);
```
sri2k1us, line 126: I wrote this component so that I could include the `Save As` Dialog as a form field. But am not entirely happy about it since this popups another dialog on top of `Path List Automation` dialog. May be this could be eliminated if `SaveAsDialog` could be split into a wrapping `Dialog` and `SaveAsField` 🤔  ?

src/components/data/viewers/Toolbar.js
```diff
@@ -125,6 +134,20 @@ function ViewerToolbar(props) {
...
+                    startIcon={<ArrowBack fontSize="small" />}
+                    onClick={() => router.back()}
+                >
+                    <Hidden xsDown>{t("back")}</Hidden>
+                </Button>
+                <Divider
+                    orientation="vertical"
+                    flexItem
+                    className={classes.separator}
+                />
```
sri2k1us, line 150: While testing, the absence of `Back` bothered me so much, I added one right away :)

sri2k1us:
image

Long menu text causes text to cut-off. Suggestions welcome to rename these menu items.

src/components/data/PathListAutomation.js
```diff
@@ -0,0 +1,405 @@
...
+            onSuccess,
+            onError,
+        });
+    };
+
+    const onSwitchChange = (setFieldValue, fieldName) => (event, checked) => {
+        setFieldValue(fieldName, checked);
+    };
+
+    const FormSwitch = ({
```
ash3rz, line 172: I think `FormCheckBox` from `ui-lib` would also work well here if you wanted less code to manage in this component.  Otherwise, this might be good to add to `ui-lib`.

src/components/data/PathListAutomation.js
```diff
@@ -0,0 +1,405 @@
...
+
+    const validate = (values) => {
+        const { selectedPaths, dest } = values;
+        const errors = {};
+        if (!selectedPaths || selectedPaths.length === 0) {
+            errors.selectedPaths = "Required";
+        }
+
+        if (!dest) {
+            errors.dest = "Required";
```
ash3rz, line 192: This `Required` text should be i18n

src/components/data/PathListAutomation.js
```diff
@@ -0,0 +1,405 @@
...
+            paths: selectedPaths,
+            dest,
+            pattern,
+            foldersOnly: foldersOnly || false,
+            recursive: true,
+            requestedInfoType,
+            selectedInfoTypes,
+        };
+        const onSuccess = () => {
+            actions.setSubmitting(false);
```
ash3rz, line 153: 🤔 It's been a while since I've worked with Formik, but I don't recall ever having to manually manage the submitting state.  I might be missing something though.

ash3rz (approved):
LGTM - I just had some minor comments and some ideas thrown out there. :shipit:

psarando (changes requested):
Looks good so far. You've tackled a fairly complex Formik form 🎉 

I think there are a couple of changes that should be considered before merging, though 😣

src/components/data/PathListAutomation.js
ash3rz: Ah ha!  I thought I had seen this somewhere but only checked `ui-lib`.  Good catch!

src/components/data/SaveAsField.js
```diff
@@ -0,0 +1,126 @@
...
+                    startingPath={startingPath}
+                    multiSelect={false}
+                    name={name}
+                    onConfirm={(path) => {
+                        setFieldValue(name, path);
+                    }}
+                />
+            </InputAdornment>
+        ),
+    };
```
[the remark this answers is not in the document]
sri2k1us: While this is a look alike component of `InputSelector`, I wouldn't go too far into having a single component for Input selection and Save as field as they are fundamentally for different purpose( InputSelector collects an existing path while SaveAsField get a new file name from the user). I don't anticipate this buying us much at this much other than saving few lines of code trading for few hours of work.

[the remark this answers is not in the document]
psarando: 👌 I agree the buttons and the dialogs they open are fundamentally different, though the form field that wraps those buttons are exactly the same. In a previous comment you mentioned refactoring the `SaveAsField`, so I thought the `InputSelector` could be refactored at the same time, but it's a minor improvement, so I'm fine with skipping that for now.

I think at least the Grids wrapping both form fields could be removed now, since they look like unneeded cruft at this point.

ash3rz (commented):
I think I caught all the changes and only found a couple small things 👍

sri2k1us:
> I think I caught all the changes and only found a couple small things 👍

Ohh I havent finished it up. I wanted to make  sure if this is the right direction to take.👍

psarando:
> Ohh I havent finished it up. I wanted to make sure if this is the right direction to take.👍

I'm liking the direction these refactors are going 👍

src/components/data/SaveAsDialog.js
```diff
@@ -55,26 +47,25 @@ function SaveAsDialog(props) {
...
 
+    const textFieldProps = {};
+    if (saveFileError) {
+        textFieldProps.error = true;
+        textFieldProps.helperText = saveFileError;
+    }
+
     return (
         <Formik
+            enableReinitialize
```
sri2k1us: Took some of the logic from #238.

package.json
```diff
@@ -70,7 +70,7 @@
         "express-session": "^1.17.1",
         "express-winston": "^4.0.2",
         "express-ws": "^4.0.0",
-        "formik": "^1.5.8",
+        "formik": "^2.2.5",
```
sri2k1us, line 73: I update formik to 2.x and it seems to work without any code changes from my testing. But would appreciate more testing from reviewers.

src/components/data/SaveAsField.js
```diff
@@ -0,0 +1,79 @@
...
+                        )}
+                        color="inherit"
+                        size={20}
+                    />
+                )}
+            </>
+        ),
+    };
+
+    if (value && !required) {
```
psarando, line 55: This should probably also check for `loading`, and not override the `CircularProgress`?
```suggestion
    if (value && !required && !loading) {
```

src/components/data/PathListAutomation.js
```diff
@@ -0,0 +1,412 @@
...
+                                )}
+                                <Field
+                                    id={buildId(
+                                        baseId,
+                                        ids.PATH_LIST_AUTO_INPUTS
+                                    )}
+                                    name="selectedPaths"
+                                    required={true}
+                                    component={MultiInputSelector}
+                                    height="20vh"
```
psarando, line 247: I noticed in the Chromium story that only 1 row at a time will display in the resource selector (at least in that view). So if the user selects more than 1 file or folder to process, they have to scroll to review all their selections. Maybe this could use an increased `height` to display more than 1 row at a time (maybe only an issue in smaller views):
https://www.chromatic.com/component?appId=5ebad5c6b991d60022b1e72e&name=Data+%2F+Path+List+Automation&buildNumber=1113

psarando (approved):
LGTM 🎉 

I had a couple of questions, and found a few minor items that could be cleaned up ✨ but I didn't see any blockers.

Also, I noticed in the Chromium story that the `info-type` selector is initially very flat, then once the info types load, it takes up more vertical space and pushes the input selector and cancel/done buttons down. Maybe a loading skeleton for the `info-type` selector could help keep things from moving around too much:
https://www.chromatic.com/component?appId=5ebad5c6b991d60022b1e72e&name=Data+%2F+Path+List+Automation&buildNumber=1113

src/components/data/PathListAutomation.js
psarando: Even in local dev mode on my laptop, I'm only seeing 1.5 rows:
resource selector sizing in dev mode

src/components/data/PathListAutomation.js
```diff
@@ -0,0 +1,412 @@
...
+            errors.dest = i18nCommon("required");
+        }
+        return errors;
+    };
+
+    return (
+        <Formik
+            onSubmit={handlePathListCreation}
+            validate={validate}
+            initialValues={{ selectedPaths: [], dest: startingPath }}
```
psarando, line 195: While testing locally, I got a warning in the console when entering a value for the `pattern` field: `A component is changing an uncontrolled input of type text to be controlled`.

I think this is because there's no initial value set for `pattern` in this `initialValues` object.

According to the Formik docs, every field should have an initial value.
Based on what's getting used in `handlePathListCreation`, there should also be initial values for `fileName`, `pattern`, and `foldersOnly`.
```suggestion
            initialValues={{ selectedPaths: [], dest: startingPath, fileName: "", pattern: "", foldersOnly: false }}
```

src/components/data/PathListAutomation.js
```diff
@@ -0,0 +1,405 @@
...
+                            </Grid>
+                            <Grid item>
+                                <Button
+                                    id={buildId(
+                                        baseId,
+                                        ids.PATH_LIST_AUTO_DONE_BTN
+                                    )}
+                                    color="primary"
+                                    type="submit"
+                                    onClick={handleSubmit}
```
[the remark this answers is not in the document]
psarando: In local dev testing, I was able to double-click the `Done` button and easily make 6 submissions in a row before the dialog closed and the page redirected. In this case it looked like only the first submission succeeded (I think the rest returned file-exists errors), but if the data store is slow, I'm not sure what kind of race-condition issues this could cause.

package.json
psarando: I tested out app launch (which uses formik the most at the moment) in storybook and in dev mode, and I didn't catch any issues 👍 

I also checked the breaking changes from v1 -> v2, and it doesn't look like any of those will affect us 🎉

src/components/data/PathListAutomation.js
sri2k1us: GTK. I always had my doubts about initialValues.

src/components/data/PathListAutomation.js
sri2k1us: The reason I had to reduce height for Multi input selector and Info type selector is that, with increased height the dialog get into scrolling. When there is an error, it is displayed at the top of screen. It is easy to miss the error since hidden as users would scrolled down to click `Done` (happened even to me twice). I can look into it bit more.

sri2k1us: I made some adjustments to the `MultiInputSelector` table
image

Also am displaying errors (if any) at the top and at the bottom to address the issue described above:
image

sri2k1us:
Thanks for the reviews.

src/components/data/PathListAutomation.js
psarando: I also noticed in local dev on my laptop that this dialog would scroll, even with the short multi-input selector, and it also took me a while to find an error that was scrolled off the top.

So adding the error message by the cancel/done buttons is much better ✨

src/components/data/PathListAutomation.js
```diff
@@ -391,7 +399,6 @@ export default function PathListAutomation(props) {
                                     )}
                                     color="primary"
                                     type="submit"
-                                    onClick={handleSubmit}
```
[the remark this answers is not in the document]
psarando: So I checked the Formik docs, and it turns out that their `Form` automagically handles submissions from this `submit` button 🎉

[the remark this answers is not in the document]
sri2k1us: Yes. Sorry forgot to respond to your comment.

src/components/data/PathListAutomation.js
```diff
@@ -0,0 +1,438 @@
...
+            requestedInfoType,
+            selectedInfoTypes,
+        };
+
+        const onError = (err_code) => {
+            if (err_code === ERROR_CODES.ERR_EXISTS) {
+                actions.setFieldError("dest", t("pathListExistError"));
+            }
+        };
+        if (status !== constants.LOADING) {
```
sri2k1us, line 165: Also, we can prevent dupe submission without disabling buttons by placing a guard just before  calling the mutation.

PR #244, opened 2020-12-03
CORE-1158 Allow public apps to be browsable by logged out users

ash3rz:
NOTE: Requires https://github.com/cyverse-de/terrain/pull/186 to be merged first and also requires the anonymous user to have a workspace created

This also disables analysis search for logged out users so they don't get an error message when searching public data and public apps under the `All` search.

### DIFF
--- a/src/components/apps/launch/index.js
+++ b/src/components/apps/launch/index.js
@@ -89,7 +89,7 @@ export default ({ app, launchError, loading }) => {
             defaultMaxMemory={defaultMaxMemory}
             defaultMaxDiskSpace={defaultMaxDiskSpace}
             app={app}
-            appError={launchError || submissionError}
+            appError={submissionError}
             loading={loading}
             submitAnalysis={(submission, onSuccess, onError) => {
                 setSubmissionError(null);
--- a/src/components/apps/launch/index.js
+++ b/src/components/apps/launch/index.js
@@ -20,6 +20,7 @@ import { submitAnalysis } from "serviceFacades/analyses";
 import { addQuickLaunch } from "serviceFacades/quickLaunches";
 
 import AppLaunchWizard from "./AppLaunchWizard";
+import WrappedErrorHandler from "../../utils/error/WrappedErrorHandler";
 
 export default ({ app, launchError, loading }) => {
     const [submissionError, setSubmissionError] = React.useState(null);
@@ -71,9 +72,17 @@ export default ({ app, launchError, loading }) => {
     const defaultMaxMemory = config?.tools?.private.max_memory_limit;
     const defaultMaxDiskSpace = config?.tools?.private.max_disk_limit;
 
+    const baseId = "apps";
+
+    if (launchError) {
+        return (
+            <WrappedErrorHandler errorObject={launchError} baseId={baseId} />
+        );
+    }
+
     return (
         <AppLaunchWizard
-            baseId="apps"
+            baseId={baseId}
             notify={notify}
             defaultOutputDir={defaultOutputDir}
             defaultMaxCPUCores={defaultMaxCPUCores}
--- a/src/components/apps/listing/Listing.js
+++ b/src/components/apps/listing/Listing.js
@@ -112,7 +112,16 @@ function Listing({
         ],
         queryFn: getAppsInCategory,
         config: {
-            enabled: category?.system_id && category?.id,
+            enabled:
+                // Disable the query if the category ID is fake and the user is
+                // logged in.  The Navigation component should update the ID to
+                // the real ID.
+                category?.system_id &&
+                category?.id &&
+                (!userProfile?.id ||
+                    ![constants.APPS_UNDER_DEV, constants.FAV_APPS].includes(
+                        category?.id
+                    )),
             onSuccess: setData,
         },
     });
--- a/src/components/apps/listing/Listing.js
+++ b/src/components/apps/listing/Listing.js
@@ -40,6 +40,7 @@ import Sharing from "components/sharing";
 import { formatSharedApps } from "components/sharing/util";
 import AppDoc from "components/apps/details/AppDoc";
 import QuickLaunchDialog from "../quickLaunch/QuickLaunchDialog";
+import { useUserProfile } from "contexts/userProfile";
 
 function Listing({
     baseId,
@@ -57,6 +58,7 @@ function Listing({
 }) {
     const { t } = useTranslation(["apps", "common"]);
     const [isGridView, setGridView] = useState(false);
+    const [userProfile] = useUserProfile();
 
     const [order, setOrder] = useState(selectedOrder);
     const [orderBy, setOrderBy] = useState(selectedOrderBy);
@@ -217,6 +219,7 @@ function Listing({
                     appTypeFilter,
                     page,
                     categoryId,
+                    userId: userProfile?.id,
                 },
             ]);
             setAppsInCategoryQueryEnabled(true);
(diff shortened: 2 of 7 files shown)


src/components/apps/listing/Listing.js
```diff
@@ -109,28 +97,61 @@ function Listing({
         isFetching: appInCategoryStatus,
         error: appsInCategoryError,
     } = useQuery({
-        queryKey: appsInCategoryKey,
+        queryKey: [
+            APPS_IN_CATEGORY_QUERY_KEY,
```
ash3rz, line 101: I left a note in my commit message, but I switched to using this method of query keys instead of the state and effect combo because I found during testing that having the key switch to null and enabled set to false doesn't seem to clear out the error, in this case `appsInCategoryError`.  So, for example, I could select `Browse All Apps` and see the listing, select any other category and see the `Sign In` page, switch back to `Browse All Apps` and I would still see the `Sign In` page instead of the listing.

I also found this easier to read because I didn't have to jump up and down between state and effect functions to see when these queries were supposed to be active.

src/components/apps/listing/Listing.js
```diff
@@ -109,28 +97,61 @@ function Listing({
...
+            enabled:
+                // Disable the query if the category ID is fake and the user is
+                // logged in.  The Navigation component should update the ID to
+                // the real ID.
+                category?.system_id &&
+                category?.id &&
+                (!userProfile?.id ||
+                    ![constants.APPS_UNDER_DEV, constants.FAV_APPS].includes(
+                        category?.id
+                    )),
```
ash3rz, line 124: Adding the check if the user is logged in and also checking the category ID is the only functional change to whether this query is enabled.
This is a bit hacky, but I'm disabling this query when the user is logged in and has one of these "fake" category UUIDs, carried over from the anonymous user, in order for the `Navigation` component to have an opportunity to update the currently selected category with the real category UUID for that user.

src/components/apps/toolbar/AppNavigation.js
```diff
@@ -124,12 +130,32 @@ function AppNavigation(props) {
...
+            // categories given to the anonymous user selected with a fake
+            // UUID
+            if (
+                [constants.FAV_APPS, constants.APPS_UNDER_DEV].includes(
+                    selectedCategory?.id
+                )
+            ) {
+                handleCategoryChange(
+                    categoryList.find(
+                        (category) => category.name === selectedCategory.name
```
ash3rz, line 143: Here's where the `AppNavigation` component updates the currently selected category with the real category ID instead of the fake category ID.

src/serviceFacades/apps.js
```diff
@@ -38,23 +39,74 @@ function getAppById(key, { systemId, appId }) {
...
+                      name: "Workspace",
+                      categories: [
+                          {
+                              system_id: "de",
+                              id: constants.APPS_UNDER_DEV,
+                              name: constants.APPS_UNDER_DEV,
+                          },
+                          {
+                              system_id: "de",
+                              id: constants.FAV_APPS,
```
ash3rz, line 61: Here's an example of a fake category ID.  For `Favorite Apps` and `Apps Under Development` the category IDs are unique for each user and I'm setting a fake ID for the anonymous user.

sri2k1us (approved):
LGTM. I found a minor issue with quick launches. When logged out and I select `Quick Launch` from dot menu, I don't get sign-in prompt. Similarly, when I select `Create Quick Launch` from `Details`, nothing happens 🤔 . This can be addressed in a separate PR though.

src/components/apps/launch/index.js
```diff
@@ -71,9 +72,17 @@ export default ({ app, launchError, loading }) => {
     const defaultMaxMemory = config?.tools?.private.max_memory_limit;
     const defaultMaxDiskSpace = config?.tools?.private.max_disk_limit;
 
+    const baseId = "apps";
+
+    if (launchError) {
+        return (
+            <WrappedErrorHandler errorObject={launchError} baseId={baseId} />
+        );
+    }
```
psarando, line 81: If we're returning this handler now for a `launchError`, then I guess it could be removed from the `appError` prop in the `AppLaunchWizard` below (it was being displayed in an `ErrorTypography` + `DEErrorDialog` by the `AppInfo` component before).

src/components/apps/utils.js
```diff
@@ -32,8 +32,11 @@ export const getListingPath = (
     rowsPerPage,
     filter,
     category
-) =>
-    `/${NavigationConstants.APPS}?selectedOrder=${order}&selectedOrderBy=${orderBy}&selectedPage=${page}&selectedRowsPerPage=${rowsPerPage}&selectedFilter=${filter}&selectedCategory=${category}`;
+) => {
+    const encodedFilter = encodeURIComponent(filter);
+    const encodedCategory = encodeURIComponent(category);
+    return `/${NavigationConstants.APPS}?selectedOrder=${order}&selectedOrderBy=${orderBy}&selectedPage=${page}&selectedRowsPerPage=${rowsPerPage}&selectedFilter=${encodedFilter}&selectedCategory=${encodedCategory}`;
```
psarando, line 38: 👍

src/serviceFacades/apps.js
psarando: "setting a fake ID for the anonymous user" might be worth adding as a comment to this source somewhere.

psarando (approved):
LGTM 👍 

I just had a couple of minor suggestions ✨

ash3rz:
> LGTM. I found a minor issue with quick launches. When logged out and I select `Quick Launch` from dot menu, I don't get sign-in prompt. Similarly, when I select `Create Quick Launch` from `Details`, nothing happens 🤔 . This can be addressed in a separate PR though.

I checked this real quick in QA and it turns out nothing happens when you're signed in either when you click on `Create Quick Launch` 😄 I think maybe that feature hasn't been fully implemented? It seems to be a link with text, but no action tied to it: https://github.com/cyverse-de/sonora/blob/364a263ce957dd5678bc845f938529d000b2bf33/src/components/apps/quickLaunch/QuickLaunchListing.js#L245-L251

As for the row dot menu, I think it's working because there is the concept of public quick launches, so I made that available for logged out users.  For example, if you check out `3.1 dev GEA Leptin RNA Seq Lab` (it's the 2nd app listed under `Browse All Apps` for me) that app has 2 public quick launches.  Clicking on those quick launch links will then prompt you to sign in since it takes you to the app launch screen.  Let me know if I missed something though.

ash3rz:
Thanks for the reviews!

sri2k1us:
> > LGTM. I found a minor issue with quick launches. When logged out and I select `Quick Launch` from dot menu, I don't get sign-in prompt. Similarly, when I select `Create Quick Launch` from `Details`, nothing happens 🤔 . This can be addressed in a separate PR though.
> 
> I checked this real quick in QA and it turns out nothing happens when you're signed in either when you click on `Create Quick Launch` 😄 I think maybe that feature hasn't been fully implemented? It seems to be a link with text, but no action tied to it:
> 
> https://github.com/cyverse-de/sonora/blob/364a263ce957dd5678bc845f938529d000b2bf33/src/components/apps/quickLaunch/QuickLaunchListing.js#L245-L251
> 
> As for the row dot menu, I think it's working because there is the concept of public quick launches, so I made that available for logged out users. For example, if you check out `3.1 dev GEA Leptin RNA Seq Lab` (it's the 2nd app listed under `Browse All Apps` for me) that app has 2 public quick launches. Clicking on those quick launch links will then prompt you to sign in since it takes you to the app launch screen. Let me know if I missed something though.

Yes. It does seem to work 👍🏽 . Not sure what happened when I was testing locally.

PR #245, opened 2020-12-04
Load showdown.js dynamically, only when app documentation is actually opened

ianmcorvidae:
Small PR:

This is a somewhat large library that's only relevant in one very specific situation, when app documentation is loaded. This makes it so it's only loaded when needed.

### DIFF
--- a/src/components/apps/details/AppDoc.js
+++ b/src/components/apps/details/AppDoc.js
@@ -6,7 +6,6 @@
 import React, { useEffect, useState } from "react";
 import { useQuery, useMutation } from "react-query";
 import sanitizeHtml from "sanitize-html";
-import showdown from "showdown";
 import { useTranslation } from "i18n";
 
 import { build as buildDebugId } from "@cyverse-de/ui-lib";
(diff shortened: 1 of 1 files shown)


src/components/apps/details/AppDoc.js
```diff
@@ -77,13 +76,22 @@ function Documentation(props) {
         baseId,
     } = props;
 
+    const [htmlDocumentation, setHtmlDocumentation] = useState("");
```
ianmcorvidae, line 79: This starting state is pretty much my main question -- should we have something inforrmative here, or not worry about it?

src/components/apps/details/AppDoc.js
```diff
@@ -77,13 +76,22 @@ function Documentation(props) {
...
+            const showdown = (await import("showdown")).default;
+            const converter = new showdown.Converter();
+            converter.setFlavor("github");
+            setHtmlDocumentation(
+                sanitizeHtml(converter.makeHtml(documentation))
+            );
+        };
+
+        regenerateMarkdown(documentation);
+    }, [documentation]);
```
sri2k1us, line 94: 👍

src/components/apps/details/AppDoc.js
sri2k1us: I wouldn't worry too much about it give the doc end-point responds really fast.

sri2k1us (approved):
LGTM :shipit:

ash3rz (approved):
LGTM :shipit:

ianmcorvidae:
Thanks for the reviews! I'll go ahead and merge this one and then go about fixing the other one later today.

PR #246, opened 2020-12-04
CyVerseAppBar optimization

ianmcorvidae:
Because I constantly find myself trying to get our first-load JS down (nextjs wants it to ideally be below 130k -- this PR takes it from 284 to 251), I spent some time working on the shared-by-everything CyVerseAppBar code. I'll leave some comments in the code to highlight particular changes.


src/components/layout/CyVerseAppBar.js
```diff
@@ -61,6 +63,9 @@ import SupervisorAccountIcon from "@material-ui/icons/SupervisorAccount";
 import LabelImportantIcon from "@material-ui/icons/LabelImportant";
 import UserMenu from "./UserMenu";
 
+// hidden in xsDown
+const GlobalSearchField = dynamic(() => import("../search/GlobalSearchField"));
```
ianmcorvidae, line 67: Just a few kb, but this isn't used on mobile at all.

src/components/layout/CyVerseAppBar.js
```diff
@@ -211,6 +216,52 @@ const BagMenu = () => {
     return <Bag menuIconClass={classes.menuIcon} />;
 };
 
+const DrawerItem = (props) => {
```
ianmcorvidae, line 219: Splitting this out doesn't help the load size, but it did help my sanity a bit. I hope others agree this is nicer. I didn't do anything with the admin drawer yet, because it uses a different pattern to some degree.

src/components/layout/CyVerseAppBar.js
```diff
@@ -352,10 +404,7 @@ function CyverseAppBar(props) {
     const onUserMenuClose = () => {
         setAnchorEl(null);
     };
-    const handleSearchClick = (event) => {
-        router.push("/" + NavigationConstants.SEARCH);
-        toggleDrawer(false);
-    };
```
ianmcorvidae, line 358: Rendered irrelevant by the DrawerItem code -- _unless_ I've misunderstood something and there's a reason we specifically need the `router.push` first here, vs. the other ones that all have `toggleDrawer` first.

src/components/layout/CyVerseAppBar.js
```diff
@@ -603,7 +576,7 @@ function CyverseAppBar(props) {
                 })}
             >
                 <Toolbar>
-                    <Hidden only={["sm", "md", "lg", "xl"]}>
+                    <Hidden smUp>
```
ianmcorvidae, line 579: This and similar is so we're using the same names for things throughout -- in this file, we only use `smUp` and `xsDown`, but we were writing those four ways (those two, plus the versions using the `only` prop). So, just minor consistency, but anyway.

src/components/layout/CyVerseAppBar.js
```diff
@@ -642,104 +613,81 @@ function CyverseAppBar(props) {
...
+                </Hidden>
+                <Divider />
+                {drawerItems}
+                {(!isXsDown || open) && adminUser && (
+                    <>
+                        <Divider />
+                        {adminDrawerItems}
+                    </>
+                )}
+            </Drawer>
```
ianmcorvidae, line 690: This whole thing is basically reorganizing: in both setups, we have a drawer with various props, some conditional stuff at the top depending on the screen size, then the main drawer items, then the admin drawer (with slightly different conditions due to the different drawer types). This helps the code size some since we're sharing more, and I think it also makes the shared vs. differing code more obvious -- and if I understand React properly, this should mean less DOM manipulation too. As an additional bonus, having more things outside of `Hidden` components means that more can be server-side rendered.

src/components/sharing/constants.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * These are the keys to each permission-lister response object and correspond
+ * to each type of resource that can be shared in Sonora
+ */
+export const TYPE = {
+    DATA: "paths",
+    APPS: "apps",
+    ANALYSES: "analyses",
+    TOOLS: "tools",
+};
```
ianmcorvidae, line 10: I separated these out into another file because this is the only part of the sharing utils namespace that the Bag component was using, but since it was in that file, it brought in the whole thing. Since Bag is also used by CyVerseAppBar, this saved a few kB.

src/components/layout/CyVerseAppBar.js
```diff
@@ -642,104 +613,81 @@ function CyverseAppBar(props) {
...
+                                onManageAccountClick={() =>
+                                    window.open(
+                                        constants.CYVERSE_USER_PORTAL,
+                                        "_blank"
+                                    )
+                                }
+                            />
+                        ) : (
+                            accountAvatar
+                        )}
```
ianmcorvidae, line 679: This makes it so the mobile view works when logged-out, by showing a plain account avatar. Open to suggestions if there's something better here, but it seems not too bad to me at least.

sri2k1us (commented):
Looks reasonable to me but when I ran this PR locally I found 
image

Is this change expected ? If so I would just show the top level `Admin` Icon and leave out the the sub-items arrows. We could simply link it to a `/admin` page which will  list / link to all admin sub pages.

Also I just merged #230 so there I anticipate some nasty merge conflicts 😖.

ianmcorvidae:
That's an odd bug, I'll look into it (and the merge conflicts, of course)

src/components/layout/CyVerseAppBar.js
```diff
@@ -642,104 +613,81 @@ function CyverseAppBar(props) {
...
-                            <Divider />
-                            {adminDrawerItems}
-                        </>
-                    )}
-                </Drawer>
-            </Hidden>
+                </Hidden>
+                <Divider />
+                {drawerItems}
+                {(!isXsDown || open) && adminUser && (
```
ianmcorvidae, line 684: I realize this line is probably why those admin items are showing up where they shouldn't, heh.

ianmcorvidae:
I've updated the PR with both merging in master and fixing things, and fixing the backwards conditional I had. I think things are working/good. The first load JS shared by all is now 282kB, compared to master at 315kB. I'm going to see if I can't bring that back down into the range I had gotten before ~250kB, probably by lazy-loading the react-joyride code.

src/components/layout/CyVerseAppBar.js
```diff
@@ -69,6 +68,9 @@ import UserMenu from "./UserMenu";
 // hidden in xsDown
 const GlobalSearchField = dynamic(() => import("../search/GlobalSearchField"));
 
+// only new users
+const Joyride = dynamic(() => import("react-joyride"));
```
ianmcorvidae, line 72: Looks like next is smart enough to understand importing react-joyride twice, both dynamically, and only making one chunk/loading once. So that's cool.

ianmcorvidae:
I'm about to head out to an appointment, so I'll merge this after that unless it's too late in the day for me to feel comfortable doing so (review still welcome, as always)

ash3rz (approved):
Really nice refactors ✨ 
LGTM :shipit: Checked it out locally as well and didn't spot anything

src/components/layout/CyVerseAppBar.js
```diff
@@ -432,137 +501,70 @@ function CyverseAppBar(props) {
...
-                                fontSize="large"
-                            />
-                        </ListItemIcon>
-                        {open && <ListItemText>{t("search")}</ListItemText>}
-                    </ListItem>
-                </Tooltip>
+            <DrawerItem
+                title={t("dashboard")}
+                id={ids.DASHBOARD_MI}
+                image={"/dashboard_selected.png"}
```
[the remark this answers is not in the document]
ianmcorvidae: Haha, fair enough. I always felt like react would complain at me at the weirdest times so I got in the habit of always putting brackets, but it is indeed cleaner not to.

I wish prettier just managed it for us though, lol

ianmcorvidae:
Thanks for the reviews! I know it's past 5 on a Friday, but I think I'm gonna go ahead and merge this anyway.

PR #247, opened 2020-12-04
Tools - Part 1

sri2k1us:
- Link `Manage Tools` from Apps View
- Migrate to `i18next`
- Add query param for paging
- Save page size in local storage
- Remove `Toggle Display` button from all the toolbars. (needed extra space for `Manage Tools` button in Apps Toolbar.)


src/pages/tool.js
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * A page to access the DE tools.
+ *
+ */
+
+import React from "react";
+
+import Listing from "components/tools/listing/Listing";
+
+export default function Tools() {
```
ash3rz, line 13: Since the apps and analyses pages are plural, maybe for consistency this file should be renamed to `/pages/tools`? I think the navigation constant is also `tools`.

I checked this out locally and it seems somehow both `/tool` and `/tools` are working.

src/components/apps/toolbar/Toolbar.js
```diff
@@ -192,6 +185,20 @@ function AppsToolbar(props) {
...
+                            id={build(appsToolbarId, ids.TOOLS_BTN)}
+                            className={classes.toolbarItems}
+                            variant="outlined"
+                            disableElevation
+                            color="primary"
+                            startIcon={<Build />}
+                        >
+                            {t("manageTools")}
+                        </Button>
+                    </Link>
```
ash3rz, line 200: Is there no way to get to the tools view on mobile? 🤔

public/static/locales/en/tools.json
```diff
@@ -0,0 +1,123 @@
...
+    "detailsLbl": "Details",
+    "resourceRequirementsLbl": "Resource Requirements",
+    "minCPUCoresLbl": "Minimum CPU Cores",
+    "maxCPUCoresLbl": "Maximum CPU Cores",
+    "minMemoryLimitLbl": "Minimum Memory",
+    "minDiskSpaceLbl": "Minimum Disk Space (bytes)",
+    "enabled": "Enabled",
+    "ariaCheckbox": "{{label}} checkbox",
+    "ariaTableListing": "Tool table listing",
+    "status": "Status"
```
ash3rz, line 122: It'd be nice to sort these lines.  Not sure about VS Code, but I know Atom has an extension that will auto-sort for you.

ash3rz (approved):
LGTM, though I think I caught a bug for mobile and had a question about the paging. :shipit:

psarando (approved):
LGTM 🎉 

I didn't catch anything beyond what @ash3rz already mentioned.

src/pages/tool.js
sri2k1us: Yup. I created `tool` page by mistake without realizing there was already a `tools` page. `tool` page should be deleted in #248.

src/components/apps/toolbar/Toolbar.js
sri2k1us: Good Catch 👍🏽

sri2k1us:
Thanks for the reviews.

PR #250, opened 2020-12-14
Enable compression for routes mapped by the custom express server

ianmcorvidae:
From lighthouse testing, it says there's savings to be made by doing this:

URL | Transfer Size | Potential Savings
-- | -- | --
https://sonora-qa.cyverse.org | 44 KB | 33.2 KB
/api/dashboard?limit=8(sonora-qa.cyverse.org) | 34.5 KB | 26.9 KB

next.js already supports compression, but doesn't do it for places where we're circumventing it, which includes both `/` and all the `/api` endpoints. So, this imports the same module that nextjs uses, and applies it to those endpoints. At some point, we'll probably put nginx in front of sonora, at which point we can either set this to be configurable (along with the nextjs default compression) or remove it outright. But for now, this will produce a small performance benefit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
         "bowser": "^2.9.0",
         "camelcase": "^5.3.1",
         "clsx": "^1.1.0",
+        "compression": "^1.7.4",
         "config": "^3.3.1",
         "connect-pg-simple": "^6.1.0",
         "dataloader": "^2.0.0",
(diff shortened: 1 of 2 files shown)


sri2k1us (approved):
LGTM :shipit:

ianmcorvidae:
Thanks for the review!

PR #251, opened 2020-12-15
Make the dashboard skeletons work again

ianmcorvidae:
Another small PR. I also hid the app/analysis detail drawers behind `next/dynamic` so those don't load outright.

The main bits here:
 - the skeletons are now returned directly rather than in a "loading" div. doing it the prior way meant nothing actually appeared
 - the skeletons now mimic the layout of the dashboard (particularly the news/events side-by-side in desktop mode), which has changed since they were originally done

### DIFF
--- a/src/components/dashboard/index.js
+++ b/src/components/dashboard/index.js
@@ -6,6 +6,8 @@
  * @module dashboard
  */
 import React, { useRef, useState } from "react";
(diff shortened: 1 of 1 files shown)


sri2k1us (approved):
LGTM :shipit:

ianmcorvidae:
Thanks for the review!

PR #252, opened 2020-12-15
CORE-1165 Integrate Metadata listing in data/ds/item page

psarando:
This PR will integrate the Metadata listing with `/data/ds/[...pathItem]` pages.

A new `?view=metadata` navigation query param constant was added that this page can use to determine whether to display metadata on a file or folder, rather than that item's contents.

For example, `/data/ds/iplant/home/ipcdev/bilbo-funny-1.jpg?view=metadata`.

I also reorganized the data listing dot menus a bit, especially since it was far too easy for me to accidentally click `Delete` right after opening the dot menu when it was the first item in the list. So now `Delete` is at the bottom under a separating divider:
Data Listing Toolbar Dot-menu

---

On mobile:
Data Listing Toolbar Dot-menu - mobile

---

Similarly the listing row dot-menus were rearranged:
Data Listing Row Dot-menu - mobile

### DIFF
--- a/src/common/NavigationConstants.js
+++ b/src/common/NavigationConstants.js
@@ -1,4 +1,7 @@
-// constants that define navigation routes
+/**
+ * Constants that define navigation routes.
+ * Warning: server/index.js expects each value to be a string.
+ */
 export default {
     ADMIN: "admin",
     ANALYSES: "analyses",
@@ -16,7 +19,13 @@ export default {
     RELAUNCH: "relaunch",
     SEARCH: "search",
     SETTINGS: "settings",
-    VICE: "vice",
     TOOLS: "tools",
+    VICE: "vice",
     YOUTUBE_EMBED_BASE: "https://youtu.be",
+};
+
+export const NavigationParams = {
+    // This could be updated with more options in the future,
+    // such as CONTENTS, DETAILS, or LISTING.
+    VIEW: { METADATA: "metadata" },
 };
--- a/src/components/data/toolbar/DataDotMenu.js
+++ b/src/components/data/toolbar/DataDotMenu.js
@@ -17,6 +17,7 @@ import UploadMenuItems from "./UploadMenuItems";
 import { useTranslation } from "i18n";
 import DetailsMenuItem from "../menuItems/DetailsMenuItem";
 import DeleteMenuItem from "../menuItems/DeleteMenuItem";
+import MetadataMenuItem from "../menuItems/MetadataMenuItem";
 import SharingMenuItem from "components/sharing/SharingMenuItem";
 import PublicLinksMenuItem from "../menuItems/PublicLinksMenuItem";
 import PathListAutomation from "../PathListAutomation";
@@ -66,6 +67,7 @@ function DataDotMenu(props) {
         canShare,
         setSharingDlgOpen,
         isSmall,
+        onMetadataSelected,
         onPublicLinksSelected,
     } = props;
 
@@ -83,6 +85,8 @@ function DataDotMenu(props) {
         ? getSelectedResources()
         : null;
     const deleteMiEnabled = !isSelectionEmpty && isOwner(selectedResources);
+    const metadataMiEnabled = selected?.length === 1;
+
     const router = useRouter();
     const routeToFile = (id, path) => {
         router.push(
@@ -186,6 +190,15 @@ function DataDotMenu(props) {
                                   onDeleteSelected={onDeleteSelected}
                               />
                           ),
+                    metadataMiEnabled && (
+                        <MetadataMenuItem
+                            key={ids.METADATA_MI}
+                            baseId={baseId}
+                            resourceId={selected[0]}
+                            onClose={onClose}
+                            onMetadataSelected={onMetadataSelected}
+                        />
+                    ),
                     isWritable(permission) && [
                         <MenuItem
                             key={build(baseId, ids.CREATE_HT_FILE_MI)}
--- a/src/components/data/toolbar/DataDotMenu.js
+++ b/src/components/data/toolbar/DataDotMenu.js
@@ -151,20 +151,16 @@ function DataDotMenu(props) {
                                       setSharingDlgOpen={setSharingDlgOpen}
                                   />
                               ),
-                              deleteMiEnabled && (
-                                  <DeleteMenuItem
-                                      key={build(baseId, ids.DELETE_MENU_ITEM)}
+                              metadataMiEnabled && (
+                                  <MetadataMenuItem
+                                      key={ids.METADATA_MI}
                                       baseId={baseId}
+                                      resourceId={selected[0]}
                                       onClose={onClose}
-                                      onDeleteSelected={onDeleteSelected}
+                                      onMetadataSelected={onMetadataSelected}
                                   />
                               ),
-                              <Divider
-                                  key={build(
-                                      baseId,
-                                      ids.UPLOAD_MENU_ITEM_DIVIDER
-                                  )}
-                              />,
+                              <Divider key={ids.UPLOAD_MENU_ITEM_DIVIDER} />,
                               isWritable(permission) && (
                                   <UploadMenuItems
                                       key={build(baseId, ids.UPLOAD_MENU_ITEM)}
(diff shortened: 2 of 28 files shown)


src/components/metadata/form/index.js
```diff
@@ -440,11 +446,19 @@ const MetadataForm = ({ loading, showErrorAnnouncer, ...props }) => {
...
+    }
+
+    const { avus, "irods-avus": irodsAVUs } = metadata;
+
     return (
         <>
             <Formik
                 enableReinitialize
-                initialValues={{ ...metadata }}
+                initialValues={{ avus, "irods-avus": irodsAVUs }}
```
psarando, line 461: I guess the DE's GXT code was sanitizing the service response before passing it to the old React code, since the metadata endpoint also returns a `path` in its response that is not acceptable in the metadata add/update endpoints. So this form is now explicitly only using the keys required in the form and submission.

src/common/NavigationConstants.js
```diff
@@ -1,4 +1,7 @@
-// constants that define navigation routes
+/**
+ * Constants that define navigation routes.
+ * Warning: server/index.js expects each value to be a string.
```
psarando, line 3: Initially I tried adding the `VIEW: { METADATA: "metadata" }` object to these default constants, but I got a `string.replace is not a function` error message (without a stack trace) that I eventually tracked down to `server/index.js`.

So I added this comment to prevent others from running into the same issue 😅

sri2k1us:
LGTM. However, when I ran this locally I found couple UX improvements that could be made:
1. Move the `Add` avu to toolbar and call it out in large screen instead of just icon. It took me few seconds to figure out where the add button was.
image

2. A `Back` button to take the user back to the listing.

3. I prefer the `Delete` button to be colored rather than its hover background color getting changed to red. In other places too, we set the color for icon (for example to `primary`) and not change the hover background color. Also in mobile, there is no hover rather the color shows up when you actually press the button. 
image

4. May be `Metadata` action could be linked into the file viewer too ? (This could be saved for later though)

src/components/data/toolbar/DataDotMenu.js
```diff
@@ -259,6 +260,15 @@ function DataDotMenu(props) {
                                 onClose={onClose}
                             />
                         ),
+                    <Divider key={ids.DELETE_MENU_ITEM_DIVIDER} />,
```
ash3rz, line 263: 👍

ash3rz (approved):
LGTM :shipit: 
I mostly agree with Sriram's UX comments as well, though I wanted to add that I think the Add AVU button should remain visible on mobile if possible since that's probably a primary action.  Feels weird if that ends up getting hidden within the dot menu on mobile, to me at least.

psarando:
> LGTM. However, when I ran this locally I found couple UX improvements that could be made:

Most of this UX is a direct import from the existing UX in the DE, and I was hoping UX changes for Sonora could have been brought up in #238, and to keep this PR just for page integration and URL / listing navigation UX.

FYI, if you don't want to checkout a PR and spin up Storybook locally, Chromium will still build our stories, even when our snapshot limit is reached: https://5ebad5c6b991d60022b1e72e-hutdrspubj.chromatic.com/?path=/story/metadata--metadata-view

With that said, I think I can implement some of these improvements in another PR, but let me address them 1 at a time, below:

> 1. Move the `Add` avu to toolbar and call it out in large screen instead of just icon. It took me few seconds to figure out where the add button was. Suggested change:

This is a good idea, however, the `+ AVU` header and the listing below it are reused for nested/child metadata listings:
Edit AVU Dialog

So I'm not sure where to move this button in the AVU Dialog 🤔 

> 2. A `Back` button to take the user back to the listing.

This shouldn't be a problem and I can add this in this PR (this is the kind of Sonora UX navigation feedback I was looking for in this PR) 👍 

> 3. I prefer the `Delete` button to be colored rather than its hover background color getting changed to red. In other places too, we set the color for icon (for example to `primary`) and not change the hover background color. Also in mobile, there is no hover rather the color shows up when you actually press the button.

I agree, the delete icon should probably be solid red, instead of red on hover, which will be nicer for mobile users.
I'll save this for a UX improvement PR, though.

> 4. May be `Metadata` action could be linked into the file viewer too ? (This could be saved for later though)

Sounds good 👍 I'll look into it (since this is a Sonora integration feature, it seems like it belongs in this PR).

sri2k1us:
> > LGTM. However, when I ran this locally I found couple UX improvements that could be made:
> 
> Most of this UX is a direct import from the existing UX in the DE, and I was hoping UX changes for Sonora could have been brought up in #238, and to keep this PR just for page integration and URL / listing navigation UX.
> 
> FYI, if you don't want to checkout a PR and spin up Storybook locally, Chromium will still build our stories, even when our snapshot limit is reached: https://5ebad5c6b991d60022b1e72e-hutdrspubj.chromatic.com/?path=/story/metadata--metadata-view
> 
> With that said, I think I can implement some of these improvements in another PR, but let me address them 1 at a time, below:
> 
> > 1. Move the `Add` avu to toolbar and call it out in large screen instead of just icon. It took me few seconds to figure out where the add button was. Suggested change:
> 
> This is a good idea, however, the `+ AVU` header and the listing below it are reused for nested/child metadata listings:
> Edit AVU Dialog
> 
> So I'm not sure where to move this button in the AVU Dialog 🤔
> 
> > 1. A `Back` button to take the user back to the listing.
> 
> This shouldn't be a problem and I can add this in this PR (this is the kind of Sonora UX navigation feedback I was looking for in this PR) 👍
> 
> > 1. I prefer the `Delete` button to be colored rather than its hover background color getting changed to red. In other places too, we set the color for icon (for example to `primary`) and not change the hover background color. Also in mobile, there is no hover rather the color shows up when you actually press the button.
> 
> I agree, the delete icon should probably be solid red, instead of red on hover, which will be nicer for mobile users.
> I'll save this for a UX improvement PR, though.
> 
> > 1. May be `Metadata` action could be linked into the file viewer too ? (This could be saved for later though)
> 
> Sounds good 👍 I'll look into it (since this is a Sonora integration feature, it seems like it belongs in this PR).

TBH - I did run and checked #238 locally but somehow didn't feel the same way about the `+` button when looking in SB at least. Here is the screenshot
image

This looks like the whole thing is inside a dialog rather than a page along with tabs and others things. It looks different (perspective) when looked in Sonora page. But anyways, I try to keep this in mind.

psarando:
> This looks like the whole thing is inside a dialog rather than a page along with tabs and others things. It looks different (perspective) when looked in Sonora page. But anyways, I try to keep this in mind.

I see what you mean. Good point 👍

sri2k1us (commented):
LGTM so far. But It seems like I cannot view metadata for community data ? Is this planned for another PR ? This feature is available from DE right now.
image
image

sri2k1us:
> LGTM so far. But It seems like I cannot view metadata for community data ? Is this planned for another PR ? This feature is available from DE right now.
> image
> image

Oops I just realized I missed checking the row level dot menu. I found it there!. Anyway for consistency, `Metadata` Menu option should also be available from `Dot` menu in the toolbar.
image

sri2k1us (approved):

src/components/utils/BackButton.js
```diff
@@ -0,0 +1,34 @@
...
+            variant={isMobile ? "text" : "contained"}
+            size="small"
+            startIcon={<ArrowBack fontSize="small" />}
+            onClick={() => router && router.back()}
+            {...props}
+        >
+            <Hidden xsDown>{t("back")}</Hidden>
+        </Button>
+    );
+}
```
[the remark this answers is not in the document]
psarando: I agree and I also prefer an icon button on mobile.
I'll make that change for this button before merging.

psarando:
> Oops I just realized I missed checking the row level dot menu. I found it there!. Anyway for consistency, `Metadata` Menu option should also be available from `Dot` menu in the toolbar.

Not sure why it's not showing up in the toolbar for read-only data, but I'll look into it 👍

psarando:
> 1. Move the `Add` avu to toolbar and call it out in large screen instead of just icon. It took me few seconds to figure out where the add button was. Suggested change:

> 3. I prefer the `Delete` button to be colored rather than its hover background color getting changed to red. In other places too, we set the color for icon (for example to `primary`) and not change the hover background color. Also in mobile, there is no hover rather the color shows up when you actually press the button.

I was going to save these updates for another UX improvement PR where I had intentions to add bigger changes for review (such as displaying child metadata in nested, collapsible rows, and only displaying form fields for editing AVUs in the edit dialog), but I already have some quick and easy fixes to address these issues ready for this PR 😅 

I added text to the "Add Metadata" button and just moved it to the far-right side of the listing title for now.
I also went ahead and removed the hover color of the AVU delete button and made its icon color red instead.
I also put the edit and delete buttons together in a button group (rather than the more hacky grid layout that was used before).
Large screen UX improvements

On mobile the "Add Metadata" button is still the `+` icon button:
Mobile UX improvements

I rebased on master and I addressed all the other comments mentioned in this PR, so I'll go ahead and merge this PR with these UX fixes so they're available before the break, then I'll come back to the bigger metadata listing UX improvements later.

Thanks for the reviews! 👍 
Additional reviews on these changes still welcome (but may not be addressed until the New Year) 🙇

PR #254, opened 2020-12-16
Tools - Part 5

sri2k1us:
- This is on top of #253 
-  Tool filter and search
- Filtering
image
- Search
image
-iPhone 5
image


src/components/tools/listing/Listing.js
```diff
@@ -51,26 +57,52 @@ function Listing(props) {
...
     useEffect(() => {
-        setToolsKey([TOOLS_QUERY_KEY, { order, orderBy, page, rowsPerPage }]);
+        let displayAll = null;
+
+        if (permFilter) {
+            if (permFilter === TOOL_FILTER_VALUES.MY_TOOLS) {
+                displayAll = false;
+            } else {
+                displayAll = true;
+            }
```
psarando, line 97:
```suggestion
            displayAll = (permFilter !== TOOL_FILTER_VALUES.MY_TOOLS);
```

src/pages/tools.js
```diff
@@ -19,18 +18,19 @@ import NavigationConstants from "common/NavigationConstants";
...
     const selectedOrderBy = query.selectedOrderBy || "name";
+    const selectedPermFilter = query.selectedPermFilter;
+    const selectedSearchTerm = query.selectedSearchTerm;
 
     const onRouteToListing = useCallback(
-        (order, orderBy, page, rowsPerPage) => {
+        (order, orderBy, page, rowsPerPage, permFilter, searchTerm) => {
             router.push(
-                `/${NavigationConstants.TOOLS}?selectedOrder=${order}&selectedOrderBy=${orderBy}&selectedPage=${page}&selectedRowsPerPage=${rowsPerPage}`
+                `/${NavigationConstants.TOOLS}?selectedOrder=${order}&selectedOrderBy=${orderBy}&selectedPage=${page}&selectedRowsPerPage=${rowsPerPage}&selectedPermFilter=${permFilter}&selectedSearchTerm=${searchTerm}`
```
psarando, line 33:
```suggestion
                {
                    pathname: `/${NavigationConstants.TOOLS}`,
                    query: {
                        selectedOrder: order,
                        selectedOrderBy: orderBy,
                        selectedPage: page,
                        selectedRowsPerPage: rowsPerPage,
                        selectedPermFilter: permFilter,
                        selectedSearchTerm: searchTerm,
                    },
                }
```
I also wonder if `pathname: NavigationConstants.TOOLS` would work too.

psarando (commented):
LGTM 👍 

I had a couple of minor code-style suggestions, but I also noticed there are now a lot of conflicts, so I'm not sure if I'll need to re-review after a rebase.

src/pages/tools.js
sri2k1us: Path needs to start with `/` as far as I know.

psarando (approved):
LGTM 👍 
:shipit: when ready.

ash3rz (approved):
Code LGTM :shipit: 
For the screenshots, I believe before the break we decided the global search was going to include things like tools and teams so the search is truly "global".  Here, you've kept a search bar for tools, but for teams I've removed the search bar that was originally there in the old DE's team dialog.  We might need a discussion on if/when it makes sense to have a secondary, inner search bar.

sri2k1us:
> Code LGTM :shipit:
> For the screenshots, I believe before the break we decided the global search was going to include things like tools and teams so the search is truly "global". Here, you've kept a search bar for tools, but for teams I've removed the search bar that was originally there in the old DE's team dialog. We might need a discussion on if/when it makes sense to have a secondary, inner search bar.

Yup remember that. Lets bring up this issue in our next meeting.

sri2k1us:
Thanks for the reviews.

PR #255, opened 2020-12-16
Tools - Part 6

sri2k1us:
- This is on top of #254 
- Add New tool request
- 
image


src/components/tools/NewToolRequestDialog.js
```diff
@@ -0,0 +1,180 @@
...
+    const handleSubmit = (values, { props }) => {
+        //test_data_path and cmd_line are required by service. Most of the times those info. can be
+        // found in the source repo. So copy the source url to those fields if not provided by user.
+        if (!values.test_data_path) {
+            values.test_data_path = values.source_url;
+        }
+        if (!values.cmd_line) {
+            values.cmd_line = values.source_url;
+        }
+        const submission = { ...values };
```
psarando, line 54: Good comments 👍 
I would not modify the form's values, though, but modify the `submission` copy instead:
```suggestion
        const submission = { ...values };
        
        //test_data_path and cmd_line are required by service. Most of the times those info. can be
        // found in the source repo. So copy the source url to those fields if not provided by user.
        if (!submission.test_data_path) {
            submission.test_data_path = submission.source_url;
        }
        if (!submission.cmd_line) {
            submission.cmd_line = submission.source_url;
        }
```

src/components/tools/NewToolRequestDialog.js
```diff
@@ -0,0 +1,180 @@
...
+                                margin="dense"
+                                id={build(baseId, ids.TOOL_REQUEST.DESCRIPTION)}
+                                component={FormMultilineTextField}
+                            />
+                            <Field
+                                name="source_url"
+                                label={t("toolSrcLinkLabel")}
+                                required={true}
+                                margin="dense"
+                                validate={validateUrl}
```
psarando, line 119: This is the only validation check I see in this form.

I haven't tried this locally yet, but does this mean the user can submit this form without filling in the other required fields? If that's the case, then the error message that tells the user to "Please try again later" is misleading if the service is returning an error about a missing form field.

In our other forms we have validation checks that will return a `Required` error message if a required field is empty, which also prevents the form submitting to the service.

psarando (commented):
Looks good so far, but I had a couple of minor suggestions and a validation question.
Also, this PR has conflicts, so I'm not sure if I'll have to review again after a rebase.

src/components/tools/NewToolRequestDialog.js
sri2k1us: Hmm..There are required fields in this form. Only the `source_url` field needs additional validation for a valid url. See screenshots below.

image

image

psarando: I see the required fields are marked as required by MUI, but I don't think Formik knows about that, and the question is what happens when you try to submit the form without filling in those fields? I think that's what you are showing in the first screenshot? Then what happens if you "touch" a required field without entering anything, then focus on another field?

Without a validation function to return a `Required` error message, I don't think the required fields will be highlighted in red, like the URL field in the second screenshot. Note that's how it works in the `Add Tool` form, if you focus on the tool name, then focus on another field:
Add Tool form errors

The App Launch form also behaves this way, as well as highlighting required fields when submitting the form even without touching any fields:
App Launch form error


We should keep our form validation behavior consistent.

sri2k1us: Yup. The first screenshot shows what happens when you try to submit the form without entering anything. And right now nothing happens when focus out of a field. I just copied these over from DE so I checked there now and it has the same behavior. But I agree that it needs to consistent. I will update the logic.

src/components/tools/NewToolRequestDialog.js
```diff
@@ -0,0 +1,180 @@
...
+                    version: "",
+                    documentation_url: "",
+                    cmd_line: "",
+                    test_data_path: "",
+                }}
+                enableReinitialize={true}
+                onSubmit={handleSubmit}
+            >
+                {() => {
+                    return (
```
ash3rz, line 96: I think if you're not using any of Formik's helper props, you can begin the `<Form>` tag as a child to `<Formik>`, meaning you don't need this extra function and return

src/components/tools/NewToolRequestDialog.js
```diff
@@ -0,0 +1,180 @@
...
+                            />
+                            <Button
+                                style={{ float: "right" }}
+                                color="primary"
+                                type="submit"
+                                id={build(baseId, ids.BUTTONS.SUBMIT)}
+                                aria-label={t("submit")}
+                            >
+                                {t("submit")}
+                            </Button>
```
ash3rz, line 173: Is this submit button within the form? I would think this would be within `DialogActions` so it's always visible and doesn't scroll

src/components/tools/toolbar/Toolbar.js
```diff
@@ -232,6 +235,11 @@ export default function ToolsToolbar(props) {
                     </Button>
                 </DialogActions>
             </Dialog>
+            <NewToolRequestDialog
+                open={requestDialogOpen}
+                onClose={() => setRequestDialogOpen(false)}
+                baseId={baseId}
```
ash3rz, line 241: Minor, but this `baseId` shouldn't be necessary.  Dialogs start with their own ID since they're not children to something else.

ash3rz (approved):
I may have caught one thing with the submit button, but otherwise LGTM :shipit:

src/components/tools/NewToolRequestDialog.js
sri2k1us: Hmm I have tried it earlier and I tried it again now but putting the `Submit` button outside doesn't seem to be working. I am newbie when it comes `Formik` so I may be missing something.

src/components/tools/NewToolRequestDialog.js
sri2k1us: Ok this should be done!.

src/components/tools/NewToolRequestDialog.js
ash3rz: Ah, I'm mixing up words - when I said `form` I meant the dialog content, not the `Form` tag.  At least for [CreateFolderDialog](https://github.com/cyverse-de/sonora/blob/eae296849154c04c1f2b6a4a58fe7db65b5e7b2c/src/components/data/CreateFolderDialog.js#L74-L86), I put the entire dialog within `Form`.  Though, Paul may know a different way to do this.

sri2k1us: Ahh ok. Even then since I am using `DEDialog`, I will have to pass it as `actions` prop and that doesn't seem to work 😩

psarando: As far as I know, the submit button has to be inside the `Form`, but wrapping the entire dialog in the `Form` should work 👍

psarando: Ah, yeah, using `DEDialog` makes this trickier 🤔

sri2k1us: Yup. May `DEDialog` shouldn't be used with `Formik` or should we refactor `DEDialog` to include `DialogActions` as children 🤔

ash3rz: Does this component work?  It's using `DEDialog` with Formik: https://github.com/cyverse-de/sonora/blob/master/src/components/data/SaveAsDialog.js

psarando: Right now the `NewToolRequestDialog` is return something like
```jsx
return (
    <DEDialog ...>
        <Formik ...>
            <Form>...</Form>
        </Formik>
    </DEDialog>
);
```
But this should work too:
```jsx
return (
    <Formik ...>
        <Form ...>
            <DEDialog>...</DEDialog>
        </Form>
    </Formik>
);
```

psarando: @ash3rz Jinx! That's pretty much what I was just thinking 👍

sri2k1us: Ok got it to work. You have to call `handleSubmit` `onClick` explicitly with this approach. Thanks @ash3rz @psarando.

sri2k1us: I will probably do the same for https://github.com/cyverse-de/sonora/blob/master/src/components/tools/edit/EditTool.js form as well .

src/components/tools/NewToolRequestDialog.js
```diff
@@ -0,0 +1,189 @@
...
+                return (
+                    <Form>
+                        <DEDialog
+                            open={open}
+                            onClose={onClose}
+                            baseId={baseId}
+                            title={t("newToolRequestDialogHeading")}
+                            actions={
+                                <Button
+                                    style={{ float: "right" }}
```
psarando, line 94: Now that this button is inside the Dialog Actions, this style isn't needed.

psarando (approved):
LGTM 🎉

sri2k1us:
Thanks for the reviews.

PR #256, opened 2020-12-17
Tools - Part 7

sri2k1us:
- This is on top #255 
- Add delete tool functionality
- 
image

image

This complete the `tools` PR series. All the features except `Use it an app` has been implemented.


src/components/tools/listing/Listing.js
```diff
@@ -278,8 +311,19 @@ function Listing(props) {
...
+            <ConfirmationDialog
+                baseId={baseId}
+                open={deleteConfirm}
+                onConfirm={() => {
+                    setDeleteConfirm(false);
+                    removeTools({ ids: selected });
+                }}
+                onClose={() => setDeleteConfirm(false)}
+                title={t("delete")}
+                contentText={t("confirmDelete")}
```
psarando, line 323: 👍

public/static/locales/en/tools.json
```diff
@@ -8,6 +8,7 @@
     "attribution": "Attribution",
     "bindToHost": "Bind to Host",
     "cancel": "Cancel",
+    "confirmDelete": "Do you really want to delete this tool?",
```
psarando, line 11: Should this say "Delete the selected tools", since the user can delete more than 1 at once?
```suggestion
    "confirmDelete": "Delete the selected tools?",
```

src/components/tools/toolbar/ToolsDotMenu.js
```diff
@@ -149,7 +172,13 @@ export default function ToolsDotMenu({
...
         selectedTools &&
         selectedTools.length > 0 &&
-        selectedTools[0].permission === "own";
+        (selectedTools[0].permission === "own" ||
+            selectedTools[0].permission === "write");
+    const allowDelete =
+        selectedTools &&
+        selectedTools.length > 0 &&
+        (selectedTools[0].permission === "own" ||
+            selectedTools[0].permission === "write");
```
psarando, line 181: These permissions checks should now use the `isWritable` util function.

Also, `allowDelete` is allowing multiple selected tools, but it's only checking permissions on the first tool.
So maybe this will need to do a `find` to check for a tool without write permissions.

src/serviceFacades/tools.js
```diff
@@ -118,6 +118,25 @@ function toolRequest(request) {
...
+
+function deleteTools({ ids: toolIds }) {
+    return (
+        toolIds &&
+        toolIds.length > 0 &&
+        new Promise((resolve, reject) =>
+            Promise.all(toolIds.map((id) => deleteTool({ id })))
+                .then((values) => resolve(values))
+                .catch((error) => reject(error))
+        )
```
psarando, line 136: It turns out the `new Promise((resolve, reject)` wrapper isn't needed:
```suggestion
        Promise.all(toolIds.map((id) => deleteTool({ id })))
```

psarando (changes requested):
So far so good, but I think I found a permissions check bug.

src/components/tools/listing/Listing.js
```diff
@@ -86,6 +97,23 @@ function Listing(props) {
...
+            });
+            //reset selection to avoid stale selected state
+            setSelected([]);
+            queryCache.invalidateQueries(toolsKey);
+        },
+        onError: (e) => {
+            showErrorAnnouncer(t("toolDeleteError"), e);
+            //reset selection to avoid stale selected state
+            setSelected([]);
+            queryCache.invalidateQueries(toolsKey);
```
ash3rz, line 113: Minor, but I wonder if it might be helpful to keep the selection on error so that the user can try to delete the tools again (in case they selected more than one)?

ash3rz (approved):
LGTM :shipit:

src/components/tools/listing/Listing.js
psarando: It would be more convenient to try again, in case of transient errors; but because the UI is batching the API calls, which can only delete 1 tool per call, then it's probably best to reset the selection and reload the listing, in case some of the selection was successfully deleted 🤔 

Otherwise we'd have to figure out something tricky with the promises in order to fix the selection based on which were successful and which failed.

psarando (approved):
LGTM 👍

src/components/tools/listing/Listing.js
ash3rz: Ah, forgot it's a batch call.  Good point.  Better to reset to avoid additional errors.

sri2k1us:
Thanks for the reviews.

PR #257, opened 2020-12-17
Fix browser navigation for data view.

sri2k1us:
- I eliminated props to state copy of `order`, `orderBy`, `page` and `rowsPerPage`. That seems have fixed most of the issues that we have seen so far with respect to browser `Back` and `Forward` buttons based navigation in Data View. We might want to do the same for other views as well.
- Request reviewers to test these changes locally to make sure edge cases aren't missed.

### DIFF
--- a/src/components/data/SelectionDrawer.js
+++ b/src/components/data/SelectionDrawer.js
@@ -28,9 +28,12 @@ import Listing from "components/data/listing/Listing";
 import ResourceTypes from "../models/ResourceTypes";
 
 import styles from "./styles";
+import constants from "../../constants";
 
 import PageWrapper from "../../components/layout/PageWrapper";
 import useComponentHeight from "../utils/useComponentHeight";
+import { DEFAULT_PAGE_SETTINGS } from "components/data/utils";
+import { getLocalStorage } from "components/utils/localStorage";
 
 const useStyles = makeStyles(styles);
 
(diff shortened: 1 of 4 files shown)


psarando (changes requested):
LGTM, except in local testing, changing the page size from 100 to 500 doesn't seem to do anything to the page, and I'm not sure why yet 😕 

If I view the "Shared with Me" listing that has 2 pages at 100 items per page, changing to 500 items per page does change the URL query param to `selectedRowsPerPage=500`, but the listing still displays 2 pages and 100 per page. Navigating to page 2 resets the URL query param to `selectedRowsPerPage=100`.

sri2k1us:
> LGTM, except in local testing, changing the page size from 100 to 500 doesn't seem to do anything to the page, and I'm not sure why yet 😕
> 
> If I view the "Shared with Me" listing that has 2 pages at 100 items per page, changing to 500 items per page does change the URL query param to `selectedRowsPerPage=500`, but the listing still displays 2 pages and 100 per page. Navigating to page 2 resets the URL query param to `selectedRowsPerPage=100`.

Ok.let me check tomorrow

ash3rz (approved):
LGTM after the one line fix 👍 tested locally

psarando (approved):
LGTM 🎉 
Tested locally and didn't catch any more issues 👍 
:shipit:

sri2k1us:
Thanks for the reviews.

PR #258, opened 2020-12-18
Route to appropriate tab when search is filtered.

sri2k1us:
- SSIA

### DIFF
--- a/src/components/search/GlobalSearchField.js
+++ b/src/components/search/GlobalSearchField.js
@@ -649,19 +649,17 @@ function GlobalSearchField(props) {
             onKeyDown={(e) => {
                 if (e.key === "Enter") {
                     setOpen(false);
-                    if (filter !== searchConstants.ALL) {
-                        router.push({
-                            pathname: `/${NavigationConstants.SEARCH}`,
-                            query: {
-                                searchTerm,
-                                filter,
-                                selectedTab:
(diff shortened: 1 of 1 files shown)


src/components/search/GlobalSearchField.js
```diff
@@ -649,9 +649,19 @@ function GlobalSearchField(props) {
             onKeyDown={(e) => {
                 if (e.key === "Enter") {
                     setOpen(false);
-                    router.push(
-                        `/${NavigationConstants.SEARCH}?searchTerm=${searchTerm}&filter=${filter}`
-                    );
+                    if (filter !== searchConstants.ALL) {
```
ash3rz, line 652: There's another check below to see if the filter is set to `ALL`.  I think we still want users to be directed to the search page when they hit the `Enter` key, so perhaps this check is unnecessary?

ash3rz (approved):
Had a small question, otherwise LGTM 👍

src/components/search/GlobalSearchField.js
sri2k1us: Good Catch. Indeed a bug. Should be fixed now.

psarando (approved):
LGTM :shipit: 👍

sri2k1us:
Thanks for the reviews.

PR #259, opened 2020-12-23
Selection drawer updates

sri2k1us:
- Remove toolbar actions and row level dot menu from Selection Drawer.
- Styling improvement
- Valid Selection
  Prod:
image

This PR:
Now the drawer is clutter and distraction free. Also  found that selecting `Metadata` or `New HT analysis path list file (or Multi-Input)` from the selection drawer throws errors in prod. So I think getting rid of these actions from here is step in the right direction. We can look into adding upload option based on user feedback. 
image

- invalid Selection
Prod:
image

This PR:
image

### DIFF
--- a/src/components/data/listing/Listing.js
+++ b/src/components/data/listing/Listing.js
@@ -73,6 +73,8 @@ function Listing(props) {
         order,
         orderBy,
         onRouteToListing,
+        toolbarVisibility = true,
+        rowDotMenuVisibility = true,
     } = props;
     const { t } = useTranslation("data");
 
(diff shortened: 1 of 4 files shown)


src/components/data/listing/Listing.js
```diff
@@ -73,6 +73,8 @@ function Listing(props) {
         order,
         orderBy,
         onRouteToListing,
+        toolbarVisibility = true,
+        rowDotMenuVisibility = true,
```
ash3rz, line 77: 👍 
I couldn't think of a simple way to make these changes when you suggested this before the break.  This is nice!

ash3rz (approved):
LGTM :shipit:

psarando (approved):
LGTM ✨

sri2k1us:
Thanks for the reviews.

PR #263, opened 2021-01-08
Admin apps

sri2k1us:
- Allow users to browse all apps (including private)
- View Details
- Mark as `Deleted` or `Disabled` 
- Apply / Remove `Beta` tag
- Add / Update Doc

image

image

- iPhone 5

image


src/components/apps/admin/details/AdminAppDetails.js
```diff
@@ -0,0 +1,324 @@
...
+                    }
+                >
+                    {t("documentationTemplate")}
+                </Link>
+            </Paper>
+            <Field
+                name={"documentation.documentation"}
+                label={t("appDocumentation")}
+                id={build(parentId, ids.ADMIN_DETAILS.DOCUMENTATION)}
+                t
```
psarando, line 257: `t`ypo?

src/components/apps/admin/details/AdminAppDetails.js
```diff
@@ -0,0 +1,324 @@
...
+                    </Button>
+                </Grid>
+                <Grid item>
+                    <Button
+                        id={build(parentId, ids.SAVE_BTN)}
+                        type="submit"
+                        color="primary"
+                    >
+                        {i18nCommon("save")}
+                    </Button>
```
psarando, line 284: These could go into the `actions` prop.

psarando (changes requested):
Looks good so far 👍 
I had some minor clean-up suggestions, but I also may have caught a bug setting `toFilter` to `null`.

src/components/apps/admin/details/AdminAppDetails.js
sri2k1us: Its funny 😂  and frustrating 😩 that how I can get away with these issue without encountering errors when testing.

src/components/apps/admin/details/AdminAppDetails.js
```diff
@@ -0,0 +1,343 @@
...
+    FormMultilineTextField,
+    FormTextField,
+} from "@cyverse-de/ui-lib";
+
+import DEDialog from "components/utils/DEDialog";
+import { getHost } from "components/utils/getHost";
+import { Skeleton } from "@material-ui/lab";
+import ErrorTypographyWithDialog from "components/utils/error/ErrorTypographyWithDialog";
+
+import { Button, CircularProgress, Grid, Link, Paper } from "@material-ui/core";
```
psarando, line 34: There's a warning now that `Grid` is no longer used.

src/serviceFacades/apps.js
```diff
@@ -207,8 +213,154 @@ function saveAppDoc({ systemId, appId, documentation }) {
...
+}
+
+function getAppAVUs(key, { appId }) {
+    return callApi({
+        endpoint: `/api/admin/apps/${appId}/metadata`,
+        method: "GET",
+    });
+}
+
+function addAVUToApp({ appId, avu }) {
```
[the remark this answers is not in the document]
psarando: I still recommend this rename ☝️

src/serviceFacades/apps.js
```diff
@@ -221,6 +366,13 @@ export {
     searchAppsInfiniteQuery,
     getAppDoc,
     saveAppDoc,
+    adminPatchApp,
```
psarando, line 369: I still recommend not exporting this until it's need elsewhere.

psarando (approved):
LGTM 🎉 
I found a new warning, and just added a couple of reminders of minor suggestions from my last review.
Otherwise :shipit: when ready 👍

src/components/apps/utils.js
```diff
@@ -31,11 +32,22 @@ export const getListingPath = (
...
     filter,
-    category
+    category,
+    isAdmin
 ) => {
-    const encodedFilter = encodeURIComponent(filter);
-    const encodedCategory = encodeURIComponent(category);
-    return `/${NavigationConstants.APPS}?selectedOrder=${order}&selectedOrderBy=${orderBy}&selectedPage=${page}&selectedRowsPerPage=${rowsPerPage}&selectedFilter=${encodedFilter}&selectedCategory=${encodedCategory}`;
+    return {
+        pathname: isAdmin
```
ash3rz, line 39: Something to consider, which is really minor, but it might be a little clearer if instead of `isAdmin`, the variable that we're using throughout these changes were `isAdminView` or something similar.   When reading through this PR, I initially got the impression that we were disabling features for administrators, rather than reusing a component for user vs. admin view.  Just a suggestion though, feel free to ignore.

ash3rz (approved):
LGTM :shipit:

src/serviceFacades/apps.js
sri2k1us: Just this end-point or all the end-points used in `adminUpdateApp` that are used elsewhere ?

psarando: Ah, good point. I think all functions not used elsewhere should be removed from the export (like keeping methods `private` until they need to be `public`).

src/serviceFacades/apps.js
```diff
@@ -221,6 +366,13 @@ export {
...
+    adminAddAppDoc,
+    adminUpdateAppDoc,
+    adminGetAppAVUs,
+    addAVUToApp,
+    adminSetAppAVUs,
```
psarando, line 374:
```suggestion
    adminGetAppAVUs,
```
So I think these are the functions that can be removed from the export.

src/components/apps/utils.js
sri2k1us: Good point although am little torn here. The variable `isAdmin`  is derived from checking the `userProfile` to see if the user belongs to admin group. May it could be copied into `isAdminView` when sending it down to `utils` and `Toolbar` files 🤔 ?. Let me know @ash3rz @psarando

psarando: I think I see the confusion now. If you don't see the code in `pages/admin/apps.js` first, then it might seem like there is only 1 app listing in our app that displays differently (or disables features) for admin users.

So I think `const isAdmin = profile?.admin` makes sense in `pages/admin/apps.js`, but it might clear up the confusion if `isAdminView` (or even just `adminView`) is used as the prop in the `Listing` on down:
```jsx
            <Listing
                baseId="apps"
                ...
                adminView={isAdmin}
            />
```

ash3rz: Yup, that would clear it up IMO

sri2k1us: Ok this should be done.

sri2k1us:
Thanks for the reviews.

PR #264, opened 2021-01-12
core-720 Allow users to download files.

sri2k1us:
- Large Screen
image

- Mobile
image

### DIFF
--- a/src/components/data/listing/Listing.js
+++ b/src/components/data/listing/Listing.js
@@ -32,6 +32,7 @@ import withErrorAnnouncer from "components/utils/error/withErrorAnnouncer";
 import Sharing from "components/sharing";
 import { formatSharedData } from "components/sharing/util";
 import { getPageQueryParams } from "../utils";
+import { getHost } from "components/utils/getHost";
 
 import {
     useUploadTrackingState,
@@ -106,6 +107,7 @@ function Listing(props) {
     const [importDialogOpen, setImportDialogOpen] = useState(false);
     const [sharingDlgOpen, setSharingDlgOpen] = useState(false);
     const [publicLinksDlgOpen, setPublicLinksDlgOpen] = useState(false);
+    const [download, setDownload] = useState(false);
 
     const onCloseImportDialog = () => setImportDialogOpen(false);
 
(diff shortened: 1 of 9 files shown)


src/components/data/listing/Listing.js
```diff
@@ -240,6 +252,20 @@ function Listing(props) {
         setDetailsEnabled(selected && selected.length === 1);
     }, [selected]);
 
+    useEffect(() => {
+        if (download) {
+            const selRes = getSelectedResources();
+            selRes.forEach((res) => {
+                window.open(
+                    `${getHost()}/api/download?path=${res.path}`,
+                    "_blank"
```
ash3rz, line 261: I can't remember if this was already discussed, but do we have plans to zip up selected resources and download them in one file in the future?  This works, but if the user wants to download 10+ files, for example, at least in FireFox you get a popup asking if you want to open or save each file.

ash3rz (approved):
LGTM - GitHub actions caught a small bug though, and I had a couple minor questions

src/components/data/menuItems/DownloadMenuItem.js
```diff
@@ -0,0 +1,31 @@
...
+import ids from "../ids";
+import { useTranslation } from "i18n";
+
+function DeleteMenuItem(props) {
+    const { baseId, onClose, onDownloadSelected } = props;
+    const { t } = useTranslation("data");
+
+    return (
+        <MenuItem
+            key={build(baseId, ids.DOWNLOAD_MENU_ITEM)}
```
psarando, line 16: This `key` is not needed here. Only `DeleteMenuItem` needs a `key` in the array in `RowDotMenu`.

psarando (approved):
I'm also wondering about the `download` state and the `useEffect` in the `Listing`.
Otherwise, LGTM 👍

src/components/data/listing/Listing.js
```diff
@@ -473,6 +492,7 @@ function Listing(props) {
                     }
                     setSharingDlgOpen={setSharingDlgOpen}
                     onPublicLinksSelected={() => setPublicLinksDlgOpen(true)}
+                    onDownloadSelected={() => setDownload(true)}
```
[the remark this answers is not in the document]
sri2k1us: So we could put the code in `useEffect` out side just as function and It may work as well but `window` object is available on the client side only. Putting in `useEffect` guarantees that the code will run only in the client-side. But I confess I have limited understanding of SSR. 🤷🏽‍♂️ 
https://medium.com/frontend-digest/why-is-window-not-defined-in-nextjs-44daf7b4604e
https://stackoverflow.com/questions/55151041/window-is-not-defined-in-next-js-react-app

src/components/data/listing/Listing.js
sri2k1us: Definitely we should add something like this for supporting bulk upload and download.

PR #267, opened 2021-01-20
Add storybook decorator for stories using nextjs Link

ash3rz:
Without this, stories using nextjs Links were erroring with `router is null` errors. Copied solution from https://github.com/lifeiscontent/storybook-addon-next-router#readme

I noticed this while checking global search stories, but a handful of other stories were also affected.

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -26,6 +26,7 @@ import { I18nextProvider } from "react-i18next";
 
 import { addDecorator } from "@storybook/react";
 import { withConsole } from "@storybook/addon-console";
+import { withNextRouter } from "storybook-addon-next-router";
 
 function MockUserProfile() {
     const [userProfile, setUserProfile] = useUserProfile();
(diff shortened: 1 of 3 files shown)


sri2k1us (approved):
Great find. I knew about this issue for while but never looked into solutions. One thing that am not clear from this PR is what happens when you click on a `Link` with this change in place? Will it perform the routing or just makes it such that there are no errors thrown. Otherwise LGTM 🎉

ash3rz:
> Great find. I knew about this issue for while but never looked into solutions. One thing that am not clear from this PR is what happens when you click on a `Link` with this change in place? Will it perform the routing or just makes it such that there are no errors thrown. Otherwise LGTM 🎉

I didn't try all the stories, but from what I did test, it tries to follow the link but from `localhost:9009`, e.g. `localhost:9009/apps/[appId]/launch` which shows a page with text saying it couldn't go there:
image

There's probably something better we could do there, but at least for now it shows you where the link would go which might be helpful.

psarando (approved):
Might be nice to eventually get Storybook to log the URL it will route to in the console, but this LGTM for now 🎉

ash3rz:
Cool, thanks for the reviews!

PR #268, opened 2021-01-20
Help

sri2k1us:
- iPhone 5 

image

- Large Screen

image

- Product Tour
image

PS: I would like to get this merged before next Tuesdays AH Demo. Thanks.


src/components/help/ProductTour.js
```diff
@@ -0,0 +1,65 @@
...
+                const stepIndex = index + (action === ACTIONS.PREV ? -1 : 1);
+                setTourStepIndex(stepIndex);
+            }
+        };
+        realCallback(callbackData);
+    };
+
+    return (
+        <Joyride
+            steps={getSteps(i18nTour, isMobile)}
```
sri2k1us, line 39: I refactored code to split out `ProductTour` into its own component.  But that also made the tour code complicated in sense that we have to do a separate (smaller) tour for mobile screens because of the sidebar status (`open` prop).

src/components/layout/CyVerseAppBar.js
```diff
@@ -760,6 +771,16 @@ function CyverseAppBar(props) {
                             {adminDrawerItems}
                         </>
                     )}
+                    <Divider />
```
ash3rz, line 774: Not a huge deal, but logged out users end up seeing 2 dividers in a row: 
image

src/constants.js
```diff
@@ -70,4 +70,9 @@ export default {
     JAVA_PATTERN_DOC:
         "https://docs.oracle.com/javase/9/docs/api/java/util/regex/Pattern.html",
     URL_REGEX: /^(?:ftp|FTP|HTTPS?|https?):\/\/[^/]+\.[^/]+.*/i,
+    DE_FAQ: "http://cyverse.github.io/UserSupport/README.html",
```
ash3rz, line 73: `https` perhaps?

ash3rz (approved):
Overall, LGTM 👍 I think I caught a few small issues, mainly the email typo.  Also in the mobile version screenshots, it seems like the section headers like `Support` don't have any left padding so they look a bit cramped.

src/components/help/Feedback.js
```diff
@@ -0,0 +1,180 @@
...
+                                        label={i18nHelp("name")}
+                                    />
+                                    <Field
+                                        id={build(baseId, ids.EMAIL)}
+                                        name="email"
+                                        required={false}
+                                        component={FormTextField}
+                                        label={i18nHelp("email")}
+                                        validate={(value) =>
+                                            nonEmptyField(value, t)
```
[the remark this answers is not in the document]
sri2k1us: I think you are right. But I think we should add a validation for a valid email ? 🤔 . If so how to do that ?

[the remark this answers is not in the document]
ash3rz: Here's something from the old DE, though it doesn't look like it's currently used by anything: https://github.com/cyverse-de/ui/blob/master/de-lib/src/main/java/org/iplantc/de/commons/client/validators/BasicEmailValidator3.java

I checked the user portal's code and seems like they're using `isEmail` from [this](https://www.npmjs.com/package/validator).  (They're also using Formik, can't recall if we told them about that or not, but good to know)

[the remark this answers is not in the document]
ash3rz: I just realized I may have misunderstood your question.  If you're asking how to add validation without requiring the field to be filled in, then I think a new validation function can be written to only return an error message if there is a nonempty value.  There might be another way to do that though.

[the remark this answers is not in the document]
sri2k1us: Thanks. I was indeed looking for how to add validation for non-required field.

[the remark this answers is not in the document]
sri2k1us: The validation lib seems nice. I think we should instead of writing our own regex.

flynn1999:
> * iPhone 5
> 
> image
> 
> * Large Screen
> 
> image
> 
> * Product Tour
>   image
> 
> PS: I would like to get this merged before next Tuesdays AH Demo. Thanks.



> * iPhone 5
> 
> image
> 
> * Large Screen
> 
> image
> 
> * Product Tour
>   image
> 
> PS: I would like to get this merged before next Tuesdays AH Demo. Thanks.

Minor style observation: 
The `Discovery Environment` header text part has 3 stylistic differences between iPhone 5 and PC browser. 
 1. The Environment part is not in bold 
 2. There is more line spacing in iPhone text than the one in PC browser
 3. Its not in uppercase

I don't know how important this is but just wanted to point out

sri2k1us:
@flynn1999 - These are good observations. But these style differences were deliberately made to account for spacing issues on small screens like iPhone 5.

public/static/locales/en/intro.json
```diff
@@ -9,7 +9,7 @@
     "searchTitle": "Search",
     "supportTitle": "Support",
     "menuTitle": "Menu",
-    "menu": "<p>Use this menu to access your Analyses, Apps and Data or to return to Dashboard from anywhere within the DE. Click to open the menu.</p>",
+    "menu": "<p>Use this menu to access your Analyses, Apps, Data or search for them. </p>",
```
psarando:
```suggestion
    "menu": "<p>Use this menu to access your Analyses, Apps, Data or to search for them. </p>",
```

src/constants.js
```diff
@@ -70,4 +70,11 @@ export default {
     JAVA_PATTERN_DOC:
         "https://docs.oracle.com/javase/9/docs/api/java/util/regex/Pattern.html",
     URL_REGEX: /^(?:ftp|FTP|HTTPS?|https?):\/\/[^/]+\.[^/]+.*/i,
+    EMAIL_REGEX: /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/,
+    DE_FAQ: "https://cyverse.github.io/UserSupport/README.html",
+    DE_GUIDE:
+        "https://cyverse-sonora-guide.readthedocs-hosted.com/en/latest/index.html",
+    CYVERSE_LC: "https://learning.cyverse.org/en/latest/",
+    DEFAULT_EMAIL: "no-reply@cyverse.org",
+    SUPPORT_EMAIL: "support@cyverse.org",
```
psarando, line 79: I wonder if these are better in configs, so it's easier to change: in the future for us, or for external deployments.

src/components/layout/CyVerseAppBar.js
```diff
@@ -805,30 +784,10 @@ function CyverseAppBar(props) {
...
-                        },
-                        tooltipContainer: {
-                            textAlign: "left",
-                        },
-                    }}
-                />
-            )}
+            <ProductTour
+                onTourExit={() => setRunTour(false)}
+                runTour={runTour}
```
psarando, line 836: I no longer see where this AppBar can set `runTour` to true.

src/components/help/Feedback.js
```diff
@@ -0,0 +1,186 @@
...
+                            <Field
+                                id={build(baseId, ids.FEEDBACK)}
+                                name="feedback"
+                                label={i18nHelp("feedback_title")}
+                                required={true}
+                                onKeyDown={(event) => {
+                                    if (event.key === "Enter") {
+                                        handleSubmit();
+                                    }
+                                }}
```
psarando, line 161: I don't think we want this key handler for a `FormMultilineTextField`.

The point of that field is to allow the user to enter multiple lines of text, separated by newlines, but I think this handler will submit the form when they try to start a new line in this field.

sri2k1us:
Merging this one. But reviews still welcome.

psarando (commented):
LGTM 👍 
I only had some fairly minor suggestions.

src/components/layout/CyVerseAppBar.js
sri2k1us: Its here - https://github.com/cyverse-de/sonora/blob/3efb99f8c190ae3eace17752e3fe153b9dae5afb/src/components/layout/CyVerseAppBar.js#L827
if am understanding your comment correctly.

src/constants.js
sri2k1us: I will save this for a future PR.

src/components/layout/CyVerseAppBar.js
psarando: Ah, I see it now. Thanks. I didn't see it in the diff, and I didn't think to check the unchanged code 👍

PR #271, opened 2021-01-26
Analyses Improvements

sri2k1us:
- Rename `Agave` to `HPC` and `Interactive` to `VICE`
- Refactor query-params for analyses listing
- Fix HT drill-down bug
- Set `My Analyses` as default filter (requested by Amanda / Sci Team)

### DIFF
--- a/src/components/analyses/listing/Listing.js
+++ b/src/components/analyses/listing/Listing.js
@@ -48,7 +48,6 @@ import TableView from "./TableView";
 import AnalysesToolbar from "../toolbar/Toolbar";
 
 import analysisStatus from "components/models/analysisStatus";
-import appType from "components/models/AppType";
 import NotificationCategory from "components/models/NotificationCategory";
 
 import { useConfig } from "contexts/config";
@@ -523,7 +522,7 @@ function Listing(props) {
                     ? globalConstants.SORT_DESCENDING
                     : globalConstants.SORT_ASCENDING,
                 property,
-                page,
+                0,
                 rowsPerPage,
                 permFilter,
                 typeFilter,
@@ -570,7 +569,7 @@ function Listing(props) {
                 rowsPerPage,
                 null,
                 null,
-                idFilter
+                null
             );
     };
 
--- a/src/components/analyses/listing/Listing.js
+++ b/src/components/analyses/listing/Listing.js
@@ -32,7 +32,7 @@ import {
 } from "serviceFacades/support";
 
 import { canShare, openInteractiveUrl } from "../utils";
-
+import globalConstants from "../../../constants";
 import ConfirmationDialog from "../../utils/ConfirmationDialog";
 import DEPagination from "../../utils/DEPagination";
 import withErrorAnnouncer from "../../utils/error/withErrorAnnouncer";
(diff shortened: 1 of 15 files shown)


src/components/analyses/listing/Listing.js
```diff
@@ -650,9 +648,14 @@ function Listing(props) {
...
-            data?.analyses.find((analysis) => analysis.id === id)
-        );
+        if (items) {
+            return items
+                .map((id) =>
+                    data?.analyses.find((analysis) => analysis.id === id)
+                )
+                .filter((item) => item !== null && item !== undefined);
+        }
+        return null;
```
ash3rz, line 658:
```suggestion
        return items?.map((id) =>
                data?.analyses.find((analysis) => analysis.id === id)
            )
            .filter((item) => item !== null && item !== undefined);
```
Also, not saying we shouldn't have the `filter` check, but seems like we should fix what's causing null values to get added to either `analyses` or `selected`.

src/components/apps/toolbar/AppNavigation.js
```diff
@@ -74,7 +74,8 @@ const useStyles = makeStyles((theme) => ({
 function getAppTypeFilters() {
     return Object.values(appType).map((type) => {
         return {
-            name: type,
+            name: type.display,
+            value: type.value,
```
ash3rz, line 78: I could be missing something, but it seems like it'd be simpler to either return `Object.values(appType)` and update any references to the `name` key to `display` or we could update the `AppType` model to use the `name` key directly instead of `display`.

ash3rz (approved):
I had a couple of suggestions, but LGTM :shipit:

ash3rz:
Also just noticed this PR needs to have a conflict resolved

src/components/analyses/listing/Listing.js
sri2k1us: This happening due to a race condition. When HT Drill down icon is clicked, that row is `selected` but in the background we start fetching the jobs associated with HT job. At this point the state is cleared and the `find` method can no longer find the analysis matching the selection and I think that results in `find` returning an array `[null]`. I don't know how to fix this yet.

src/components/apps/toolbar/AppNavigation.js
sri2k1us: Good point. I also found that the filter used `apps` and `analyses` listing is pretty much the same. So I extracted the `AppsTypeFilter` into its own component.

src/components/analyses/listing/Listing.js
```diff
@@ -542,45 +490,95 @@ function Listing(props) {
...
-        setPage(0);
+        const isAsc =
+            orderBy === property && order === globalConstants.SORT_ASCENDING;
+        onRouteToListing &&
+            onRouteToListing(
+                isAsc
+                    ? globalConstants.SORT_DESCENDING
+                    : globalConstants.SORT_ASCENDING,
+                property,
+                page,
```
psarando, line 526:
```suggestion
                0,
```

src/components/analyses/listing/Listing.js
```diff
@@ -542,45 +490,95 @@ function Listing(props) {
...
         setSelected([]);
+        onRouteToListing &&
+            onRouteToListing(
+                order,
+                orderBy,
+                0,
+                rowsPerPage,
+                null,
+                null,
+                idFilter
```
psarando, line 573: The `setIdFilter(null)` call was in this function before, so should `idFilter` be `null` in this `onRouteToListing` now?
```suggestion
                null
```
I'm not really sure about this handler now. It seems like it's setting `parentAnalyses` state, but then it's navigating the page away with `onRouteToListing`. So will the `parentAnalyses` state be preserved after the route navigation?

src/components/apps/toolbar/AppNavigation.js
```diff
@@ -71,14 +71,6 @@ const useStyles = makeStyles((theme) => ({
     },
 }));
 
-function getAppTypeFilters() {
-    return Object.values(appType).map((type) => {
```
psarando, line 75: There's a GitHub Actions warning now `'appType' is defined but never used` for the `import` on line 12.

psarando (approved):
I caught a couple of warnings, and had 1 question, otherwise LGTM 👍

src/components/analyses/listing/Listing.js
sri2k1us: At least thats my understanding. The components are not unmounted util you route to a different page.

src/components/apps/toolbar/AppNavigation.js
sri2k1us: 😩 I have had enough of this so I fixed the github actions lint checking at 10 pm so that it will fail on lint issues.

sri2k1us:
Thanks for the reviews.

PR #272, opened 2021-01-27
Trash & Refresh

sri2k1us:
- Large Screens
image
image

-iPhone 5
image


src/components/data/toolbar/Toolbar.js
```diff
@@ -188,6 +238,32 @@ function DataToolbar(props) {
...
+                        }}
+                        title={t("Delete")}
+                        contentText={t("deleteConfirmation")}
+                        onConfirm={() => {
+                            onDeleteSelected();
+                            setDeleteConfirmOpen(false);
+                        }}
+                    />
+                    <ConfirmationDialog
+                        baseId={baseId}
```
ash3rz, line 255: In this context, `baseId` is actually the ID for the entire dialog.  So in this case, this dialog and the one above should have their own, new IDs, not the same ID as the toolbar. 
Maybe the `baseId` prop in `ConfirmationDialog` can be renamed to `dialogId` to make it less confusing in another PR.

src/components/data/toolbar/TrashMenu.js
```diff
@@ -0,0 +1,98 @@
...
+    MenuItem,
+} from "@material-ui/core";
+
+import {
+    ArrowDropDown,
+    Delete,
+    DeleteForever,
+    RestoreFromTrash,
+    DeleteSweep,
+} from "@material-ui/icons";
```
ash3rz, line 30: GitHub Actions complaining about a handful of different unused imports in this file, FYI

src/components/data/toolbar/TrashMenu.js
```diff
@@ -0,0 +1,98 @@
...
+                onClick={onTrashMenuClick}
+                aria-haspopup={true}
+                aria-controls={baseId}
+                startIcon={<Delete />}
+                endIcon={<ArrowDropDown />}
+            >
+                {t("trash")}
+            </Button>
+            <Menu
+                id={(baseId, "trashBtnMenu")}
```
ash3rz, line 74: Maybe `trashBtnMenu` and `trashButton` above could move to an IDs file?

src/components/data/toolbar/TrashMenu.js
```diff
@@ -0,0 +1,98 @@
...
+            <Button
+                id={build(baseId, "trashButton")}
+                variant="outlined"
+                size="small"
+                disableElevation
+                color="primary"
+                className={classes.button}
+                onClick={onTrashMenuClick}
+                aria-haspopup={true}
+                aria-controls={baseId}
```
ash3rz, line 67: I think `baseId` in this context is the toolbar ID?  I think `aria-controls` is meant to have the ID of the element this opens/controls.  In this case, this should be the ID of the `Menu` below.

src/components/data/toolbar/TrashMenu.js
```diff
@@ -0,0 +1,98 @@
...
+                endIcon={<ArrowDropDown />}
+            >
+                {t("trash")}
+            </Button>
+            <Menu
+                id={(baseId, "trashBtnMenu")}
+                anchorEl={trashAnchor}
+                open={Boolean(trashAnchor)}
+                onClose={onTrashClose}
+                getContentAnchorEl={null}
```
ash3rz, line 78: I don't see this `getContentAnchorEl` prop listed in MUI's `Menu` API.  This might be a typo?

src/components/data/toolbar/UploadMenuBtn.js
```diff
@@ -16,7 +16,7 @@ import styles from "../styles";
 import { processSelectedFiles, trackUpload } from "../../uploads/UploadDrop";
 import { useUploadTrackingDispatch } from "../../../contexts/uploadTracking";
 
-import { useTranslation } from "react-i18next";
+import { useTranslation } from "i18n";
```
ash3rz, line 19: 👍

src/serviceFacades/filesystem.js
```diff
@@ -286,3 +286,18 @@ export const pathListCreator = ({
         params,
     });
 };
+
+export const restore = (paths) => {
+    return callApi({
+        endpoint: "/api/filesystem/restore",
+        method: "POST",
+        body: [paths],
```
ash3rz, line 294: Looking at the [API](https://cyverse-de.github.io/api/endpoints/filesystem/restore.html#restoring-a-file-or-directory-from-a-users-trash), the body is `{ paths: [paths] }`.  Does this work too?

ash3rz (commented):
LGTM overall 👍 I think I caught a few small issues though and potentially a bug with an API call.  Also this branch seems to have some conflicts at the moment.

src/serviceFacades/filesystem.js
sri2k1us: Actually it works the way as it is (although am not sure why it works) but to simplify it can be just `body: paths`.

src/components/data/listing/Listing.js
```diff
@@ -355,6 +384,21 @@ function Listing(props) {
...
+        setSelected([]);
+    };
+
+    const onRestoreSelected = (resourceId) => {
+        const items = resourceId ? [resourceId] : null;
+        const paths = getSelectedResources(items).map(
+            (resource) => resource.path
+        );
+        doRestore({ paths });
+        setSelected([]);
```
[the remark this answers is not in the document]
sri2k1us: Not really. `resourceId` is an optional parameter. So if it is `null` or `undefined`, `getSelectedResources` will get the resourceIds from `selected` state.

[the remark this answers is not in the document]
ash3rz: Ah right, I forgot it was optional 👍

src/components/data/toolbar/TrashMenu.js
sri2k1us: Tthat prop comes `Popover` without that we will get the following warning in console.
`Material-UI: You can not change the default `anchorOrigin.vertical` value 
when also providing the `getContentAnchorEl` prop to the popover component.
Only use one of the two props.
Set `getContentAnchorEl` to `null | undefined` or leave `anchorOrigin.vertical` unchanged.`

ash3rz (approved):
:shipit:

src/components/data/utils.js
```diff
@@ -178,6 +178,14 @@ const useSelectorDefaultFolderPath = () => {
     );
 };
 
+const isPathInTrash = (path, trash_path) => {
+    if (!path || !trash_path) {
+        return false;
+    } else {
+        return path?.startsWith(trash_path);
+    }
+};
```
psarando, line 187: Minor style suggestion, so feel free to skip it:
```suggestion
const isPathInTrash = (path, trash_path) =>
    (path && trash_path && path.startsWith(trash_path));
```

src/components/data/toolbar/Toolbar.js
```diff
@@ -66,19 +71,31 @@ function DataToolbar(props) {
...
+    const [emptyTrashConfirmOpen, setEmptyTrashConfirmOpen] = useState(false);
     const selectedResources = getSelectedResources();
     const canShare = isOwner(selectedResources);
     const theme = useTheme();
     const isSmall = useMediaQuery(theme.breakpoints.down("sm"));
+
+    const inTrash = isPathInTrash(path, config?.irods?.trash_path);
+    const uploadEnabled = !inTrash && isWritable(permission);
+    const sharingEnabled = !inTrash && canShare;
+    const bagEnabled = !inTrash && selected && selected.length > 0;
```
psarando, line 94: 👍

psarando (approved):
Tidy! ✨ 
LGTM :shipit: 👍

sri2k1us:
Thanks for the reviews.

PR #273, opened 2021-01-28
Use svgr-generated SVG icons for sidebar analyses and data

ianmcorvidae:
Sort of a proof of concept. I'm setting the `font-size` manually, but we could also use the MUI classes, probably, instead. Using `--icon` with svgr sets the width/height to `1em` which is the other part (other than font-size) for getting them the right size.


sri2k1us (approved):
image
image

LGTM in-terms of the concept of using svgs. But is it just me or the analyses icon in mobile view looks bigger than the others ?

ianmcorvidae:
Hm, possibly. I'll maybe compare it to the help icon and see what's different there.

ianmcorvidae:
It is slightly larger, I think because our SVG extends to the edge of the canvas while the mui one doesn't. I did a simple transform in Inkscape shrinking the thing from 28x28 to 24x24, but keeping the viewBox the same, and that seems to look fairly nice. Looking at the help icon, it's a 20x20 circle inside a 24x24 canvas, so I might try to duplicate that.

ianmcorvidae:
Screenshot_2021-01-28 Discovery Environment

amended analyses icon, now with the same metrics as the help icon below it

sri2k1us:
> Screenshot_2021-01-28 Discovery Environment
> 
> amended analyses icon, now with the same metrics as the help icon below it

Nice 👍🏽

ianmcorvidae:
I'll go ahead and merge this so it can be built on top of. Easy enough to change things.

PR #275, opened 2021-02-01
Admin DOI Requests Listing

sri2k1us:
- Request status update.
- View Folder details.
- Refactoring Layout code.

- Large Screen

image

image

image

- iPhone 5
image


src/components/doi/UpdateRequestDialog.js
```diff
@@ -0,0 +1,237 @@
...
+                                                        key={index}
+                                                        value={status}
+                                                        id={baseId}
+                                                    >
+                                                        {status}
+                                                    </MenuItem>
+                                                ))}
+                                            </FormSelectField>
+                                        )}
+                                    </Field>
```
psarando, line 206: I'm not sure I understand why this `Field` is using the `FormSelectField` this way.

The `FormSelectField` was implemented so that it can be used like this:
https://github.com/cyverse-de/sonora/blob/60d5cb0370cbd33b045df40d3fc9e0ed7aecea76/src/components/apps/launch/ResourceRequirements.js#L113-L124
or like this:
https://github.com/cyverse-de/sonora/blob/60d5cb0370cbd33b045df40d3fc9e0ed7aecea76/src/components/tools/edit/ToolRestrictions.js#L123-L143

So there shouldn't be a need to manually call `onChange` in the select field. For example:
```suggestion
                                    <Field
                                        id={build(baseId, ids.STATUS)}
                                        name="status"
                                        label={t("status")}
                                        required
                                        validate={(value) =>
                                            nonEmptyField(value, i18nUtil)
                                        }
                                        size="small"
                                        component={FormSelectField}
                                    >
                                                {Object.values(
                                                    RequestStatus
                                                ).map((status, index) => (
                                                    <MenuItem
                                                        key={index}
                                                        value={status}
                                                        id={baseId}
                                                    >
                                                        {status}
                                                    </MenuItem>
                                                ))}
                                    </Field>
```

Alternatively, we can also use an outlined `FormTextField` with a `select` prop instead (I actually think its UX is a little nicer then the `Select` field):
```suggestion
                                    <Field
                                        id={build(baseId, ids.STATUS)}
                                        name="status"
                                        label={t("status")}
                                        required
                                        validate={(value) =>
                                            nonEmptyField(value, i18nUtil)
                                        }
                                        component={FormTextField}
                                        select={true}
                                        variant="outlined"
                                        size="small"
                                    >
                                                {Object.values(
                                                    RequestStatus
                                                ).map((status, index) => (
                                                    <MenuItem
                                                        key={index}
                                                        value={status}
                                                        id={baseId}
                                                    >
                                                        {status}
                                                    </MenuItem>
                                                ))}
                                    </Field>
```

By the way, the `type` field in the `EditTool` form can be refactored in a similar way.

psarando (commented):
LGTM 👍 🎉 

But I think I caught a rebase issue with a missing Teams nav button.
Otherwise I just had some fairly minor cleanup suggestions ✨

src/components/doi/TableView.js
```diff
@@ -0,0 +1,210 @@
...
+                                                )}
+                                            </Typography>
+                                        </TableCell>
+                                        <TableCell>
+                                            <Typography variant="body2">
+                                                {formatDate(request.type)}
+                                            </Typography>
+                                        </TableCell>
+                                        <TableCell
+                                            id={build(rowId, ids.SUPPORT_CELL)}
```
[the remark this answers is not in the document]
sri2k1us: Well it is kinda there but not there. It is `undefined` 😂 . I will go ahead add ids.

src/components/doi/UpdateRequestDialog.js
sri2k1us: Yup. I took it from https://github.com/cyverse-de/ui/blob/06c9fa859c8a366de30a8afa201231bee333eb50/react-components/src/tools/edit/EditTool.js#L178. But Agree `FormTextField` with `select` variant is nice. Will update it.

psarando (approved):
LGTM 👍 🎉

sri2k1us:
Thanks the review. Am gonna merge this one but reviews welcome.

PR #276, opened 2021-02-01
Show Manage Tools button in `small` screen.

sri2k1us:
- Large Screen
image
- Small Screen
image
- Mobile
image

### DIFF
--- a/src/components/apps/ids.js
+++ b/src/components/apps/ids.js
@@ -40,6 +40,7 @@ export default {
     ADD_TO_BAG_MENU_ITEM: "addToBag",
     TOOLS_INFO_TAB: "toolsInfoTab",
     TOOLS_BTN: "manageToolsButton",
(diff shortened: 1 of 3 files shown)


ianmcorvidae (approved):
Seems reasonable to me. I guess the summary here is changing `smDown` -> `xsDown` makes the button visible in `small`, and removing the `isMobile` check also ensures that the button always appears any time the dot menu does?

sri2k1us:
> Seems reasonable to me. I guess the summary here is changing `smDown` -> `xsDown` makes the button visible in `small`, and removing the `isMobile` check also ensures that the button always appears any time the dot menu does?

I couldn't have summarized this change any better. 🎉

ash3rz (approved):
:shipit:

sri2k1us:
Thanks for the reviews.

PR #277, opened 2021-02-02
Add mathematica syntax highlighting

ianmcorvidae:
Silly and small. Was looking at a bundle analysis and mathematica is the second-largest language in the highlight.js code, so on a whim I decided to see if Tika detects it, and it does. So I added it here. Now mathematica files have syntax highlighting. Woo.


sri2k1us (approved):
LGTM :shipit:

ianmcorvidae:
Cool, thanks for the review. Possibly will ask the science team if there's other languages we should do this with that aren't difficult (re: my messages in chat the other day), but this seems good for now!

PR #278, opened 2021-02-02
Remove react-intl.

sri2k1us:
- SSIA


ianmcorvidae:
At least as far as bundle size, looks good. 258kB first load JS to 245kB, and I don't see react-intl anywhere (presumably what's left in ui-lib is getting tree-shaken away.

src/components/analyses/details/InfoPanel.js
```diff
@@ -142,6 +143,8 @@ function Updates(props) {
                 columnData={columns}
                 rowsInPage={updates.length}
                 baseId={baseId}
+                selectAllLabel={i18nCommon("selectAllLabel")}
+                sortLabel={i18nCommon("sortLabel")}
             />
```
sri2k1us, line 148: May be we could have another component in `Sonora` that will pass `selectAllLabel` and `sortLabel` prop by default to `EnhancedTableHead`.

ianmcorvidae: That would make sense to me, to save some repetition. I imagine it'd pretty much just be `export default (props) => <EnhancedTableHead selectAllLabel={t("selectAllLabel")} sortLabel={t("sortLabel")} {...props} />`, easily thrown into a file somewhere or other.

ash3rz (approved):
LGTM, thanks for tackling this 🎉  :shipit: 
One thing I didn't see was any page updates for `namespacesRequired` - not sure if those pages already had everything set up or not.
Also not sure if there should also be a `package-lock.json` update with this?  Not important, just figured if we removed a dependency that it would have updates.

ianmcorvidae:
It looks like it probably should change package-lock, yeah. I think running `npm uninstall react-intl` will update the file appropriately.

sri2k1us:
> LGTM, thanks for tackling this 🎉 :shipit:
> One thing I didn't see was any page updates for `namespacesRequired` - not sure if those pages already had everything set up or not.
> Also not sure if there should also be a `package-lock.json` update with this? Not important, just figured if we removed a dependency that it would have updates.

Good point. Let me update the lock file. As far as pages are concerned, I haven't checked the individual pages. But should be able track them with warnings and fix it easily.

sri2k1us:
Thanks for reviews.

PR #279, opened 2021-02-03
CORE-1212 Add Metadata Template Viewer and Listing

psarando:
This PR will import and update the Metadata Template Viewer dialog from the cyverse-de/ui repo, and adds a Metadata Template listing dialog (previously only in the GXT code).

This will also add the 3rd party endpoint handlers for the `Ontology Lookup Service` and `Unified Astronomy Thesaurus` services, used by the corresponding metadata template search fields.

Metadata Template listing dialog:
Metadata Template listing dialog

Mobile view:
Metadata Template listing dialog - mobile

---

If a Template has a `description` defined, then the listing will include the `i` info icon-button next to the template download button. Select the `i` info button will show the template's description in a popover:
Metadata Template description popover

Mobile view:
Metadata Template description popover - mobile

---

The Metadata Template dialog looks essentially the same as in the DE, except cleaned up or standardized for Sonora: no more App bar in the dialog header (it's a `DEDialog` now) and the dialog's action buttons are now `Cancel` and `Save`:
Metadata Template dialog

Mobile view:
Metadata Template dialog - mobile

The OLS and UAT search fields are still using the old Autocomplete field from `ui-lib`:
Ontology Lookup Service search field

They should be updated with the newer `@material-ui/lab/Autocomplete` field used by Sonora, but that will require refactoring the `FormSearchField` in `ui-lib`.


src/components/metadata/templates/Listing.js
```diff
@@ -0,0 +1,275 @@
...
+            <TemplateInfoDisplay
+                templateInfo={templateInfo}
+                anchorEl={templateInfoAnchor}
+                onClose={() => setTemplateInfoAnchor(null)}
+            />
+        </DEDialog>
+    );
+};
+
+export default MetadataTemplateListing;
```
sri2k1us, line 275: I prefer this listing in a `List` view rather than in a table. Then we can eliminate an additional click for `description`(`i` button) and put description as `secondary` text prop and `download` action as `secondary` action. See examples here - https://material-ui.com/components/lists/
Regardless the buttons needs in a theme color rather than `black`.

public/static/locales/en/metadata.json
```diff
@@ -10,6 +10,8 @@
     "attrTypeLabel": "Type",
     "avus": "AVUs",
     "back": "Back",
+    "confirmDiscardChangesDialogHeader": "Discard Changes?",
+    "confirmDiscardChangesDialogMsg": "You have unsaved changes. Do you wish to continue without saving?",
```
sri2k1us, line 14: When I trying this PR locally, I ran into confusion here with the generic `Ok` and `Cancel` action button. We should change this to prompt the user for `saving` the changes rather than `discarding`. 
`"confirmSaveChangesDialogHeader": "Save Changes?",
"confirmSaveChangesDialogMsg": "You have unsaved changes. Do you wish to save now?"`
and the dialog options should be `Save`, `Discard` and `Cancel` rather than generic `OK` and `Cancel`.

sri2k1us: I don't know if we are doing this consistently throughout Sonora, if not we should start doing this (i.e providing specific actions instead of generic `OK` and `Cancel`).

sri2k1us (approved):
LGTM 🎉 . I had some comments on UX issues.

sri2k1us:
Ohh forgot to mention, we have been using `text` variant buttons for actions in dialog. I see all the dialogs in this PR use `contained` variant buttons. That needs to changed as well.

public/static/locales/en/metadata.json
psarando: I seem to remember having this discussion before where we actually reached the opposite conclusion. I think in the old DE we had 3 buttons like that for some dialogs, but we decided it was too much cognitive load on the user, and we decided on only 1 or 2 buttons for cancel/confirm or cancel/ok instead. I think this was also a recommendation from the Material spec. I will see if I can find a reference to that discussion again.

sri2k1us: Ahh I do remember now something about having 3 buttons. But I still recommending having actions specific buttons instead of generic `Ok / Cancel`.

psarando: I couldn't find where we had that discussion, but I did find on the Material spec they recommend a max of 2 buttons: https://material.io/components/dialogs#actions (under the `Number of actions` section).

I also agree the button text could be more specific 👍
That's recommended by the Material spec as well (https://material.io/components/dialogs#alert-dialog).

I think we should stick with the `Cancel` / `Discard` buttons, though, because this confirmation dialog only shows when the user is trying to close the template without saving, so we're confirming that choice. Otherwise they should `Cancel` the discard action and specifically select the `Save` button (which is an action that can add or modify many AVUs at a time).

sri2k1us: Sounds good to me 👍🏽

src/server/api/metadata.js
```diff
@@ -7,15 +7,49 @@
...
 
 import * as auth from "../auth";
+import { olsURL, uatURL } from "../configuration";
 import logger from "../logging";
 
+import { handler as externalHandler } from "./external";
 import { handler as terrainHandler } from "./terrain";
+import { metadataTemplateCSVhandler } from "./downloads";
 
 export default function notificationsRouter() {
```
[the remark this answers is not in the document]
psarando: Oops, looks like we all missed this in #238 😆

public/static/locales/en/metadata.json
psarando: I also updated the content text, so the `Discard` button with dialog's message is hopefully clearer:
`You have unsaved changes. Do you wish to continue and discard these changes?`

src/components/metadata/templates/Listing.js
psarando: Good idea. I think the `List` looks better:
Select Template dialog

Mobile:
Select Template dialog on mobile

psarando:
I think I've addressed all PR comments, so I'll go ahead and merge this now, but additional reviews are still welcome 🙇 

Side note: the Metadata Template UX was one of the first (large) components we built with React and MUI, so I'm sure there is still room for improvement in a future PR. Perhaps lessons learned in the new App Editor could be applied to Metadata Templates in the future as well.

ash3rz (commented):
LGTM :shipit:

PR #281, opened 2021-02-06
CORE-1225 Fix analysis submissions with empty file/folder inputs

psarando:
Currently the relaunch endpoint will return an error for submissions with empty strings for optional file/folder inputs.

This will fix the app launch form to no longer submit empty strings for file and folder inputs.


psarando:
I'm going to merge this right away, but reviews are still welcome 🙇

PR #282, opened 2021-02-08
search whitespace issue

flynn1999:
Added `.trim` for the bug of search items not showing up if input had whitespaces near it

### DIFF
--- a/src/components/notifications/ids.js
+++ b/src/components/notifications/ids.js
@@ -25,5 +25,5 @@ export default {
     STATUS: "requestStatusColumn",
     TIME_STAMP: "time-stamp",
     TOOLBAR: "toolbar",
-    VIEW_ALL_NOTIFICATIONS: "view-all-notificaitons",
+    VIEW_ALL_NOTIFICATIONS: "view-all-notifications",
 };
--- a/src/components/search/GlobalSearchField.js
+++ b/src/components/search/GlobalSearchField.js
@@ -580,7 +580,7 @@ function GlobalSearchField(props) {
             setTeamSearchQueryEnabled(false);
             setOptions([]);
         }
-        setSearchTerm(value.trim());
+        setSearchTerm(value?.trim());
     };
 
     const handleFilterChange = (event) => {
--- a/src/components/search/GlobalSearchField.js
+++ b/src/components/search/GlobalSearchField.js
@@ -580,7 +580,7 @@ function GlobalSearchField(props) {
             setTeamSearchQueryEnabled(false);
             setOptions([]);
         }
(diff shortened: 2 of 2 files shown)


src/components/notifications/ids.js
```diff
@@ -25,5 +25,5 @@ export default {
     STATUS: "requestStatusColumn",
     TIME_STAMP: "time-stamp",
     TOOLBAR: "toolbar",
-    VIEW_ALL_NOTIFICATIONS: "view-all-notificaitons",
+    VIEW_ALL_NOTIFICATIONS: "view-all-notifications",
```
ash3rz, line 28: 😆 thank you for fixing this typo

src/components/search/GlobalSearchField.js
```diff
@@ -580,7 +580,7 @@ function GlobalSearchField(props) {
             setTeamSearchQueryEnabled(false);
             setOptions([]);
         }
-        setSearchTerm(value);
+        setSearchTerm(value.trim());
```
sri2k1us, line 583: This works mostly but you might want to cover the case where `value` might be `null` or `undefined` ?

sri2k1us: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining might help.

ianmcorvidae:
So, this is good and should happen, but it would be nice if _in addition_, the API did the same sort of trim. I forget if you're interested in Clojure/have done any of that, but, if you are, here's the outline of how that'd happen (in my opinion, anyway):

1.) https://clojuredocs.org/clojure.string/trim

2.) https://github.com/cyverse-de/kameleon/blob/master/src/kameleon/util/search.clj

3.) https://github.com/cyverse-de/apps/blob/master/src/apps/persistence/app_search.clj

Use 1 in the file at 2, then get one of us to do a release of the `kameleon` library after that PR is done. You'll then want to update the version of `kameleon` in the apps service (edit the `project.clj` file), so that the code in 3 will do the appropriate trimming.

src/components/search/GlobalSearchField.js
flynn1999: Thank you, I'll remember to use it for future functions

flynn1999:
> https://github.com/cyverse-de/apps/blob/master/src/apps/persistence/app_search.clj

I've been meaning to look more into Clojure and did think a better solution would be a fix from the API side. I'll try this. Thank you for the info!

ash3rz (approved):
LGTM :shipit:

sri2k1us (approved):
🎉 LGTM

PR #283, opened 2021-02-09
View doi metadata

sri2k1us:
- on top of #275.
- SSIA

### DIFF
--- a/src/pages/admin/doi.js
+++ b/src/pages/admin/doi.js
@@ -41,13 +41,11 @@ export default function Doi() {
         [router]
     );
 
-    const onRouteMetadataView = useCallback(
-        (path, type, resourceId) => {
+    const onRouteToMetadataView = useCallback(
+        (path) => {
             router.push({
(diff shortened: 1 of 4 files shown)


src/pages/admin/doi.js
```diff
@@ -40,6 +41,20 @@ export default function Doi() {
...
     );
 
+    const onRouteMetadataView = useCallback(
+        (path, type, resourceId) => {
+            router.push({
+                pathname: `/${NavigationConstants.DATA}/${constants.DATA_STORE_STORAGE_ID}${path}`,
+                query: {
+                    type,
+                    resourceId,
+                    view: NavigationParams.VIEW.METADATA,
```
psarando, line 51: Only the `view` query param should be required for viewing metadata, so the `type` and `resourceId` could be removed from this callback.

psarando (approved):
LGTM ✨ :shipit: 👍

sri2k1us:
Thanks for reviews. More reviews welcome.

PR #284, opened 2021-02-09
Create doi

sri2k1us:
- On top of #283 
- Allow Admin to create DOI.

PR #288, opened 2021-02-10
Request doi

sri2k1us:
- On top of #284 
image

image


src/components/data/listing/Listing.js
```diff
@@ -561,6 +589,53 @@ function Listing(props) {
...
+                    </>
+                }
+            >
+                <Trans
+                    t={t}
+                    i18nKey="requestDOIPrompt"
+                    components={{
+                        manual: <ExternalLink href={constants.DOI_GUIDE} />,
+                    }}
+                />{" "}
```
ash3rz, line 627: Would removing the `{" "}` and replacing it with `</br>` put the next message on a new line?  
Since these two messages change perspectives (first message is in the second person with using "you" and the second message is in the first person with using "I"), it'd be nice if these two messages didn't look like they were part of the same paragraph.

src/components/data/toolbar/DataDotMenu.js
```diff
@@ -88,7 +90,7 @@ function DataDotMenu(props) {
         : null;
     const deleteMiEnabled = !isSelectionEmpty && isOwner(selectedResources);
     const metadataMiEnabled = selected?.length === 1;
-
+    const requestDoiEnabled = isOwner(selectedResources);
```
ash3rz, line 93: I could be wrong, not the expert on this, but I think DOI requests also have to be done on folders and not files?

ash3rz (approved):
LGTM 👍 I had some minor comments and may have caught something with respect to when the DOI button should be enabled, though not 100% sure.

public/static/locales/en/data.json
```diff
@@ -71,6 +72,7 @@
     "modifyTagsError": "Unable to update tags. Please try again.",
     "myData": "My Data",
     "name": "Name",
+    "needDoi": "I need DOI",
```
psarando, line 75:
```suggestion
    "needDoi": "I need a DOI",
```

public/static/locales/en/data.json
```diff
@@ -93,6 +95,9 @@
     "publicLinkWarning": "Note: Anyone with access to a file's URL can access that file.",
     "publicLinkError": "Unable to get public link(s)",
     "read": "Read",
+    "requestDOI": "Request DOI",
+    "requestDOIPrompt": "Please read <manual>DOI Manual</manual> and check if you meet the requirements for submitting a DOI request.",
+    "requestDOIAgreement": "By clicking <b>I need DOI</b>, I have read and agree to the terms of the <agreement> Data Commons </agreement>",
```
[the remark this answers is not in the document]
psarando:
```suggestion
    "requestDOIAgreement": "By clicking <b>I a need DOI</b>, I have read and agree to the terms of the <agreement> Data Commons </agreement>.",
```

src/components/data/toolbar/DataDotMenu.js
psarando: Yes, this should also ensure that only 1 folder selected.

psarando (changes requested):
LGTM, except the `requestDoiEnabled` flag should be updated 👍

src/components/data/toolbar/DataDotMenu.js
```diff
@@ -88,7 +90,8 @@ function DataDotMenu(props) {
         : null;
     const deleteMiEnabled = !isSelectionEmpty && isOwner(selectedResources);
     const metadataMiEnabled = selected?.length === 1;
-
+    const requestDoiEnabled =
+        isOwner(selectedResources) && containsFolders(selectedResources);
```
psarando, line 94: Also `selected?.length === 1`?

sri2k1us: I thought about that but then `Request DOI` menu item is inside `metadataMiEnabled` check so I didn't add that check 🤷🏽‍♂️

psarando: Ah, right, I missed that. That will work then.

psarando (approved):
LGTM :shipit: 👍

sri2k1us:
Thanks for the reviews.

PR #289, opened 2021-02-12
Low-hanging-fruit syntax highlighting

ianmcorvidae:
This adds a whole bunch of syntax highlighting languages. My other PR, needing UX work, isn't going out super immediately, but these are the easy ones to add (and we'll want both things anyway). I changed how the mode detection code works since now there's a lot more in the list and the switch seemed like it would be quite unwieldy. I haven't tested all of these, but I've tested at least one in each category (exact mimetype -> mode, exact except for the x- prefix, and completely custom mapping).

### DIFF
--- a/src/components/models/MimeTypes.js
+++ b/src/components/models/MimeTypes.js
@@ -2,14 +2,17 @@ const mimeTypes = {
     PNG: "png",
     GIF: "gif",
     JPEG: "jpeg",
+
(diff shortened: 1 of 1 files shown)


ash3rz (approved):
:shipit:

ianmcorvidae:
Cool, thanks for the review.

PR #290, opened 2021-02-12
Disable news and events in the dashboard.

johnworth:
The RSS feeds on the webiste haven't updated recently, so these sections are very out of date.

They're just commented out for now, since hopefully we'll be able to bring them back in the future.


ash3rz (approved):
:shipit:

sri2k1us (approved):
:shipit:

PR #291, opened 2021-02-12
Rename and more

sri2k1us:
- Allow users to rename file or folder.
- 
image

image

- Fix Create New Folder validation issue.
- Fix small `Login` link in banner for FF.
-Before:
image
-After:
image


src/components/data/toolbar/DataDotMenu.js
```diff
@@ -369,6 +408,17 @@ function DataDotMenu(props) {
...
             />
+
+            {selectedResources && (
+                <RenameDialog
+                    path={selectedResources[0]?.path}
+                    open={renameDlgOpen}
+                    onClose={onRenameDlgClose}
+                    onRenamed={() => {
+                        onRenameDlgClose();
+                    }}
```
psarando, line 419:
```suggestion
                    onRenamed={onRenameDlgClose}
```

src/components/data/RenameDialog.js
```diff
@@ -0,0 +1,163 @@
...
+                                        type="submit"
+                                        onClick={handleSubmit}
+                                    >
+                                        {t("rename")}
+                                    </Button>
+                                </>
+                            }
+                        >
+                            <Typography
+                                classes={{ root: classes.bottomPadding }}
```
psarando, line 118: Is this different from setting `gutterBottom={true}`?
https://material-ui.com/api/typography/#props

psarando (approved):
LGTM :shipit: 👍

src/components/data/RenameDialog.js
sri2k1us: Hmm entirely sure but `gutterBottom` is for margins and this classes is setting `padding`. I copy-pasted from `CreateFolderDialog`.

sri2k1us:
Thanks for the reivew. Additional reviews welcome.

src/components/data/RenameDialog.js
psarando: 👌 Well I just messed around with it in Storybook real quick, and they seem to have the same effect to me 🤷

sri2k1us: May be we can get rid of `custom` style and replace it `gutterBottom` it in another PR from these 2 dialog 👍🏽

ash3rz: Ah neat!  I either didn't know or forgot about `gutterBottom` 😄 I'll keep it in mind going forward

ash3rz (commented):
LGTM :shipit: 
Not critical, but looks like the Rename option didn't make it into the row dot menu.

PR #294, opened 2021-02-13
Merge namespacesRequired into pageProps

ianmcorvidae:
I think this prevents the "prop title doesn't match" thing that's been a bit maddening for me for a little bit here. This makes sure that a.) `bags` are in the default list, since they're used in the AppBar and appear everywhere, and b.) that those are merged into the `pageProps`, not isolated out somewhere separate that I'm not sure is used.

Would appreciate someone checking my work that this is, in fact, the right thing to do. next-i18next doesn't seem to be very well-documented, but I might well be missing something.

### DIFF
--- a/src/pages/_app.js
+++ b/src/pages/_app.js
@@ -268,5 +268,15 @@ MyApp.getInitialProps = async ({ Component, ctx }) => {
         pageProps = await Component.getInitialProps(ctx);
     }
 
-    return { pageProps, namespacesRequired: ["common"] };
+    // merge the two always-needed namespaces into the pageProps namespacesRequired array
+    return {
+        pageProps: {
+            ...pageProps,
+            namespacesRequired: [
+                ...pageProps?.namespacesRequired,
+                "common",
+                "bags",
+            ],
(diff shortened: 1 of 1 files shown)


ash3rz (approved):
LGTM, and yes, the documentation is not great ☹️ 

One thing I did notice when checking this out is that I got a new error instead of the `title Use Bags` one from before:

```
Warning: Expected server HTML to contain a matching <div> in <div>.
    in div (created by Joyride)
    in Joyride (created by ForwardRef(LoadableComponent))
    in ForwardRef(LoadableComponent) (at ProductTour.js:38)
    in ProductTour (at AppBar.js:453)
    ...
```
I didn't see any mismatched divs in our code, so it might actually be coming from Joyride 😖

ianmcorvidae:
Yeah, I was seeing that with or without this change. Haven't dug into it, but my guess was indeed that it was coming from Joyride itself.

ianmcorvidae:
Anyway, I'll merge it for now. I think this is still better, anyway

PR #297, opened 2021-02-19
Prevent errors when pageProps.namespacesRequired isn't present

ianmcorvidae:
This fixes the `TypeError: undefined is not iterable` caused by my #294 fix. I can't get apps pages to load (so I can't create an analysis to relaunch), but I tested this with the settings page, which is another that doesn't have any required namespaces defined.

### DIFF
--- a/src/pages/_app.js
+++ b/src/pages/_app.js
@@ -273,7 +273,7 @@ MyApp.getInitialProps = async ({ Component, ctx }) => {
         pageProps: {
             ...pageProps,
             namespacesRequired: [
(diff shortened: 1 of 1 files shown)


psarando (approved):
Nice, I forgot about https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

LGTM 👍

ianmcorvidae:
Cool, thanks for the review. I'll merge it now, hopefully get things working.

PR #298, opened 2021-02-22
CORE-1284 Intercom events

sri2k1us:
Track user activity in DE using intercom events.

### DIFF
--- a/src/common/intercom.js
+++ b/src/common/intercom.js
@@ -71,6 +71,8 @@ const IntercomEvents = {
     DELETED_TEAM: "deleted team",
     REQUESTED_JOIN_TEAM: "request to join team",
     ENCOUNTERED_ERROR: "encountered error",
+    VIEWED_METADATA: "viewed metadata",
+    SUBMITTED_DOI_REQUEST: "submitted doi request",
 };
 
 export {
--- a/src/common/intercom.js
+++ b/src/common/intercom.js
@@ -36,7 +36,33 @@ function intercomShow() {
 }
 
 function trackIntercomEvent(event, metadata) {
-    console.log("track event");
+    console.log("track event: " + event + " " + JSON.stringify(metadata));
+    if (window.Intercom) {
+        if (metadata) {
+            window.Intercom("trackEvent", event, metadata);
+        } else {
+            window.Intercom("trackEvent", event);
+        }
+    }
 }
 
-export { intercomLogin, intercomLogout, intercomShow, trackIntercomEvent };
+const IntercomEvents = {
+    VIEWED_FOLDER: "viewed folder",
+    VIEWED_FILE: "viewed file",
+    VIEWED_APPS: "viewed apps",
+    VIEWED_ANALYSES: "viewed analyses",
+    VIEWED_TOOLS: "viewed tools",
+    VIEWED_NOTIFICATIONS: "viewed notifications",
+    LAUNCHED_JOB: "launched job",
+    VIEWED_HELP: "viewed help",
+    URL_IMPORT_SUBMITTED: "url import submitted",
+    FILE_UPLOADED: "file uploaded",
+};
+
+export {
+    intercomLogin,
+    intercomLogout,
+    intercomShow,
+    trackIntercomEvent,
+    IntercomEvents,
+};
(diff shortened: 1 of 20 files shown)


src/common/intercom.js
```diff
@@ -36,7 +36,33 @@ function intercomShow() {
...
+    VIEWED_FILE: "viewed file",
+    VIEWED_APPS: "viewed apps",
+    VIEWED_ANALYSES: "viewed analyses",
+    VIEWED_TOOLS: "viewed tools",
+    VIEWED_NOTIFICATIONS: "viewed notifications",
+    LAUNCHED_JOB: "launched job",
+    VIEWED_HELP: "viewed help",
+    URL_IMPORT_SUBMITTED: "url import submitted",
+    FILE_UPLOADED: "file uploaded",
+};
```
sri2k1us, line 60: This is not an exhaustive list of events. Suggestions welcome on what other events when tracked in intercom will be useful.

ash3rz: We have teams now, so we could add that.  
I wonder how difficult it would be to track when users are seeing the ErrorHandler or some of the other error components?  Not sure how straightforward that is since it looks like the pattern is to add these events after an API call.  One of the easier things might be to add a tracker when `showErrorAnnouncer` is called [here](https://github.com/cyverse-de/sonora/blob/master/src/components/utils/error/withErrorAnnouncer.js#L46) and we could add the `text` and `error` data perhaps.

src/components/apps/launch/index.js
```diff
@@ -36,6 +38,7 @@ const Launch = ({ app, launchError, loading }) => {
             onSuccess: (resp, { onSuccess }) => {
                 router.push(`/${NavigationConstants.ANALYSES}`);
                 onSuccess(resp);
+                trackIntercomEvent(IntercomEvents.LAUNCHED_JOB, resp);
```
ash3rz, line 41: I see a few of these are sending the response for tracking.  If the response body were large, does this noticeably affect performance?  I'm guessing not, but thought I'd ask

ash3rz (approved):
LGTM :shipit: 
I had a couple comments, but nothing major.

src/common/intercom.js
psarando: I'm wondering why we should only track events when the API returns a success response? Would it better to track the event whenever the user selects a related action in the UI, so we can see what they did no matter if the API returns a success or error response?

For example, maybe we can track the `VIEWED_APPS` event whenever the Apps icon is selected from the main nav bar, or whenever they change the listing filter; and the `VIEWED_FOLDER` event could trigger when they select a folder name from the data listing, or when selecting the "View Outputs" action in the analyses listing. In fact, those could be separate events, whereas we can only have 1 right now in the success callback which can't tell how the user got there in the first place 🤔

psarando: I'd also like to suggest events for viewing metadata, templates, and submitting a DOI request.

src/components/apps/launch/index.js
sri2k1us: I didn't see any noticeable performance issue. And when am tracking listings, I only log the query parameters and not the actual listing so I we are not sending large response bodies to intercom.

src/common/intercom.js
sri2k1us: I agree We can track when Apps Icon is clicked or when `View Output Folder` is clicked. But this could be lead to flood of events 🤔 ?  So we need to selective IMO. Also I would do this in addition to tracking  events on `success` or `error` response and not replace it. For example: When a job is launched, the response contains some useful information like `analysis id`.

sri2k1us: I think we should also track search related events.

sri2k1us: I think I have added some events to track based on the feedback. Tracking things that involve nextjs `Link` (e.g: Go to output folder) seems to challenging as we are allowed to write custom logic `onClick`.

psarando (approved):
Nice LGTM :shipit: 👍 

Let's all remember to check a user's events in intercom when we do support, so we can see which are useful and if any need tweaking.

sri2k1us:
Thanks for the reviews.

PR #299, opened 2021-03-03
Add new, non-admin VICE API calls for loading page

johnworth:
Requires the following PRs be merged:
* https://github.com/cyverse-de/terrain/pull/202
* https://github.com/cyverse-de/app-exposer/pull/18

Contains the following changes.

* Adds GET /vice/resources which returns a filtered VICE analysis listing for the logged in user.
* Adds GET /vice/:host/description which returns a VICE analysis listing entry for the analysis associated with :host. Access permission for the logged in user is checked.
* Adds GET /vice/:host/url-ready which returns whether or not the VICE analysis associated with :host is ready for access by browsers. Access permission for the logged in user is checked.

### DIFF
--- a/src/server/api/vice.js
+++ b/src/server/api/vice.js
@@ -148,7 +148,7 @@ export default function viceRouter() {
         })
     );
 
-    logger.info("adding the GET /vice/listing handler");
+    logger.info("adding the GET /vice/resources handler");
     api.get(
         "/vice/resources",
         auth.authnTokenMiddleware,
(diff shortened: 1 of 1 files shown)


src/server/api/vice.js
```diff
@@ -148,5 +148,35 @@ export default function viceRouter() {
         })
     );
 
+    logger.info("adding the GET /vice/listing handler");
```
ash3rz, line 151: Small thing, but the log doesn't match the actual path being called

ash3rz (approved):
LGTM :shipit:

src/server/api/vice.js
johnworth: D'oh, thanks.

PR #300, opened 2021-03-10
Vice Access request and errors

sri2k1us:
Limit Reached:
image

Request VICE Access:
image

Account Banned:
image

P.S: Admin feature to list and process VICE Access request will be in a separate PR.


src/components/apps/AppName.js
```diff
@@ -40,6 +100,50 @@ function AppName(props) {
                 />
             </Link>
         );
+    } else if (runErrorCodes?.length > 0) {
+        if (runErrorCodes[0] === ERROR_CODES.ERR_PERMISSION_NEEDED) {
```
ash3rz, line 104: Maybe `const runErrorCode = runErrorCodes[0];` could be added for some reuse?

src/components/apps/AppName.js
```diff
@@ -40,6 +100,50 @@ function AppName(props) {
...
+                        component="button"
+                        href="#"
+                        onClick={() => setAccessRequestDialogOpen(true)}
+                        color="primary"
+                    >
+                        {name}
+                    </MuiLink>
+                    <AccessRequestDialog
+                        open={accessRequestDialogOpen}
+                        baseId={baseDebugId}
```
ash3rz, line 117: This dialog and the `RunErrorDialog` should have their own unique IDs

src/components/apps/launch/index.js
```diff
@@ -78,31 +89,68 @@ const Launch = ({ app, launchError, loading }) => {
...
+                    <Button
+                        onClick={() => setAccessRequestDialogOpen(true)}
+                        color="primary"
+                        variant="contained"
+                    >
+                        {t("requestAccess")}
+                    </Button>
+                    <AccessRequestDialog
+                        open={accessRequestDialogOpen}
+                        baseId={baseId}
```
ash3rz, line 121: Same thing here with having a unique ID

src/components/apps/launch/index.js
```diff
@@ -78,31 +89,68 @@ const Launch = ({ app, launchError, loading }) => {
...
-            submitAnalysis={(submission, onSuccess, onError) => {
-                setSubmissionError(null);
-                submitAnalysisMutation({ submission, onSuccess, onError });
-            }}
-            saveQuickLaunch={(quickLaunch, onSuccess, onError) => {
-                setSubmissionError(null);
-                addQuickLaunchMutation({ quickLaunch, onSuccess, onError });
-            }}
-        />
+        <>
```
ash3rz, line 133: This extra fragment doesn't look to be necessary

public/static/locales/en/vice.json
```diff
@@ -0,0 +1,9 @@
+{
+    "accessRequestTitle": "Request access to VICE analyses (e.g: Jupyter, RStudio etc...)",
```
ash3rz, line 2: Picking a random spot - for these messages maybe we should pick a particular word to use.  Here we're saying `VICE analyses`, in another message it says `... not permitted to launch VICE apps`, and in one other it says `... currently running VICE jobs`.  We've had some discussions in Slack about how some users are confused by "analyses" in regards to VICE, so it's stuck in my head as something to think about 😄

ash3rz (approved):
LGTM :shipit: 
I just had some minor comments and a suggestion about the wording we're using in these dialogs.  From the screenshots, it looks like the `contact support` link is in a smaller font again compared to the rest.

src/components/vice/AccessRequestDialog.js
```diff
@@ -0,0 +1,136 @@
...
+
+import requestAccess from "serviceFacades/vice/accessRequest";
+import constants from "../../constants";
+
+import DEDialog from "components/utils/DEDialog";
+import { useConfig } from "contexts/config";
+import { Button, CircularProgress, Typography } from "@material-ui/core";
+import ErrorTypographyWithDialog from "components/utils/error/ErrorTypographyWithDialog";
+
+const MIN_USE_CASE_LENGTH = 60;
```
psarando, line 21: This should probably come from configs, so we (and 3rd party installs) can change it on the fly.

psarando (approved):
LGTM 👍 

I only had one suggestion in addition to what @ash3rz already mentioned.

public/static/locales/en/vice.json
sri2k1us: Good catch. You are right, we should be consistent in our messaging.

src/components/vice/AccessRequestDialog.js
sri2k1us: I can easily add this to configs but I don't see how this value will be changed frequently or vary in different environments / 3rd party installs 🤔 (unlike `concurrent job limit` which needs to be tweaked for resource constraints in different environment / installs) .  As a matter of fact, we don't put other form field validations rules in config.

psarando: Specifically this value seems somewhat arbitrary to me, and it's a value only high enough to force users asking for VICE access to give us a valid science use case for that access.

I can easily see this gatekeeping value not being high enough to filter out invalid use cases, and we may be asked to increase it. I can also easily see users complaining about a value that is too high, then we'll be asked to reduce it. I can also imagine a 3rd party install not wanting to require this text at all, and they'll want this to be `0`.

It seems a bit heavy to force a rebuild of the software for an arbitrary value like this, which could be changed on the fly by a config setting instead.

sri2k1us: I added this to config. IMO managing our growing number of configs is more painful than this occasional / rare builds. Also don't we do builds on a config change ? Asking for my own knowledge.

sri2k1us:
> LGTM :shipit:
> I just had some minor comments and a suggestion about the wording we're using in these dialogs. From the screenshots, it looks like the `contact support` link is in a smaller font again compared to the rest.

The `contact support` link is not inside `Typography` and do so breaks the `Link` 😩

sri2k1us:
Thanks for the reviews. I think I have addressed all the PR comments.

src/components/vice/AccessRequestDialog.js
psarando: That's a good question, and I'm not sure 😬 
If so, I would think that defeats the purpose of configs 🤔

PR #301, opened 2021-03-10
CORE-1309 Vice Loading page revamp

ash3rz:
Some things to note from a high level:
* The navigation to the loading page from notifications and the analysis page is temporary - the backend should handle directing the user to the loading page if the instance is not ready
* If a container (init container or pod container) has restarted 3 times, then we display an error to the user
* If the launch has been stuck on deployment/services/ingresses/config maps for a certain amount of time, then we present a warning to the user that it's taking longer than expected
* There's another timer for pods, similar to above.  This timer is based on whether there are any pods returned in the description endpoint (i.e. pods have been assigned to a host), not for how long the pods have been running.

Some failing apps in QA to try out: `firefly` and `3.1 dev GEA Leptin RNA Seq Lab`

I updated most of the screenshots below with Mariah's new VICE loading gif.  The rocket image is completely gone now.

Desktop, sample message:
image

Desktop, sample error:
image

Details:
image

Message to support:
image

Mobile view :
image


src/components/vice/loading/index.js
```diff
@@ -0,0 +1,322 @@
...
+            onSuccess: (resp) => setReady(resp.ready),
+            refetchInterval: 5000,
+        },
+    });
+
+    useEffect(() => {
+        setTimeoutError(null);
+
+        let timeout = null;
+        const timeoutMilliseconds = config?.vice?.deploymentTimeoutMs || 180000;
```
sri2k1us, line 75: Not a huge issue but this default value of `180000` could be set in `next.conf.js` itself.

sri2k1us (commented):
The code look ✨ but I have found a bug / edge case. So everything works as expected until you launch a VICE analysis. But once you `Cancel` the analysis (after the analysis is actually cancelled), and if you click on the `Running` notification again, we get the loading page again, but it is stuck at 0%. It will be nice to detect that the analysis is actually cancelled and show an appropriate message in that page.

ash3rz:
Yes - good catch on the text maybe being complicated for beginners.  I meant to send this to Tina/MMS to review the text and completely forgot.  I'll do that now.

psarando (approved):
LGTM 👍

johnworth (approved):
LGTM, :shipit:

ash3rz:
Thanks for the reviews!  I have a meeting with MMS and Tina this afternoon to demo this and go over the text displayed to users in more depth, so more updates will be incoming before merging.

ash3rz:
Thanks again for the reviews! I've updated the wording on a few things based on feedback from 411 so far.  If we get more feedback, we can add them to a new PR.
For now, we're going to keep in the commit that takes users directly to the loading page (as opposed to taking them to the access URL which then takes them to the loading page if the instance is not ready) until that part of the backend is ready.
As for notifying users in the loading page that their analysis was already cancelled/complete, that will come in a later PR too once we have an endpoint in terrain to check the status given an access url.

PR #303, opened 2021-03-22
VICE Access Request admin processing

sri2k1us:
- Admin Processing Page

image

- Set Concurrent Job limit for approved request
image

- Reject access request
image

### DIFF
--- a/public/static/locales/en/vice-admin.json
+++ b/public/static/locales/en/vice-admin.json
@@ -1,4 +1,6 @@
 {
+    "accessRequests": "Access Requests",
+    "showAllRequests": "Show All Requests",
     "accessRequestApprovedMsg": "Your request to access VICE apps is approved.",
     "accessRequestDeniedMsg": "We are unable approve your request to access VICE apps at this moment.",
     "accessRequestUpdateSuccess": "Request updated successfully.",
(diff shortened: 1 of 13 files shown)


psarando (approved):
LGTM 👍 
I only had a couple of minor style suggestions 🎉 ✨

public/static/locales/en/vice-admin.json
```diff
@@ -1,4 +1,10 @@
 {
+    "accessRequests": "Access Requests",
+    "showAllRequests": "Show All Requests",
+    "accessRequestApprovedMsg": "Your request to access VICE apps is approved.",
+    "accessRequestDeniedMsg": "We are unable approve your request to access VICE apps at this moment.",
```
ash3rz, line 5: I was going to comment on how this message might be a little misleading, but I don't think it's actually used anywhere? 🤔

src/components/vice/admin/accessRequests/TableView.js
```diff
@@ -0,0 +1,143 @@
...
+                    const original = row.original;
+                    return (
+                        <IconButton
+                            color={
+                                original?.status?.toLowerCase() ===
+                                ACCESS_REQUEST_REJECTED
+                                    ? "primary"
+                                    : "default"
+                            }
+                            onClick={() =>
```
ash3rz, line 108: Having ids on the approve and deny buttons would be helpful for testing

ash3rz (approved):
LGTM :shipit:

public/static/locales/en/vice-admin.json
sri2k1us: Good catch. I added a feature of admins to add a custom instead of this one.

PR #304, opened 2021-03-24
Core 1346 Move files / Folders

sri2k1us:
- Move Dialog in Large Screen

image

- iPhone 5

image


src/components/data/MoveDialog.js
```diff
@@ -0,0 +1,192 @@
...
+                                style={{
+                                    marginTop: theme.spacing(1),
+                                    marginBottom: theme.spacing(1),
+                                }}
+                            >
+                                {t("selectedFilesFolders")}
+                            </Typography>
+                            {(selectedResources === null ||
+                                selectedResources === undefined ||
+                                selectedResources?.length === 0) && (
```
ash3rz, line 127: I think `selectedResources?.length === 0` also covers the null and undefined check?

src/components/data/MoveDialog.js
```diff
@@ -0,0 +1,192 @@
...
+                                <Typography color="error">
+                                    {t("selectFileFolderPrompt")}
+                                </Typography>
+                            )}
+                            <List>
+                                {selectedResources?.map((resource) => {
+                                    return (
+                                        <>
```
ash3rz, line 135: Looks like this fragment isn't needed

src/components/data/MoveDialog.js
```diff
@@ -0,0 +1,192 @@
...
+                                            <ListItem
+                                                id={build(baseId, resource?.id)}
+                                                key={resource?.id}
+                                            >
+                                                <IconButton
+                                                    edge="end"
+                                                    aria-label="remove"
```
ash3rz, line 142: This aria-label could be i18n

src/components/data/SelectionDrawer.js
```diff
@@ -93,6 +101,68 @@ function SelectionToolbar(props) {
...
+                {!multiSelect &&
+                    acceptedType === ResourceTypes.FOLDER &&
+                    selectedTotal === 0 && (
+                        <>
+                            <Button
+                                startIcon={<HomeIcon />}
+                                id={build(
+                                    baseId,
+                                    ids.SELECTION_TOOLBAR,
+                                    "home"
```
ash3rz, line 130: This is minor, but this part of the ID could also go into the IDs file

src/components/data/listing/Listing.js
```diff
@@ -705,6 +712,18 @@ function Listing(props) {
...
+                path={path}
+                open={moveDlgOpen}
+                selectedResources={getSelectedResources()}
+                onClose={() => setMoveDlgOpen(false)}
+                onRemoveResource={(resource) => {
+                    const newSelected = selected?.filter(
+                        (sel) => sel !== resource?.id
+                    );
+                    setSelected(newSelected);
+                }}
```
ash3rz, line 725: 🤔 I might be missing something, but seems to me like we don't need to update the selections here in the listing if the user deselects something from the move dialog.

src/components/data/toolbar/DataDotMenu.js
```diff
@@ -160,14 +164,6 @@ function DataDotMenu(props) {
                                       <ListItemText primary={t("folder")} />
                                   </MenuItem>
                               ),
-                              renameEnabled && (
```
ash3rz, line 163: I might be missing something but is there a reason why the rename menu item was removed?

ash3rz (approved):
LGTM, I just had some minor comments 👍 

From the screenshots, one thing to consider is that it seems like we're mostly putting action buttons on the right.  In the Move Dialog, the buttons to remove selections is on the left.  Another thing is on mobile it could be possible to accidentally hit remove unintentionally.  Not sure what others think.
I also wonder how this would look if the gray circle background around the folder and file icons were removed.  At least next to the remove buttons, these icons start to look like they're buttons too, but that may just be me.

src/components/data/SelectionDrawer.js
```diff
@@ -93,6 +101,68 @@ function SelectionToolbar(props) {
...
+                            : theme.spacing(1),
+                    }}
+                    size="small"
+                >
+                    {t("cancel")}
+                </Button>
+
+                {!multiSelect &&
+                    acceptedType === ResourceTypes.FOLDER &&
+                    selectedTotal === 0 && (
```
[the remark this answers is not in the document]
sri2k1us: Makes sense. Will update the logic. I had that condition when I initially had it to select users Home folder.

src/components/data/MoveDialog.js
```diff
@@ -0,0 +1,192 @@
...
+                                                        )
+                                                    }
+                                                    style={{
+                                                        margin: theme.spacing(
+                                                            1
+                                                        ),
+                                                    }}
+                                                >
+                                                    <RemoveCircleIcon color="error" />
+                                                </IconButton>
```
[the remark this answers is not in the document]
sri2k1us: I initially used `ListItemSecondaryAction` for `remove` button but longer paths bleed over the icon 😩.

src/components/data/listing/Listing.js
sri2k1us: 🤔 Users select files / folders to move from the listing. So deselecting from listing made sense to me. Also I didn't see another way to update `selectedResources` prop for `MoveDialog`.

src/components/data/MoveDialog.js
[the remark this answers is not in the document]
sri2k1us: Here is an example in large screen. It is worse in mobile:
image

src/components/data/listing/Listing.js
psarando: Maybe in this case, it might make sense for the Dialog to manage its own `selectedResources` state, but it could be pre-populated from the listing's prop: `const [selectedResources, setSelectedResources] = useState(initialSelection);`

ash3rz: Yeah, that seems to be what the Bag component is doing too.

sri2k1us: Items in a Bag are fetched from service and not from the listing. We can revisit this if there is specific feedback from user(s).

sri2k1us:
Thanks for the reviews.

PR #306, opened 2021-03-25
Add log in link in notifications menu

flynn1999:
The component renders and runs fine but failed to pass `npm test` due to lack of error boundaries in 'NotificationsMenu.js`. I understand why this is happening, but I am unsure how to fix it in this specific case. Any advice/direction would be appreciated.

### DIFF
--- a/src/components/notifications/NotificationsMenu.js
+++ b/src/components/notifications/NotificationsMenu.js
@@ -114,7 +114,6 @@ function NotificationsMenu(props) {
         showErrorAnnouncer,
     } = props;
     const [notifications, setNotifications] = useState([]);
-    const [error, setError] = useState();
     const [userProfile] = useUserProfile();
     const [errorMessage, setErrorMessage] = useState(null);
     const [errorDialogOpen, setErrorDialogOpen] = useState(false);
@@ -160,7 +159,7 @@ function NotificationsMenu(props) {
                 if (results?.unseen_total > 0) {
                     setUnSeenCount(results?.unseen_total);
                 }
-                setError(null);
+                setErrorObject(null);
             },
             onError: (e) => {
                 setErrorObject(e);
@@ -179,8 +178,8 @@ function NotificationsMenu(props) {
         onSuccess: () => {
             setAllNotificationsSeen();
         },
-        onError: (error) => {
-            showErrorAnnouncer(t("errorMarkAsSeen"), error);
+        onError: (errorObject) => {
+            showErrorAnnouncer(t("errorMarkAsSeen"), errorObject);
         },
     });
 
@@ -265,16 +264,16 @@ function NotificationsMenu(props) {
             {isFetching && (
                 <Skeleton variant="rect" height={400} animation="wave" />
             )}
-            {!isFetching && error !== null && userProfile?.id && (
+            {!isFetching && errorObject !== null && userProfile?.id && (
                 <ListItem>
                     <ErrorTypographyWithDialog
-                        errorMessage={t("notificationError", error)}
+                        errorMessage={t("notificationError", errorObject)}
                     />
                 </ListItem>
             )}
 
             {!isFetching &&
-                error === null &&
+                errorObject === null &&
                 (notifications === null || notifications.length === 0) && (
                     <ListItem>
                         <Typography variant="body2">
--- a/src/components/notifications/NotificationsMenu.js
+++ b/src/components/notifications/NotificationsMenu.js
@@ -3,7 +3,7 @@ import { useMutation, useQuery } from "react-query";
 import Link from "next/link";
 import { formatDistance, fromUnixTime } from "date-fns";
 import classnames from "classnames";
-import { useUserProfile} from "../../contexts/userProfile";
+import { useUserProfile } from "../../contexts/userProfile";
 import {
     getNotifications,
     markAllSeen,
@@ -43,7 +43,6 @@ const NOTIFICATION_MENU_SORT_FIELD = "timestamp";
 const NOTIFICATION_MENU_SORT_ORDER = "desc";
 const NOTIFICATION_MENU_LIMIT = 10;
 const NOTIFICATION_MENU_OFFSET = 0;
-
 
 /*
  * Takes in a notification object and returns the time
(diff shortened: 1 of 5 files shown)


sri2k1us:
So the test failed with error message `useUserProfile must be used within a UserProfileProvider`. Now that you are using `userProfile` context within notifications, these tests must be wrapped inside the `UserProfileProvider` . Checkout `apps.js` under `__tests__` for an example of how to fix this.

src/components/notifications/NotificationsMenu.js
```diff
@@ -156,7 +162,15 @@ function NotificationsMenu(props) {
                 }
                 setError(null);
             },
-            onError: setError,
```
psarando, line 159: I noticed this is still in a `Draft`, but in case this got missed, I just wanted to point out that this was the only place where an error was getting set with `setError`.

So now that it's replaced here with `setErrorObject`, the `error` state should probably also be refactored everywhere to use `errorObject` instead (or refactor your new code to reuse `error` and `setError` instead).

flynn1999 (commented):
Thank you for the help/review!

src/components/notifications/NotificationsMenu.js
```diff
@@ -201,6 +214,27 @@ function NotificationsMenu(props) {
...
+                            open={errorDialogOpen}
+                            id={build(
+                                ids.BASE_DEBUG_ID,
+                                ids.SIGN_IN_ERR_DIALOGUE
+                            )}
+                            errorObject={errorObject}
+                            handleClose={() => {
+                                setErrorDialogOpen(false);
+                            }}
+                        />
```
psarando, line 234: We also have a convenience component that wraps these in `components/utils/error/ErrorTypographyWithDialog`:
https://github.com/cyverse-de/sonora/blob/f028685b75c40ed7f48bb254a7d23eed55a46256/src/components/utils/error/ErrorTypographyWithDialog.js#L16-L35

I would suggest using that component, since it also tracks the fact that this user encountered this sign-in error message (unless we don't want to track that for some reason).

src/components/notifications/ids.js
```diff
@@ -35,6 +35,7 @@ export default {
     REQUEST_HISTORY_DLG: "requestHistoryDlg",
     SET_PRIVILEGE_DLG: "setPrivilegeDlg",
     STATUS: "requestStatusColumn",
+    SIGN_IN_ERR_DIALOGUE: "singInErrorDialogue",
```
ash3rz, line 38: 😄 This is absolutely trivial, but when talking about the UI component it's `dialog` (no `ue` at the end)

ash3rz (approved):
I agree with Paul's comment, otherwise LGTM 👍

sri2k1us:
Also - Can you please add some screenshots ?

src/components/notifications/ids.js
flynn1999: Thank you for pointing out. There is a typo in the id as well 😅

flynn1999:
> Would like to discuss some UX around this. is at @flynn1999 available during our Tuesday meetings ?

I can attend the next one or whichever suitable to discuss the UX around it. 

> Would like to discuss some UX around this. is at @flynn1999 available during our Tuesday meetings ?

I can be available for the next meeting to discuss the UX for this.

flynn1999:
Screenshots:

sri2k1us:
> Screenshots:
> 
> Screen Shot 2021-04-01 at 9 08 53 AM
> Screen Shot 2021-04-01 at 9 08 45 AM

So this is the reason I wanted to discuss this UX - Instead of showing an error message with a button that will again show a login prompt, how about we show the login prompt directly in notification menu? This will eliminate at least 1 click and also makes it very clear that they need to sign-in to view notifications.

flynn1999:
> > Screenshots:
> > Screen Shot 2021-04-01 at 9 08 53 AM
> > Screen Shot 2021-04-01 at 9 08 45 AM
> 
> So this is the reason I wanted to discuss this UX - Instead of showing an error message with a button that will again show a login prompt, how about we show the login prompt directly in notification menu? This will eliminate at least 1 click and also makes it very clear that they need to sign-in to view notifications.

Alright. I'll make the necessary changes! 👍

flynn1999:
Updated screenshot:
Screen Shot 2021-04-01 at 11 19 45 AM

src/components/notifications/NotificationsMenu.js
```diff
@@ -201,52 +204,61 @@ function NotificationsMenu(props) {
...
-                        <DoneAllIcon size="small" />
-                    </IconButton>,
-                ]}
+
+                {!userProfile?.id && (
+                    <ListItem>
+                        <NotAuthorized errorObject={errorObject} />
+                    </ListItem>
+                )}
+
```
sri2k1us, line 213: While this works, `NotAuthorized` is for `HTTP status code 403` error (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403). An example of this error will be a regular user trying to access Admin features in DE. In this case, users are simply trying to access a feature that requires them to login (`Http status code 401`). So creating a component similar to `NotAuthorized` will be useful. Say `NotLoggedIn` ? or We should rename `NotAuthorized` to `Forbidden` and name the new component for `401` to be `Not Authorized`.

flynn1999: I've updated it using a similar component `NotLoggedIn` with `status code 401`. Let me know if this is what was expected and I'll merge this.

sri2k1us: Almost there. Now that #308 is merged, if you look at `src/components/utils/error/errorCode.js`, the `errorObject`  for `401` is defined there as `signInErrorResponse`. So `NotLoggedIn` should be refactored to use that. And then this is good to go!.

ash3rz: One more thing before you merge - you have a commit that still has `WIP` in the commit title.  Let's fix that commit message before merging.  

I think there's some optimization that can be done with this component - there are a lot of checks for whether the user is signed in or not, mobile or not, signed in and mobile, signed in and not mobile, etc. but let's address that in the next PR 👍

PR #308, opened 2021-03-26
CORE-702 Initial App Editor forms

psarando:
This PR will add the initial App Editor forms to Sonora.

This is only the UX in Storybook for now, and a the next PR will integrate the editor into a nextjs page.

The App Editor is in a stepper-form, the same as the App Launch stepper:
App Editor Step 1 - desktop

For mobile:
App Editor Step 1 - mobile

---

In fact, the App Launch form was also refactored so that both forms can use the same stepper:
App Launch Step 1 - desktop

---

The new stepper does not keep the bottom buttons on-screen, so long forms now scroll the entire form. I think it's no longer as essential to keep the bottom buttons in the same place on the screen since there are now navigation buttons in the top-right corner of each step which will stay in the same place as the user navigates from step to step.

The last step of an App Launch replaces only the bottom `Next` button with the `Save` and `Launch` buttons (which means the user may have to scroll down to find them).
App Launch Save/Launch Step - desktop

---

The `Parameters` step in the App Editor uses the same styling for the app group accordions, but adds buttons in the accordion header for editing the label, rearranging, and deleting groups.
App Editor Step 2 cropped - desktop

---

Inside the group is a button for adding parameters. Selecting the `Add Parameter` button will open a dialog for selecting a parameter type:
Parameter Selection Palette - mobile

Selecting the `?` help icon next to each parameter type will expand that type's help text below it:
Parameter Selection Palette help text

Selecting a parameter type anywhere else (the name or left icon, but not the help text or help icon) will add that type of parameter to the top of the group.

---

Each parameter in a group in Step 2 also has buttons for rearranging, deleting, and editing their properties.
App Editor Step 2 - desktop

---

On small screens, the parameter and group row action buttons are replaced with a dot-menu, containing the same actions.
App Editor Step 2 - mobile

---

Selecting the Parameter edit action will replace the groups and parameters arrangement view with the parameter property editor. Since the form can be very long, opening this property editor also auto-scrolls the view back to the top of the form. Replacing the groups and parameters arrangement view also improves form performance, since only the parameter preview field in the property editor needs to re-render as the user edits that parameter's properties.
App Parameter Property Editor - Step 2 - desktop

Once the user clicks `Done` in the property editor, it's replaced once again with the groups and parameters arrangement view, but the form will auto-scroll to the parameter (or group) that the user was editing.

The group editor only has one field for editing the label:
App Group Label Editor - Step 2 - desktop

---

The App Preview step reuses the same components as the `Parameters` step in the App Launch form, so it looks exactly the same as it will in the live App Launch form. The fields in this preview form can be changed, which can show validation errors, and any changes in the preview will have no effect on the App Editor form fields:
App Editor Step 3 - desktop

---

The Command Line Order step shows a text field with the Command Line Preview in it, using the tool's name plus an approximation of the argument option and default value of each parameter. These are parsed and formatted in the UI, since the existing command-line preview endpoint is basically doing the same thing. In the future we will want to have a command-line preview endpoint in the apps service, which can run all parameter preprocessing and return a more accurate "dry run" command-line.

The Command Line Order form also has a list of each parameter (displaying the same preview as used in the preview text field) which can be rearranged, setting each parameter's command-line order (but only if the user decides to move any parameter):
App Editor Step 4 - desktop

### TODO

Currently only the App name and description are required before the user is allowed to save the App. In DEv1 other parameter property fields can be required which also prevent the user from saving if not filled in. For example, at least one Checkbox argument field is required in that type of parameter, but this App Editor will allow the user to save the form without filling them in:
Checkbox Property Editor
They appear as required in the form (the field labels are marked with an `*`), but they do not display an error if both are left blank.

I'm not sure if we want to allow users to save "empty" Checkbox or List parameters like this, or if we want to implement the same requirements the old DE enforces.

#### Other TODO items

* Mobile view for Step 4 may need improvement.
* Default inputs need to validate that the file/folder still exists through the API.
* Editing labels-only in a public app is not yet supported. For now, the user will be shown an informative message that this feature is coming soon.

Public App Editing Coming Soon


src/components/utils/SaveButton.js
```diff
@@ -14,43 +13,45 @@ import {
...
+            {...custom}
         >
-            {t("apply")}
+            {t("save")}
         </Button>
     );
 };
 
-export default ApplyButton;
+export default SaveButton;
```
sri2k1us, line 57: 👍🏽

src/components/utils/SaveButton.js
```diff
@@ -14,43 +13,45 @@ import {
...
             </span>
         </Tooltip>
     ) : (
         <Button
             id={buttonId}
-            disabled={applyDisabled}
-            onClick={onApply}
+            disabled={disabled}
+            onClick={onSave}
             color="primary"
```
sri2k1us, line 47: Minor - May be `color` could also customized by `props` but default can be `primary`.

sri2k1us (approved):
Great Job. Code looks as ❇️ ✨ as it can be! I only had minor comments about the code itself. But have 2 UX related feedback

- Show help / tooltips in parameter type selection screen in line for large screens (right now `?` need to clicked even for large screens) .
- In `Edit` parameter view, i think it would be nice if can hide `Next/Back` buttons and just have `Done` button as `primary` action.  When I playing with it in storybook, I accidentally clicked `Next` button and lost all the edited changes. 
image

From above screenshot, you can see there are more than 1 `primary` action, blue colored button and so it is easy to mix up with the buttons if you are in a hurry. Want to see what others think about this.

public/static/locales/en/common.json
```diff
@@ -61,6 +65,7 @@
     "tools": "Tools",
     "username": "Username",
     "userMenuAriaLabel": "user menu",
+    "useOldDEForThisFeature": "Please use the <oldDELink>previous version of the Discovery Environment</oldDELink> for access to this feature, until we can add it to this version.",
     "viewAllNotifications": "View All Notifications & Activities",
```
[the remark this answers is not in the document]
sri2k1us: This wording can use some review from the 411 team.

sri2k1us:
Also in mobile view, it would be nice if we can reduce the margins / paddings: 
image

sri2k1us:
The Wizard like UX is nice since app launch also shares the same UX. But I feel at the last step, users are left hanging about what to do next. May be there can be options like `Save and Use this app`, `Save and Return to app listing`  in addition to `Back` button and  get rid of `Next` Button at the last step.

public/static/locales/en/app_editor_help.json
```diff
@@ -0,0 +1,43 @@
...
+    "ToolUsed": "A <b>tool</b> is the software program (binary or script) that is installed in a Docker image and integrated into the backend of the Discovery Environment (DE) for use in DE apps.<br/>An <b>app</b> is the <i>interface</i> of a tool that has been configured and made available for use in the DE.<br/>The same <i>tool</i> can be used in more than one <i>app</i>, but an app is created by different integrators with different intentions for that same tool.",
+    "validationRules": "<p><b>Validation rules</b> are the rules you define describing what value the user may enter that will successfully execute the app. If a user enters a value that does not follow the rules you set, the app will display a warning with explanation and not launch the app.</p>"
+}
```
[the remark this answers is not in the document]
psarando: Not sure either. It makes sense to me that the rule is validating which values will allow the app to successfully run (for the analysis to complete successfully); but `to successfully launch the app` makes sense as well, since a validation error will prevent an app from launching from the UI.

I think I mentioned this before, but most of this help text was imported from the DE v1 repo; though I did tweak some text that needed obvious changes, like removing references to tool-tips.

src/components/apps/editor/GroupPropertyForm.js
```diff
@@ -0,0 +1,48 @@
...
+                        }
+                    }}
+                />
+                <Button
+                    id={buildID(baseId, fieldName, ids.BUTTONS.CLOSE_BTN)}
+                    color="primary"
+                    variant="contained"
+                    onClick={onDone}
+                >
+                    {t("common:done")}
```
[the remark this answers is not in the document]
psarando: Good catch 👍

src/components/apps/editor/ParamGroups.js
```diff
@@ -0,0 +1,235 @@
...
+                }
+            >
+                <Typography className={classes.flex} variant="subtitle2">
+                    {group.label}
+                </Typography>
+                <ParamLayoutActions
+                    baseId={groupBaseId}
+                    ButtonProps={{
+                        color: "primary",
+                        variant: isMobile ? undefined : "contained",
```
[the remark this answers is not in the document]
psarando: I thought this was required before I did some refactoring, since `variant="contained"` was getting passed to the DOM element of the dot-menu button; but it's not happening anymore, so I can just remove this `isMobile` check now.

src/components/apps/queries.js
```diff
@@ -0,0 +1,32 @@
...
+        config: {
+            enabled,
+            onSuccess: (resp) => {
+                const genomes = resp?.genomes || [];
+                onSuccess(
+                    stableSort(genomes, (a, b) => a.name.localeCompare(b.name))
+                );
+            },
+            onError: (e) => {
+                console.error(e);
```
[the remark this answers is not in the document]
psarando: This is how it was before in the app launch form, since the reference genome select fields will display an inline error if they are not loading but have no options to display:
https://github.com/cyverse-de/sonora/blob/ed97e9fa7a6583051ee039a27a6f1b02cf090c6c/src/components/apps/launch/ReferenceGenomeSelect.js#L24-L27

Should we also show an announcer error message from here too?

src/components/utils/SaveButton.js
psarando: It actually already works this way now 🎉 

Since all props except `baseId` and `onSave` are being passed along to the button in `{...custom}` (but this reminds me `disabled` will also get passed along and I don't have to set it explicitly).

psarando:
> When I playing with it in storybook, I accidentally clicked `Next` button and lost all the edited changes.

I cleared this up in zoom today, but for the PR: The changes won't be lost by using these stepper navigation buttons, as long as the user stays on the editor page. Formik is "live" updating all changes the user makes in the form, though the user still needs to use the `Save` button save changes in the backend.

src/components/apps/queries.js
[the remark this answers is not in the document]
sri2k1us: Ahh makes sense. Can we add some comments around the `console` message so that it is clear when we encounter this next time.`

sri2k1us:
> > When I playing with it in storybook, I accidentally clicked `Next` button and lost all the edited changes.
> 
> I cleared this up in zoom today, but for the PR: The changes won't be lost by using these stepper navigation buttons, as long as the user stays on the editor page. Formik is "live" updating all changes the user makes in the form, though the user still needs to use the `Save` button save changes in the backend.
> 
> So the `Done` button is not needed to confirm parameter changes (it's really only there to switch the view back to the group/param arrangement view).

My vote will still be get to rid of `Back` and `Next` buttons from the `Edit` parameter view for the following reasons:

- `Back` and `Next` buttons  serve any purpose in `Edit` parameter view. There are no additional views particularly in `Edit` parameter view. If users want to  skip editing they can click `Done` and come out of that view.
- I feel `Back` and `Next` buttons are cluttering the space around the `Done`. Without them, I believe the view will be clean, distraction free and there is only one thing for the user to do.

stories/apps/launch/data/KitchenSinkApp.js
```diff
@@ -0,0 +1,671 @@
...
+//                 ...FlagParamsApp.groups[0].parameters,
+//                 ...NumberParamsApp.groups[0].parameters,
+//             ],
+//             step_number: 0,
+//         },
+//         ...SelectParamsApp.groups,
+//         ...OutputParamsApp.groups,
+//         ...ReferenceGenomeApp.groups,
+//     ],
+// };
```
ash3rz, line 40: Looks like the above can be removed?

stories/apps/launch/data/ReferenceGenomeListing.js
```diff
@@ -3627,3 +3629,11 @@ export default {
         },
     ],
 };
+
+export const initMockAxiosReferenceGenomeListing = () => {
```
ash3rz, line 3633: 👍 I like this pattern of reusing mock API calls

stories/axiosMock.js
```diff
@@ -6,8 +6,19 @@
...
+const errorResponseJSON = {
+    error_code: "ERR_FALSE_ALARM",
+    reason: "Nothing to see here... Please try again!",
+};
+
+const mockErrorResponse = {
+    response: { status: 500, data: errorResponseJSON },
+};
+
+export { AXIOS_DELAY, mockAxios, errorResponseJSON, mockErrorResponse };
```
ash3rz, line 24: 👍

ash3rz (approved):
✨ very readable and clean!  Pretty much all of my comments were minor things about pesky IDs.  I still need to spend some time with the UI/UX, but the code looks ready to go 👍

src/components/apps/AppStepper.js
```diff
@@ -0,0 +1,151 @@
...
+                    <Button
+                        size="small"
+                        variant="contained"
+                        onClick={handleNext}
+                        disabled={isLastStep()}
+                        id={buildID(baseId, ids.APP_STEPPER.STEP_NEXT)}
+                        color="primary"
+                    >
+                        {t("next")}
+                        {theme.direction === "rtl" ? (
```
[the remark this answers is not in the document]
psarando: I didn't either! This is copied from @sri2k1us code somewhere (and I think it's in one of the MUI example docs) 😅

public/static/locales/en/app_editor_help.json
```diff
@@ -0,0 +1,43 @@
...
+    "MultiFileOutput": "Allow the user to name the file containing the results the app generates. You may also set a default output name.",
+    "MultiFileSelector": "Adds an input file field that allows users to browse their data and select <i>one or many</i> input files to use. As the integrator, you can specify the type of files that the user may use.",
+    "MultiLineText": "Similar to Single-line Text, except the user may enter more than one line of text.",
+    "ReferenceAnnotation": "<p>Allow you to add drop-down lists of a <b>Reference Annotation</b> from which the user can make one selection.</p><p>The Discovery Environment has reference annotations uploaded and available for use.</p><p>To add a reference species that is not currently supported, please email the CyVerse support team at support@cyverse.org.</p>",
```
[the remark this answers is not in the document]
psarando: Good catch. I forgot to circle back to this and replace these emails with `contact support` buttons that should open the intercom chat (we've done that in a couple of other places too).

stories/apps/launch/data/KitchenSinkApp.js
psarando: Oops, yep, some cruft that shouldn't have been committed 😅

src/components/apps/editor/params/CheckboxPropertyFields.js
```diff
@@ -0,0 +1,104 @@
...
+                alignItems="flex-start"
+            >
+                <Grid item xs={6}>
+                    <FastField
+                        id={buildID(baseId, ids.PARAM_FIELDS.CHECKED_OPTION)}
+                        name={`${fieldName}.name.checked.option`}
+                        label={t("checkboxArgOptionCheckedLabel")}
+                        margin="normal"
+                        required
+                        component={FormTextField}
```
psarando, line 51: I mentioned this in the PR description, but these Checkbox `option` fields are marked as `required`, though they don't have any form validation checks. So the user can leave both blank and still save the App (although it probably won't be able to run).

The old DE would display an error if both were left blank and not allow the user to save the App in this state. I'm not sure if we want to do the same thing here, or go ahead and allow the user to save a "broken" app (but we could at least provide some additional help text explaining that below these fields).

src/components/apps/launch/AppLaunchForm.js
```diff
@@ -588,90 +179,43 @@ const AppLaunchForm = (props) => {
...
-                setReferenceGenomes(
-                    stableSort(genomes, (a, b) => a.name.localeCompare(b.name))
-                );
-                setReferenceGenomesQueryEnabled(false);
-            },
-            onError: (e) => {
-                console.error(e);
-            },
-        },
-    });
```
psarando, line 641: I initially put this reference genomes listing query at the top-level of the App Launch form with the intention that the listing would only have to be fetched once, no matter how many reference genome selectors were included in the form.

It turns out that react-query is already smart enough to cache the listing for some amount of time after it's fetched once, at least on the server-side. For example, now when I run my example app in dev mode with this code, where 3 reference genome selectors are running their own query for the listings, I will see 3 requests from the browser to dev server, but I only see 1 request in the console from the server to the API.

src/components/apps/launch/ReferenceGenomeSelect.js
```diff
@@ -7,17 +7,29 @@
...
-    referenceGenomesLoading,
-    ...props
-}) => {
+const ReferenceGenomeSelect = ({ param, ...props }) => {
+    const [referenceGenomes, setReferenceGenomes] = React.useState([]);
+
+    const { isFetching: referenceGenomesLoading } = useReferenceGenomes(
+        true,
+        setReferenceGenomes
+    );
```
psarando, line 23: An odd side effect of moving the listing query into the reference genome selector components, is that when a form with multiple selectors first loads, each selector appears to be loading the listing (as intended), but if the view changes, and the selectors are hidden and then re-shown, then only the first selector in the form appears to be loading again, but the remaining selectors apparently are getting a `false` value for `isFetching`, but also an empty listing, so they will display an error message that the listing couldn't be loaded.

It's only temporary though. Once the first selector loads, they all refresh and load the listing as normal.

You can test this in Storybook with the "kitchen sink" test app: https://5ebad5c6b991d60022b1e72e-cgpehakrzf.chromatic.com/?path=/story/apps-editor--kitchen-sink-editor
Go to step 2 and quickly scroll down to see the 3 selectors loading, then click the "Next" button at the bottom to go to the preview step. The view should stay scrolled to the bottom of the preview where you can see the first selector reloading, and the last 2 showing an error until the first one finishes loading; then they all appear as normal.

I'm not sure if I should be checking something else besides `isFetching` here (the `isLoading` flag appears to only be set when the list is initially loaded from the API, but not after its cached).

src/components/apps/editor/formatters.js
```diff
@@ -0,0 +1,387 @@
...
+const initAppValues = (app) => {
+    const { groups } = app;
+
+    const initializedGroups = groups?.map(({ name, ...group }, groupIndex) => ({
+        ...group,
+        key: group.id || `group${groupIndex}`,
+        parameters: group.parameters?.map((param, paramIndex) => {
+            const { type: paramType, arguments: paramArgs } = param;
+
+            const key = param.id || `group${groupIndex}.param${paramIndex}`;
```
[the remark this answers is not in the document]
psarando: I'll go ahead and sum this up in a comment on the code here.

psarando:
I've addressed most of the feedback and comments given so far, so I'll go ahead and merge this now, but more reviews are still welcome! 🙇 

I'll address the navigation buttons on group/param edit views, and the always-expanded help text in the param selection palette in subsequent PRs.

src/components/apps/editor/ParamSelectionPalette.js
```diff
@@ -0,0 +1,274 @@
...
+                                                UtilIds.CONTACT_SUPPORT_BUTTON
+                                            )}
+                                            component="button"
+                                            variant="body2"
+                                            onClick={(event) => {
+                                                // prevent parameter selection
+                                                event.stopPropagation();
+                                                intercomShow();
+                                            }}
+                                        />
```
psarando, line 174: By the way, setting the `variant` prop on a `Link` that is also using `component="button"` will fix the font appearance (in Firefox at least), when set to the same `Typography` variant as the surrounding text.

[The docs](https://material-ui.com/components/links/#simple-links) say that the "Link component is built on top of the Typography component. You can leverage its properties."

PR #309, opened 2021-03-29
core-1354 Allow users to rate apps from notification message.

sri2k1us:
- Rating widget
image

- User Rating
image

- iPhone 5
image

### DIFF
--- a/src/components/apps/RatingWidget.js
+++ b/src/components/apps/RatingWidget.js
@@ -1,0 +1,109 @@
+/**
+ * @author sriram
+ *
+ * A widget that allows users to rate an app.
+ */
+import React, { useEffect, useState } from "react";
+import { useTranslation } from "i18n";
+import { queryCache, useMutation, useQuery } from "react-query";
+
+import { Rate } from "@cyverse-de/ui-lib";
+
+import constants from "../../constants";
+import { getAppById, rateApp, APP_BY_ID_QUERY_KEY } from "serviceFacades/apps";
+import ErrorTypographyWithDialog from "components/utils/error/ErrorTypographyWithDialog";
+
+import { Grid, Typography } from "@material-ui/core";
+
+export default function RatingWidget(props) {
+    const { appId, systemId, appName } = props;
+    const { t } = useTranslation("apps");
+
+    const [ratingMutationError, setRatingMutationError] = useState(null);
+    const [selectedApp, setSelectedApp] = useState(null);
+    const [appRating, setAppRating] = useState({
+        average: 0,
+        total: 0,
+        user: 0,
+    });
+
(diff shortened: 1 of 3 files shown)


src/components/apps/RatingWidget.js
```diff
@@ -0,0 +1,109 @@
...
+        });
+    };
+
+    useEffect(() => {
+        if (selectedApp) {
+            setAppRating(selectedApp.rating);
+        }
+    }, [selectedApp]);
+
+    const loading = appByIdStatus || ratingMutationStatus === constants.LOADING;
```
ash3rz, line 75: I think this was meant to check both variables for the loading status:
```suggestion
    const loading = appByIdStatus === constants.LOADING || ratingMutationStatus === constants.LOADING;
```

This should also work:
```
import isQueryLoading from "components/utils/isQueryLoading";

const loading = isQueryLoading([appByIdStatus, ratingMutationStatus])
```

ash3rz (approved):
LGTM :shipit: 
One minor thing that I think would need to come from ui-lib, is the trash can is a little misaligned with the rating stars when they're on the same line.

src/components/apps/RatingWidget.js
psarando: The `appByIdStatus` var is actually an alias for the app listing query's `isFetching` flag, so this works, though maybe the alias is misnamed. The `isQueryLoading` util function is still a good idea 👍

src/components/notifications/Message.js
```diff
@@ -65,7 +67,25 @@ function AnalysisLink(props) {
...
+                    notification.payload?.status === analysisStatus.FAILED ||
+                    notification.payload?.status === analysisStatus.CANCELED);
+            return (
+                <>
+                    <MessageLink message={message} href={href} as={as} />
+                    <br />
+                    {allowRating && (
+                        <RatingWidget
+                            appId={notification.payload?.app_id}
+                            appName={notification?.payload.app_name}
```
psarando, line 83: Other places in this code are checking for `notification.payload?.`, but here it's `notification?.payload.`

Maybe these could be refactored into a `payload` var that replaces all of them?
```javascript
const payload = notification?.payload;
```
Then `payload?.app_name` etc.

psarando (approved):
LGTM 👍 
Though I had a couple of minor refactor suggestions.

sri2k1us:
> LGTM :shipit:
> One minor thing that I think would need to come from ui-lib, is the trash can is a little misaligned with the rating stars when they're on the same line.

Will file an issue for this and move-on since there are other priority items to work on now.

sri2k1us:
Thanks for the reviews.

PR #311, opened 2021-03-30
CORE-1282: support OAuth authorization callbacks in Sonora

slr71:
image
image


public/static/locales/en/oauth.json
```diff
@@ -0,0 +1,15 @@
+{
+    "invalid_request": "The authorization request sent to the remote system by the DE was invalid. Please contact support.",
+    "unauthorized_client": "The DE is not authorized to request access to the remote system. Please contact support.",
+    "access_denied": "The remote system denied the authorization request. Please try again or contact support.",
+    "unsupported_response_type": "The remote system does not support the response requested by the DE. Please contact support.",
+    "invalid_scope": "The remote system does not support the scope requested by the DE. Please contact support.",
+    "server_error": "The remote system system encountered an error. Please try again or contact support.",
+    "temporarily_unavailable": "The remote system is temporarily unavailable. Please try again or contact support.",
```
slr71, line 8: These error codes come directly from the OAuth specification. Having a unique message for each code will help us to troubleshoot these errors if they occur.

public/static/locales/en/oauth.json
```diff
@@ -0,0 +1,15 @@
...
+    "no_auth_code_provided": "Invalid authorization attempt. Please contact support.",
+    "no_state_id_provided": "Required information was missing from the authorization attempt. Please contact support.",
+    "general_service_error": "The DE service encountered an error. Please contact support.",
```
slr71, line 11: These error codes are specific to the DE:

- `no_auth_code_provided` means that the `code` query parameter was missing from the callback URL.
- `no_state_id_provided` means that the `state` query parameter was missing from the callback URL.
- `general_service_error` means that the API call to finish the authorization process failed. In this case, we can look at the `apps` and `terrain` logs for more information.

src/components/oauth/OAuthCodeHandler.js
```diff
@@ -0,0 +1,100 @@
...
+    const { baseId, apiName, code, stateId } = props;
+    const handlerId = build(baseId, ids.OAUTH_CODE_HANDLER);
+    const router = useRouter();
+    const [userProfile] = useUserProfile();
+    const [categoryQueryKey, setCategoryQueryKey] = useState([
+        APP_CATEGORIES_QUERY_KEY,
+        userProfile?.id,
+    ]);
+    const [listingUrl, setListingUrl] = useState(
+        getListingPath(null, null, null, null, null, null, false)
```
slr71, line 38: I used a general app listing URL as the default. That seemed to be the most reasonable thing to do if we can't determine the URL for HPC app listings.

src/components/oauth/OAuthCodeHandler.js
```diff
@@ -0,0 +1,100 @@
...
+            setListingUrl(
+                getListingPath(
+                    null,
+                    null,
+                    null,
+                    null,
+                    null,
+                    JSON.stringify(hpcCategory),
+                    false
+                )
```
slr71, line 55: Redirecting users here means that this component is specific to Tapis If we ever add support for other third-party APIs, we'll have to revisit this.

src/components/utils/error/ClientInfo.js
```diff
@@ -0,0 +1,70 @@
+/**
+ * A component used to display details about the user and user agent that may
+ * be helpful when troubleshooting errors.
+ */
```
slr71, line 4: I extracted this from `ErrorHandler.js`. That seemed like the most reasonable thing to do to me. I'm curious to know what others think, though.

src/components/oauth/OAuthCodeHandler.js
```diff
@@ -0,0 +1,100 @@
...
+} from "serviceFacades/apps";
+import {
+    OAUTH_CALLBACK_QUERY_KEY,
+    doOAuthCallback,
+} from "serviceFacades/oauth";
+
+import OAuthErrorHandler from "./OAuthErrorHandler";
+import OAuthLinkingMessage from "./OAuthLinkingMessage";
+import ids from "./ids";
+import { getListingPath } from "components/apps/utils";
```
slr71, line 22: Importing `getListingPath` directly from `components/apps/utils.js` seemed reasonable to me, but I'm curious about what others think.

src/components/utils/error/ContactSupport.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * A component that can be used to allow a user to contact support.
+ */
```
slr71, line 3: I also extracted this from `ErrorHandler.js`. Once again, this seemed to me to be the most reasonable thing to do, but I'm curious about what others think.

slr71:
I'm not sure why the lint check is failing. I haven't been able to get the check to fail on my machine.

slr71:
Still no clue why the lint check is failing. 🤷‍♀️

slr71:
We figured out why the lint check was failing. There was a change that had been merged into master that wasn't in my initial PR to add `useEffect` on line 61 of `ErrorHandler.js`. I didn't see the changes locally because I initially didn't know that another PR had been merged and I wasn't merging in the changes from master. After I did the rebase, I probably could have found that `useEffect` call if I'd done the lint check locally after doing the rebase. I want to believe that I did, but I probably didn't. My only excuse would be that it was the end of the day and I was tired. 🤦‍♀️  At any rate, mystery solved. 😂

sri2k1us:
LGTM. I recommend getting the messages reviewed by the 411 team to make sure the error messages are clear and easy to understand for non-tech users.

sri2k1us (approved):
:shipit:

slr71:
Thanks for the review! I'll ping the 411 team to ask them to review the messages.

public/static/locales/en/oauth.json
```diff
@@ -0,0 +1,15 @@
+{
+    "invalid_request": "The authorization request sent to the remote system by the DE was invalid. Please contact support.",
+    "unauthorized_client": "The DE is not authorized to request access to the remote system. Please contact support.",
```
ash3rz, line 3: Just picking a random spot - I agree we should have 411 review the text.  I wouldn't be surprised if a user thought "remote system" meant their system and they started messing with permission settings on their computer or something like that 😄

ash3rz (approved):
LGTM and I like the refactors 👍 🎉 
As Sriram already said, having 411 review the text would be good 👍

src/components/oauth/OAuthCodeHandler.js
```diff
@@ -0,0 +1,100 @@
...
+        config: {
+            onSuccess: determineListingUrl,
+            staleTime: Infinity,
+            cacheTime: Infinity,
+        },
+    });
+
+    useEffect(() => {
+        setCategoryQueryKey([APP_CATEGORIES_QUERY_KEY, userProfile?.id]);
+    }, [userProfile]);
```
psarando, line 73: As an alternative to using state to manage the app category query key, the query's `config` can use the `enabled` flag based on the presence of the user's profile ID, and this query key array can be used directly in the `queryKey`.

Here's an example where the `enabled` flag is based on having a user ID (though it's not used in the query key here):
https://github.com/cyverse-de/sonora/blob/3d8cb7640ef34c6192d8559f8d33f7fbbc48931f/src/pages/apps/%5BsystemId%5D/%5BappId%5D/edit.js#L34-L44

src/components/oauth/OAuthCodeHandler.js
```diff
@@ -0,0 +1,100 @@
...
+
+    if (isFetching || isFetchingCategories) {
+        return <OAuthLinkingMessage baseId={handlerId} />;
+    }
+
+    if (error) {
+        return <OAuthErrorHandler errorCode={ERR_SERVICE} baseId={handlerId} />;
+    }
+
+    router.push(listingUrl);
```
psarando, line 96: If we use `router.replace` here, then the browser's back button won't take the user back to the `oauth/callback` page from that `listingUrl`.

src/components/oauth/OAuthErrorHandler.js
```diff
@@ -0,0 +1,71 @@
...
+    const title = (
+        <Typography color="error" variant="h6">
+            {t("oops")}
+        </Typography>
+    );
+    const subheader = <Typography color="error">{t(errorCode)}</Typography>;
+
+    return (
+        <Card id={errorBaseId}>
+            <CardHeader avatar={avatar} title={title} subheader={subheader} />
```
psarando, line 56: FYI, the `CardHeader` component accepts props `titleTypographyProps` and `subheaderTypographyProps`, since it's already wrapping the `title` and `subheader` in its own `Typography` components.

That means this text will be double-wrapped in `Typography` components (which probably isn't a big deal).

This code looks pretty tidy, though, so I'll leave it to personal preference, but here's an example where I'm using `titleTypographyProps`:
https://github.com/cyverse-de/sonora/blob/3d8cb7640ef34c6192d8559f8d33f7fbbc48931f/src/components/apps/AppStepDisplay.js#L58-L65

psarando (approved):
LGTM 👍 ✨ 

I just had a couple of minor alternate-style suggestions.

stories/OAuth.stories.js
```diff
@@ -0,0 +1,65 @@
...
+import OAuthLinkingMessage from "components/oauth/OAuthLinkingMessage";
+import OAuthErrorHandler from "../src/components/oauth/OAuthErrorHandler";
+
+export default { title: "OAuth" };
+
+const OAuthErrorTemplate = (args) => (
+    <OAuthErrorHandler baseId="someBaseId" {...args} />
+);
+
+export const InvalidRequest = OAuthErrorTemplate.bind({});
```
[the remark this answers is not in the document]
slr71: I really like this feature. It was really convenient for this case.

public/static/locales/en/oauth.json
```diff
@@ -0,0 +1,15 @@
...
+    "access_denied": "The remote system denied the authorization request. Please try again or contact support.",
+    "unsupported_response_type": "The remote system does not support the response requested by the DE. Please contact support.",
+    "invalid_scope": "The remote system does not support the scope requested by the DE. Please contact support.",
+    "server_error": "The remote system system encountered an error. Please try again or contact support.",
```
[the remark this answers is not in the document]
slr71: Thanks for catching that.

slr71:
Thanks for the reviews! I'll wait for feedback from 411 before merging this PR.

slr71:
I'm going to merge this now.

PR #315, opened 2021-04-05
Add CSS for table resizing bug

flynn1999:
Before:
Screen Shot 2021-04-05 at 10 48 04 AM
After:
Screen Shot 2021-04-05 at 10 48 41 AM

Before:
Screen Shot 2021-04-05 at 10 49 07 AM
After:
Screen Shot 2021-04-05 at 10 51 47 AM

Before:
Screen Shot 2021-04-05 at 10 50 47 AM
After:
Screen Shot 2021-04-05 at 10 51 03 AM

### DIFF
--- a/src/components/utils/DEPagination.js
+++ b/src/components/utils/DEPagination.js
@@ -37,6 +37,9 @@ const useStyles = makeStyles((theme) => ({
         paddingBottom: theme.spacing(1),
         color: theme.palette.info,
     },
+    mainGrid: {
+        justifyContent: "center",
+    },
     buttonPadding: {
         paddingTop: theme.spacing(1.5),
     },
@@ -171,8 +174,7 @@ function DEPagination(props) {
     const matches = useMediaQuery(theme.breakpoints.down("sm"));
     return (
         <Paper className={classes.paper}>
-            <Grid container>
-                <Grid item xs={4} sm={5} md={5} lg={5}></Grid>
+            <Grid container className={classes.mainGrid}>
                 <Grid item>
                     <Pagination
                         id={build(baseId, ids.PAGINATION_TOOLBAR)}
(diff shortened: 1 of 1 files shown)


sri2k1us (changes requested):
LGTM. But might need some more tweak. See 👇🏽 
- Your PR:
image

- Prod
image

flynn1999:
**v2 (using justifyContent)**

Before (Pixel 2):
Screen Shot 2021-04-05 at 12 01 11 PM

After (Pixel 2):
Screen Shot 2021-04-05 at 12 01 33 PM

Before (Medium):
Screen Shot 2021-04-05 at 12 03 28 PM

After(Medium):
Screen Shot 2021-04-05 at 12 03 10 PM

Before (Small):
Screen Shot 2021-04-05 at 12 04 09 PM

After (Small):
Screen Shot 2021-04-05 at 12 04 32 PM

Before(XS):
Screen Shot 2021-04-05 at 12 05 08 PM

After(XS):
Screen Shot 2021-04-05 at 12 05 20 PM

sri2k1us (approved):
🎉  Looks good. Great job! I had difficulty figuring this out when I developing this.

src/components/utils/DEPagination.js
```diff
@@ -171,8 +174,7 @@ function DEPagination(props) {
     const matches = useMediaQuery(theme.breakpoints.down("sm"));
     return (
         <Paper className={classes.paper}>
-            <Grid container>
-                <Grid item sm={4} md={5}></Grid>
+            <Grid container className={classes.mainGrid}>
```
psarando, line 177: This works for now, but `Grid` container components have an equivalent `justify` prop that can be used instead:
```suggestion
            <Grid container justify="center">
```
This might be better, since the `Grid` component may be upgraded in the future to use CSS grid layouts, instead of the flex layout it's using under the hood now, and I'm not sure if the `justifyContent` CSS will still apply then.

FYI, the https://material-ui.com/components/grid/ page has an interactive demo that shows the effect of the different container prop settings.

PR #316, opened 2021-04-05
CORE-1356 Recent Contacts List

ash3rz:
This feature is replacing the Collaborators window feature from the old DE. The Collaborators window worked by creating a `default` collaborator list for every user.  This list was manually populated by the user with collaborators that they searched for and added to the Collaborators window, and also any other named collaborator lists that the user created.

For recent contacts, we're just reusing the `default` collaborator list for each user and automatically populating that list with users and groups that they select in the sharing dialog or the teams dialog, or wherever the `SubjectSearchField` ends up being used.

Known bug: on mobile, the trash icon overlaps with the text.  Still looking into a fix.

Desktop example showing fuzzy matches in Recent Contacts, but exact match in Search Results:
image

Mobile example:
image


src/components/utils/DeleteButton.js
```diff
@@ -0,0 +1,51 @@
...
+    );
+}
+
+DeleteButton.propTypes = {
+    baseId: PropTypes.string.isRequired,
+    ariaLabel: PropTypes.string,
+    component: PropTypes.oneOf(["Button", "IconButton"]),
+};
+
+export default DeleteButton;
```
sri2k1us, line 51: Nice 🎉

sri2k1us (approved):
Code LGTM but I have 1 question - 
- When I tried this locally against QA, for `api/collaborator-lists/default/members`  I got `404` with the following error `{"description":"group","id":"iplant:de:qa:users:sriram:collaborator-lists:default","error_code":"ERR_NOT_FOUND","reason":null}` for the first time and then it worked the next after shared it once with `aramsey`.  Is this an expected behavior ? Wonder if we should even show an error if we cannot fetch recent contacts in the background.
- Feature request / for another PR - If you check Google drive, as soon as you click on the search bar, you will get some results even before typing anything. Similarly I think we can pre-fetch the recent contacts and show top `x`  results even before something is typed. And then when user types some chars, we can filter the recent contacts based on it.  
- Regarding `the trash icon overlaps with the text` issue, that is the same issue I had in https://github.com/cyverse-de/sonora/pull/304#discussion_r604355534. So I had to move the icon to the left.

sri2k1us:
Ohh Just noticing it - Think we should make sharing dialog full screen for mobile to make better use of screen real estate.

src/components/utils/DeleteButton.js
```diff
@@ -0,0 +1,51 @@
...
+
+    const isButton = component === "Button";
+    const Component = isButton ? Button : IconButton;
+
+    return (
+        <Component
+            id={build(baseId, ids.BUTTONS.DELETE)}
+            aria-label={ariaLabel || t("delete")}
+            className={isButton ? classes.deleteBtn : null}
+            classes={!isButton ? { root: classes.deleteBtn } : null}
```
psarando, line 37: Ah, I tried adding a Delete button in a previous PR, but the correct styling tripped me up 👍

psarando (approved):
LGTM! 🎉 ✨

psarando:
> Regarding `the trash icon overlaps with the text` issue, that is the same issue I had in [#304 (comment)](https://github.com/cyverse-de/sonora/pull/304#discussion_r604355534). So I had to move the icon to the left.

I wonder why we don't see this same issue in their example code (even if I edit the example in the code sandbox with longer list item text) 😕 
https://material-ui.com/components/lists/#list-controls

ash3rz:
Making the sharing dialog full screen on mobile helps!
And yes - good catch - in the old DE we were hiding errors because the first time you open up the collaborators window was when the default list would get created if it wasn't already there.  I've removed the error announcer.
Not sure about using ellipses since it could maybe make it hard to figure out which option is which  🤔 For now, I'll move forward with moving the trash icon to the left.

ash3rz:
Thanks for the reviews!

src/components/utils/DeleteButton.js
```diff
@@ -0,0 +1,51 @@
...
+import PropTypes from "prop-types";
+
+import ids from "./ids";
+import { makeStyles } from "@material-ui/styles";
+import { useTranslation } from "../../i18n";
+
+const useStyles = makeStyles((theme) => ({
+    deleteBtn: {
+        color: theme.palette.error.main,
+    },
```
psarando, line 20: Forgot to mention this before, but in the app editor I had to use `backgroundColor: theme.palette.error.contrastText` for its delete icons, since the red on a dark background (such as in a groups accordion summary) doesn't stand out well:
App Editor group action icons

I'm wondering if we should also add this background contrast color to this common delete button in a future PR.

PR #317, opened 2021-04-06
Upgrade amqplib to 0.7.1, which supports nodejs 15

ianmcorvidae:
Just a small thing, but in looking at the lockfile version 2 thing that changed recently, I upgraded to nodejs 15, which amqplib 0.6.0 doesn't support. Doesn't appear to break things or make any other meaningful changes.


package.json
```diff
@@ -72,7 +72,6 @@
         "express-winston": "^4.0.2",
         "express-ws": "^4.0.0",
         "formik": "^2.2.5",
-        "fsevents": "^1.2.11",
```
ianmcorvidae, line 75: I'm not sure why this was removed, but `npm install --upgrade amqplib` is what did this.

ash3rz (approved):
Tried it out locally on my machine and LGTM :shipit:

johnworth (approved):
LGTM

sri2k1us:
See #327.

PR #318, opened 2021-04-07
Core 1359 Webhook preferences

sri2k1us:
- Large Screen
image

- Mobile
image

- Test Failed
image

- Test Success
image


src/server/api/user.js
```diff
@@ -96,5 +99,59 @@ export default function userRouter() {
...
+                res.send(apiResponse.data);
+            })
+            .catch(async (err) => {
+                logger.error(err);
+                const e = await err;
+
+                res.status(e.response?.status || 500);
+                res.send(e.response?.data);
+            });
+    });
```
sri2k1us, line 154: This is a copy-pasta from `/server/external.js` since I couldn't figure out how to pass the test url by query param to `external.js` yet.

src/components/preferences/Webhooks.js
```diff
@@ -0,0 +1,178 @@
...
+                                label={t("type")}
+                                required
+                                field={field}
+                                onChange={(event) => {
+                                    onChange(event);
+                                }}
+                                variant="outlined"
+                                size="small"
+                            >
+                                {webhookTypes?.map((type, index) => (
```
ash3rz, line 94: Paul's better at Formik stuff, but I think since you're not altering the behavior of the textfield, I think you can simplify this to:
```suggestion
                    <Field
                        name="webhook.type.type"
                        id={build(baseId, ids.WEBHOOK_TYPES_SELECT)}
                        label={t("type")}
                        required
                        select
                        variant="outlined"
                        size="small"
                        component={FormTextField}
                    >
                      {webhookTypes?.map((type, index) => (
```
I found an example [here](https://github.com/cyverse-de/sonora/blob/c2d5711cd26d5233f51fcfb428c9794fb8a0076b/src/components/apps/editor/params/FileOutputPropertyFields.js#L61)

src/server/api/user.js
```diff
@@ -10,8 +10,11 @@ import express from "express";
 import * as auth from "../auth";
 import logger from "../logging";
 
+import axiosInstance from "../../common/getAxios";
 import { handler as terrainHandler } from "./terrain";
 
+const WEBHOOK_TEST_BODY = { text: "This is a test message from DE." };
```
ash3rz, line 16: Minor, but seems like this could be on the client-side and set in the service facade instead of here on the server-side.

ash3rz (approved):
LGTM :shipit: 

From the screenshots, I feel like on desktop the URL field could be a little wider so users can see more of the URL. 
The `Test` button is also black which I don't think we're using elsewhere.  Maybe a primary color or there's the dark gray color which I think is `theme.palette.info.main`.

src/components/preferences/Webhooks.js
psarando: I agree. It's cleaner to set props directly on the `Field`, which will pass them along to its `component` (which is `FormTextField` in this case).

src/components/preferences/Preferences.js
```diff
@@ -340,10 +425,39 @@ function Preferences(props) {
...
+    const mapPropsToValues = (bootstrap) => {
+        if (bootstrap === null || bootstrap === undefined) {
+            return {};
+        }
+        if (bootstrap?.preferences) {
+            if (bootstrap?.apps_info?.webhooks[0]) {
+                let hook = bootstrap?.apps_info?.webhooks[0];
+                const selectedTopics = hook?.topics;
+                const reducer = (acc, currVal) => ({ ...acc, [currVal]: true });
+                const topics = selectedTopics?.reduce(reducer, {}, 0, {});
```
psarando, line 438: To avoid `uncontrolled to controlled` form field warnings, it's best for every field in the form to have an initial value.

Maybe it's not an issue with `Switch` or `Checkbox` fields, but because the `Webhooks` component is creating a field for every topic in `webhookTopics`, I would initialize a value for every topic:
```suggestion
                const selectedTopics = hook?.topics;
                const reducer = (acc, currVal) => ({ ...acc, [currVal]: selectedTopics?.includes(currVal) });
                const topics = webhookTopics.reduce(reducer, {});
```

src/server/api/user.js
psarando: I'm not sure either, but it looks like the external API handler is not setup to handle `POST` requests with a `body` anyway (though I'm sure it could be modified to accept a `body` param if needed).

psarando (approved):
LGTM 👍 🎉 

I just had a couple of Formik recommendations (one @ash3rz already pointed out).

src/components/preferences/Preferences.js
sri2k1us: I always stumble on that `uncontrolled to controlled` but I keep forgetting the solution. Thanks 👍🏽.

sri2k1us:
Thanks for the reviews.

src/components/preferences/Preferences.js
```diff
@@ -340,10 +423,42 @@ function Preferences(props) {
...
+        }
+        if (bootstrap?.preferences) {
+            if (bootstrap?.apps_info?.webhooks[0]) {
+                let hook = bootstrap?.apps_info?.webhooks[0];
+                const selectedTopics = hook?.topics;
+                const reducer = (acc, currVal) => ({
+                    ...acc,
+                    [currVal]: selectedTopics?.includes(currVal),
+                });
+                const topics = selectedTopics.reduce(reducer, {});
```
psarando, line 439: This should be reducing on all possible topics:
```suggestion
                const topics = webhookTopics.reduce(reducer, {});
```

sri2k1us: Oops! Yes.

sri2k1us: Heh..Actually there is little more to it.  line `437` should be:
` const reducer = (acc, currVal) => ({
                    ...acc,
                    [currVal?.topic]: selectedTopics?.includes(currVal?.topic),
                });`

PR #319, opened 2021-04-09
CORE-702 Hide App Editor navigation buttons when editing groups/params

psarando:
This PR will update the App Editor so that the step navigation buttons are hidden when editing app groups and params.

The simplest way to do this was to refactor the state management of displaying the group and param property editors, out of the groups form, up into the main editor form. That way the navigation buttons can simply be hidden while one of those editors is open.

Here is the App Editor Step 2 when not editing a group or param:
App Editor - Step 2

---

Now when editing param properties:
App Editor - Step 2 - edit param

---

And the group label editor:
App Editor - Step 2 - edit section

---

These changes do no affect the mobile stepper, which is always displayed at the bottom of the screen, on top of everything else.

### DIFF
--- a/src/components/apps/editor/ParamGroups.js
+++ b/src/components/apps/editor/ParamGroups.js
@@ -13,10 +13,8 @@ import { useTranslation } from "i18n";
 import ids from "./ids";
 import styles from "./styles";
 
-import GroupPropertyForm from "./GroupPropertyForm";
 import Parameters from "./Parameters";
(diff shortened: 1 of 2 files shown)


sri2k1us:
LGMT :shipit: .  I think the `Edit` view now looks a lot cleaner. ✨

sri2k1us (approved):

psarando:
Thanks for the review! 👍 

I'll go ahead and merge this now, though additional reviews still welcome for those so inclined 🙇

PR #320, opened 2021-04-09
Core 1366 - Admin App publication Requests

sri2k1us:
- Large Screen
image

- iPhone 5
image

- Tools Used
image

- Error Handling
image


src/components/apps/admin/publicationRequests/RequestListing.js
```diff
@@ -0,0 +1,267 @@
...
+    const { tools, appName, parentId } = props;
+    const [toolsDialogOpen, setToolsDialogOpen] = useState(false);
+    const { t } = useTranslation("tools");
+    return (
+        <React.Fragment>
+            <Button color="primary" onClick={() => setToolsDialogOpen(true)}>
+                {t("viewTools")}
+            </Button>
+            <DEDialog
+                id={build(parentId, ids.PUBLICATION_REQUESTS.TOOLS_USED_DIALOG)}
```
ash3rz, line 85: Minor as always, but you can remove the `parentId` from here and the rest of the elements in the dialog

src/components/apps/admin/publicationRequests/RequestListing.js
```diff
@@ -0,0 +1,267 @@
...
+
+const appColumnData = (t) => [
+    { name: t("name"), enableSorting: false, key: "name" },
+    {
+        name: t("integratorName").split(":")[0],
+        enableSorting: false,
+        key: "integrator_name",
+    },
+    {
+        name: t("integratorEmail").split(":")[0],
```
ash3rz, line 49: Does the integrator email and integrator name have `:` in them?

src/components/apps/admin/publicationRequests/RequestListing.js
```diff
@@ -0,0 +1,267 @@
...
+                                        />
+                                    </TableCell>
+                                    <TableCell>
+                                        <Tooltip
+                                            title={t("publishApp", {
+                                                appName: request.app.name,
+                                            })}
+                                        >
+                                            <IconButton
+                                                onClick={() =>
```
ash3rz, line 241: This button could use an id for testing

ash3rz (approved):
LGTM :shipit:

psarando (changes requested):
I noticed a `vice-loading.json` file was added in the PR, but it's not being used anywhere? If that's the case, it should be removed from this PR.

Also there's a conflict with the `package-lock.json` file now after the latest PR, but the changes in this PR can be discarded in a rebase.

ash3rz:
> I noticed a `vice-loading.json` file was added in the PR, but it's not being used anywhere? If that's the case, it should be removed from this PR.


The `vice-loading.json` is used by the new vice loading page, which has already been merged.  At least on my diff, it looks like only a new line at the end of the file was added, as it was missing before.

psarando (approved):
> The `vice-loading.json` is used by the new vice loading page, which has already been merged. At least on my diff, it looks like only a new line at the end of the file was added, as it was missing before.

Oh, I see that now. I was viewing the GitHub diff with whitespace changes hidden, and for some reason in that view GitHub was telling me in this file

> 267 additions, 0 deletions not shown because the diff is too large. Please use a local Git client to view these changes.

That made me think the file was added in the PR 😖 

In that case, I only have a minor suggestion about the i18n messages, and this still needs the lockfile rebase fix, otherwise LGTM 👍

src/components/apps/admin/publicationRequests/RequestListing.js
sri2k1us: Yup. They have `:` in  i18n messages. I didn't want to touch lots of unrelated files or dupe it, so I decided to remove it from the message. I can go ahead and add another entry for these.

sri2k1us:
Thanks for the reviews. I have addressed all the PR comments.

PR #321, opened 2021-04-13
CORE-1369 Fix App Launch form cutoff at page bottom

psarando:
This PR will add an `overflow` style to the `PageWrapper` component that wraps all Sonora page content.

This fixes the App Launch form getting cutoff at the bottom of the page, and doesn't appear to affect appearance of other components using it.

The 100% width style of the `AppInfo` title and description were causing horizontal scrolling with the new `overflow` style, but it appears that style is not needed anyway.

Before:
Long App Launch form - before CORE-1369

---

After (notice the scroll bar now in the bottom right):
Long App Launch form - after CORE-1369

### DIFF
--- a/src/components/apps/editor/styles.js
+++ b/src/components/apps/editor/styles.js
@@ -14,8 +14,6 @@ export default (theme) => ({
     },
 
     formContainer: {
-        overflow: "auto",
-        height: "100%",
         padding: theme.spacing(1),
     },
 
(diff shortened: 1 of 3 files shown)


ash3rz (approved):
Tested it a bit locally, LGTM :shipit:

psarando:
Thanks for the review! 👍 

Since this is a critical bug, I'll go ahead and merge it now, but additional reviews still welcome 🙇

PR #322, opened 2021-04-13
Core 1368 Fix Sidebar scrolling

sri2k1us:
- Large Screen 
  
image

- Tour Prompt
image

Note:
Scrollbars were not showing up cos of some markups / sytles with react-joyride lib. I tried putting the `ProductTour` component in different fragments in `AppBar.js` and none them fixed it. So I took a different approach to fix this. I moved the product tour prompt to be a `Dashboard` item and all users will get the tour prompt until it is dismissed. We discussed a while back in UX meeting about providing a tour to all users. So I think this will solve this requirement.


src/components/layout/DrawerItem.js
```diff
@@ -47,17 +46,6 @@ const DrawerItem = (props) => {
...
-                    <img
-                        className={
-                            clsxBase
-                                ? clsx(clsxBase, classes.drawerIcon)
-                                : classes.drawerIcon
-                        }
-                        src={image}
-                        alt={title}
-                    />
-                )}
```
sri2k1us, line 60: Since we use react components for custom SVG images, this is no longer needed.

psarando (approved):
LGTM :shipit: 👍

ash3rz (approved):
LGTM :shipit:

sri2k1us:
Thanks for the reviews.

PR #324, opened 2021-04-14
CORE-702 Update App Editor Param Selector help text display

psarando:
This PR will update the App Editor Param Selector's help text display on larger screens, by increasing the dialog's width and inlining each type's help text, eliminating the help text expansion by the `?` icon (which is how it will still display on mobile screens).

Before, with help text collapsed:
App editor select param - collapsed help - desktop

Note that when the help text was expanded for one of the param types, the user would only be able to choose a parameter by selecting the line containing the param type's name and icon (but not the `?` icon), and the help text below it would display as plain text (and not a part of the selectable area for choosing the parameter):
App editor select param - collapsed help - one expanded - desktop

With these changes, the help text is always displayed along with the param type's name and icon; which also makes the entire block selectable for choosing that parameter, including the help text as well:
App editor select param - inline help - desktop

---

Based on feedback provided by our 411 and sci-analyst teams, the mobile view was also updated to include a brief help message at the top of the dialog, and the `?` icon was replaced with a blue info `i` icon. Also, the items in each sub-list have been alphabetized (in desktop and mobile views), and the `contact support` message has been removed from Reference Genome items (we will no longer support adding to those lists):
App editor select param - mobile


sri2k1us (approved):
:shipit:

psarando:
I think I've addressed all feedback at this point, so I'll merge this one now, but more reviews are still welcome 🙇

PR #325, opened 2021-04-14
CORE-1299 and CORE-1298 Add VICE Urls to nav drawer and notification menu

ash3rz:
Desktop:
image

Mobile:
image

image


public/static/locales/en/common.json
```diff
@@ -26,49 +27,50 @@
...
-    "policies": "Policies",
+    "openDrawer": "Open Drawer",
     "openMenu": "Open Menu",
-    "closeMenu": "Close Menu",
+    "policies": "Policies",
     "refGenomes": "Reference Genomes",
     "refresh": "Refresh",
     "remove": "Remove",
     "required": "Required",
+    "runningVice": "Currently Running VICE Analyses:",
```
ash3rz, line 55: My IDE has a tool to sort lines - this is the only actual addition in this file

src/components/analyses/details/Drawer.js
```diff
@@ -70,14 +70,6 @@ const useStyles = makeStyles((theme) => ({
...
     },
 
-    tabIndicator: {
-        backgroundColor: theme.palette.secondary.main,
-        color: theme.palette.secondary.contrastText,
-    },
-    tabSelected: {
-        color: theme.palette.primary.contrastText,
-        backgroundColor: theme.palette.primary.main,
-    },
```
ash3rz, line 80: I noticed these styles were duplicated for tabs in a few places, so I moved them to a common file with `DETabs`, `DETab`, and `DETabPanel`

src/components/utils/DETabs.js
```diff
@@ -0,0 +1,65 @@
...
+                id={build(tabId, ids.PANEL)}
+                aria-labelledby={tabId}
+            >
+                <Box p={isMobile || dense ? 1 : 3}>{children}</Box>
+            </Typography>
+        </>
+    );
+}
+
+export { DETab, DETabs, DETabPanel };
```
sri2k1us, line 65: Nice 🎉

src/contexts/pushNotifications.js
```diff
@@ -67,8 +67,16 @@ function NotificationsProvider(props) {
...
+            const data = event.data;
+            console.log(data);
+
+            let pushMsg = null;
+            try {
+                pushMsg = JSON.parse(data);
+            } catch (e) {
+                return;
+            }
+            setCurrentNotification(pushMsg);
```
sri2k1us, line 79: 👍🏽

sri2k1us (approved):
LGTM :shipit:  Some nice refactors and clean ups 👍🏽 ✨

public/static/locales/en/common.json
```diff
@@ -26,49 +27,50 @@
...
-    "moveDown": "Move Down",
     "more": "More",
+    "moveDown": "Move Down",
+    "moveUp": "Move Up",
     "newNotificationAriaLabel": "Show New Notifications",
     "next": "Next",
-    "notifications": "Notifications",
     "noNotifications": "No Notifications to display.",
+    "noOptions": "No results to display.",
+    "noRunningVice": "No currently running VICE jobs",
```
sri2k1us, line 43: Might want to use `analyses` instead of `jobs`. May be `No running VICE analyses. `

src/serviceFacades/analyses.js
```diff
@@ -131,6 +151,31 @@ function getAnalysisPermissions({ analyses }) {
...
+function useRunningViceJobs({ enabled, onSuccess, onError }) {
+    return useQuery({
+        queryKey: RUNNING_VICE_JOBS_QUERY_KEY,
+        queryFn: getAnalyses,
+        config: {
+            enabled,
+            onSuccess,
+            onError,
+            staleTime: Infinity,
+            cacheTime: Infinity,
```
psarando, line 174: What is the benefit of using infinite cache and stale time for this query?

I would think those settings should be used when API data isn't changing (or at least not changing within a typical user session), or if we are manually invalidating the query somewhere else; but I don't see an invalidation in this PR, and this kind of data seems like it could change fairly often 🤔

psarando (approved):
Very tidy! ✨ 
LGTM :shipit: 👍

src/serviceFacades/analyses.js
ash3rz: Oops, yeah, this is leftover from a few different refactorings. Initially, I did have an invalidation that would fire when a new analysis notification would come in.  Then Sriram pointed out that the notification data actually has enough information to update the list without hitting the service again. So for this particular feature, it is only fetched once on login, then updated by parsing notifications.

I do think it's confusing to keep the `staleTime` and `cacheTime` settings here though, like you said - I could make them configurable and add them to the point where `useRunningViceJobs` is actually called for this feature.

psarando: >  Then Sriram pointed out that the notification data actually has enough information to update the list without hitting the service again. So for this particular feature, it is only fetched once on login, then updated by parsing notifications.

Ah, that's right, I forgot about that feature 👍

ash3rz:
Thanks for the reviews!

PR #331, opened 2021-04-20
CORE-1329 Update App Launch file/folder input params to validate with API

psarando:
This PR will update App Launch file/folder input params to validate that the path exists with the API.

This allows the App Editor to display "not found" errors for default values, but it is not a form validation error, so the user is still allowed to save the app in this state (currently the same behavior for checkbox and validation rule errors in number and text params).

It should not impact the App Launch form, since the API will remove default inputs that do exist from app launch description responses.

Error message in the Step 2 preview / arrangement view:
App editor - step 2 - preview file not found

---

Error message in the param property edit view:
App editor - param editor - default file not found


src/components/apps/launch/params/DiskResourceSelector.js
```diff
@@ -0,0 +1,84 @@
...
+            onSuccess: () => {
+                setStatError(null);
+            },
+            onError: (error) => {
+                if (ERROR_CODES.ERR_DOES_NOT_EXIST === getErrorCode(error)) {
+                    setStatError(t("errorResourceDoesNotExist"));
+                } else {
+                    setStatError(null);
+                }
+            },
```
psarando, line 52: This is the current behavior in old DE, where only 404 errors are explicitly checked, and everything else is ignored.

ash3rz: 🤔 Seems strange that we would ignore other errors, but if it's been that way in the old DE for however long, I'm guessing it's fine or for a reason. Any final blocking errors would pop up after they tried to launch the app anyway, I assume.

ash3rz (approved):
LGTM :shipit:

src/components/apps/launch/params/DiskResourceSelector.js
psarando: Yeah, this was odd, but I did some testing in old DE, and it seems there can be a case where I can share a private app with `write` perms with you, and if it uses a default input that you don't have read permissions on, it will still display in the app editor without an error message for you (even though this check will return an `ERR_NOT_READABLE`).

I suppose that still makes sense, since the file does exist, only it won't pre-populate in the app launch form if you try to run it.

I did have a generic error message for other errors (such as "Error loading file/folder details. If this path does not exist, the app may not run successfully"), but I removed it in the last commit since old DE doesn't bother 🤷

sri2k1us (approved):
:shipit:

psarando:
Thanks for the reviews! 👍

PR #332, opened 2021-04-20
Pref fix

sri2k1us:
- Fix preference update error happening in Dashboard and File/Folder input selector.
-  Implement `Remember Last path for apps` preference.

### DIFF
--- a/src/components/apps/launch/InputSelector.js
+++ b/src/components/apps/launch/InputSelector.js
@@ -97,17 +97,12 @@ const InputSelector = ({
     const { t } = useTranslation("launch");
     const { t: prefI18n } = useTranslation("preferences");
     const { setFieldValue } = form;
-    const [bootstrapInfo, setBootstrapInfo] = useBootstrapInfo();
+    const bootstrapInfo = useBootstrapInfo()[0];
 
     //update last folder used.
-    const [mutatePreferences] = useSavePreferences(
-        ({ preferences }) => {
-            setBootstrapInfo({ ...bootstrapInfo, preferences });
-        },
-        (e) => {
-            showErrorAnnouncer(prefI18n("savePrefError"), e);
-        }
-    );
+    const [mutatePreferences] = useSavePreferences(null, (e) => {
+        showErrorAnnouncer(prefI18n("savePrefError"), e);
+    });
 
     const inputProps = {
         readOnly: true,
@@ -125,7 +120,7 @@ const InputSelector = ({
                             ...bootstrapInfo.preferences,
                             lastFolder: getParentPath(selection),
                         };
-                        mutatePreferences(updatedPref);
+                        mutatePreferences({ preferences: updatedPref });
                     }}
                 />
             </InputAdornment>
--- a/src/components/apps/launch/params/FileInput.js
+++ b/src/components/apps/launch/params/FileInput.js
@@ -9,11 +9,7 @@ import InputSelector from "../InputSelector";
 
 import ResourceTypes from "components/models/ResourceTypes";
 
-import { useBootstrapInfo } from "contexts/bootstrap";
-
 export default function FileInput({ param, ...props }) {
-    const bootstrapInfo = useBootstrapInfo()[0];
-    const pref = bootstrapInfo?.preferences;
     return (
         <InputSelector
             margin="normal"
@@ -21,11 +17,6 @@ export default function FileInput({ param, ...props }) {
             label={param?.label}
             helperText={param?.description}
             required={param?.required}
-            startingPath={
-                pref?.rememberLastPath
-                    ? pref?.lastFolder
-                    : pref?.defaultFileSelectorPath
-            }
             {...props}
         />
     );
--- a/src/components/apps/launch/params/FileInput.js
+++ b/src/components/apps/launch/params/FileInput.js
@@ -9,7 +9,11 @@ import InputSelector from "../InputSelector";
 
 import ResourceTypes from "components/models/ResourceTypes";
 
+import { useBootstrapInfo } from "contexts/bootstrap";
+
 export default function FileInput({ param, ...props }) {
+    const bootstrapInfo = useBootstrapInfo()[0];
+    const pref = bootstrapInfo?.preferences;
     return (
         <InputSelector
             margin="normal"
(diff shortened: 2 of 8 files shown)


src/components/apps/launch/InputSelector.js
```diff
@@ -97,17 +97,12 @@ const InputSelector = ({
...
     const { t: prefI18n } = useTranslation("preferences");
     const { setFieldValue } = form;
-    const [bootstrapInfo, setBootstrapInfo] = useBootstrapInfo();
+    const bootstrapInfo = useBootstrapInfo()[0];
 
     //update last folder used.
-    const [mutatePreferences] = useSavePreferences(
-        ({ preferences }) => {
-            setBootstrapInfo({ ...bootstrapInfo, preferences });
-        },
```
sri2k1us, line 106: No need to update bootstrap context here. That already happens in `useSavePreferences`

src/components/dashboard/index.js
```diff
@@ -99,47 +95,17 @@ const Dashboard = (props) => {
...
-        } else {
-            setBootstrapQueryEnabled(true);
-        }
-    }, [preProcessData, bootstrapInfo, bootstrapCache]);
-
-    useBootStrap(
-        bootstrapQueryEnabled,
-        (respData) => preProcessData(respData),
-        setBootstrapError
-    );
```
sri2k1us, line 131: This code is not needed at all. I updated AppBar to update bootstrap context when it fetches it for the first time.

src/components/layout/AppBar.js
```diff
@@ -154,12 +156,12 @@ function DEAppBar(props) {
...
     useEffect(() => {
-        if (userProfile) {
+        if (userProfile && !bootstrapInfo) {
             setBootstrapQueryEnabled(true);
         }
-    }, [userProfile]);
+    }, [userProfile, bootstrapInfo]);
 
-    useBootStrap(bootstrapQueryEnabled, null, setBootstrapError);
+    useBootStrap(bootstrapQueryEnabled, setBootstrapInfo, setBootstrapError);
```
sri2k1us, line 164: While we had bootstrap context, we never put it to full use by setting it up when bootstrap was fetched from services. This will fix it.

src/serviceFacades/users.js
```diff
@@ -40,14 +40,21 @@ function bootstrap(key) {
...
         method: "POST",
         body: preferences,
     });
-    const hookPromise = updateWebhooks(webhooks);
+    promises.push(prefPromise);
+
+    if (webhooks) {
+        const hookPromise = updateWebhooks(webhooks);
+        promises.push(hookPromise);
+    }
```
sri2k1us, line 55: Update webhooks only when needed.

src/serviceFacades/users.js
```diff
@@ -130,10 +137,15 @@ function useSavePreferences(onSuccessCallback, onError) {
         onSuccess: (updatedPref) => {
             //update preference in cache
             queryCache.setQueryData(BOOTSTRAP_KEY, (bootstrapData) => {
-                if (bootstrapData && updatedPref) {
+                if (
+                    bootstrapData &&
+                    updatedPref &&
+                    updatedPref[0].preferences
+                ) {
+                    const updatePref = updatedPref[0].preferences;
```
sri2k1us, line 145: The response is in an array now cos we do a `Promise.all` 🤔 . Not sure about this...

src/components/apps/launch/params/FileInput.js
```diff
@@ -9,14 +9,23 @@ import InputSelector from "../InputSelector";
...
             margin="normal"
             acceptedType={ResourceTypes.FILE}
             label={param?.label}
             helperText={param?.description}
             required={param?.required}
+            startingPath={
+                pref?.rememberLastPath
+                    ? pref?.lastFolder
+                    : pref?.defaultFileSelectorPath
+            }
```
psarando, line 28: This isn't needed here, since the `InputSelector` is already using logic like this in its call to `useSelectorDefaultFolderPath`.

src/components/apps/launch/params/FolderInput.js
```diff
@@ -9,14 +9,23 @@ import InputSelector from "../InputSelector";
...
             margin="normal"
             acceptedType={ResourceTypes.FOLDER}
             label={param?.label}
             helperText={param?.description}
             required={param?.required}
+            startingPath={
+                pref?.rememberLastPath
+                    ? pref?.lastFolder
+                    : pref?.defaultFileSelectorPath
+            }
```
psarando, line 28: Also not needed here, since the `InputSelector` should already be doing this internally when `startingPath` is not set.

psarando (changes requested):
I don't think I was aware of the `defaultFileSelectorPath` preference before. Maybe it should be added to the logic in `useSelectorDefaultFolderPath`?

sri2k1us:
> I don't think I was aware of the `defaultFileSelectorPath` preference before. Maybe it should be added to the logic in `useSelectorDefaultFolderPath`?

Same here...But I was looking at the `MockBootstrap.js` and I found it in mock response. I could remove this and just use home folder as default.

src/components/apps/launch/params/FileInput.js
sri2k1us: Oops! Missed that. I guess I tested it without checking if the actual preference was set to remember last path 🤔 .

src/components/apps/launch/InputSelector.js
```diff
@@ -97,17 +97,12 @@ const InputSelector = ({
     const { t } = useTranslation("launch");
     const { t: prefI18n } = useTranslation("preferences");
     const { setFieldValue } = form;
-    const [bootstrapInfo, setBootstrapInfo] = useBootstrapInfo();
+    const bootstrapInfo = useBootstrapInfo()[0];
```
psarando, line 100: No need to change this (makes sense as-is), but this also works:

```suggestion
    const [bootstrapInfo] = useBootstrapInfo();
```

psarando (approved):
LGTM! 👍

ash3rz (approved):
LGTM :shipit:

sri2k1us:
Thanks for the reviews.

PR #333, opened 2021-04-22
Adjust PathListAutomation.js

flynn1999:
Add DialogContent & DialogActions for css bug of Cancel & Done button not scrolling. Still not scrolling after adding needed components and using scroll from [Material-UI docs](https://material-ui.com/components/dialogs/#ScrollDialog.js) 😖

### DIFF
--- a/src/components/data/PathListAutomation.js
+++ b/src/components/data/PathListAutomation.js
@@ -51,14 +51,22 @@ import {
 } from "@material-ui/core";
 
 import Skeleton from "@material-ui/lab/Skeleton";
+import DEDialog from "../utils/DEDialog";
+import DialogContent from "@material-ui/core/DialogContent";
+import DialogActions from "@material-ui/core/DialogActions";
 
 export default function PathListAutomation(props) {
     const {
+        scroll,
         baseId,
         requestedInfoType,
         onCreatePathList,
         onCancel,
         startingPath,
+        open,
+        onClose,
+        title,
+        maxWidth,
     } = props;
 
     const theme = useTheme();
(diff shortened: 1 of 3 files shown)


src/components/data/PathListAutomation.js
```diff
@@ -193,246 +201,295 @@ export default function PathListAutomation(props) {
...
-                                    label={t("suggestionSelection_any_plural")}
-                                />
-                            </Grid>
-                            <Grid item xs>
-                                <Field
-                                    id={buildId(
-                                        baseId,
-                                        ids.PATH_LIST_AUTO_FOLDERS_ONLY_SWITCH
+        <>
+            <DEDialog
```
ash3rz, line 205: I haven't done a deep review, but 3 things:
* If you switch to using DEDialog, then you cannot use `DialogContent` or `DialogActions` directly.  
  * You can see the component [here](https://github.com/cyverse-de/sonora/blob/master/src/components/utils/DEDialog.js). 
  * It's already using `DialogContent` and `DialogActions` for you, as well as setting up `DialogTitle`. 

* If you recall from our zoom call, we were looking at [this](https://github.com/cyverse-de/sonora/blob/master/src/components/data/CreateFolderDialog.js#L77:L90) as an example to follow.  Formik, then Form, then Dialog. OR if you want to continue to use `DEDialog`, then [here's](https://github.com/cyverse-de/sonora/blob/master/src/components/doi/UpdateRequestDialog.js#L88:L105) a similar example.  Formik, then Form, then DEDialog.

* If you've set up the `Dialog` or `DEDialog` correctly, you don't need to add anything else to get it to scroll.  It will scroll automatically by default.

Hope this helps!

flynn1999: I've replaced the appropriate components but it still does not show this error message when `Done` is clicked although it highlights it red since its marked as required. 
 
Screen Shot 2021-04-23 at 11 02 21 AM

src/components/data/PathListAutomation.js
```diff
@@ -193,246 +200,293 @@ export default function PathListAutomation(props) {
...
+                validate={validate}
+                initialValues={{
+                    selectedPaths: [],
+                    dest: startingPath,
+                    fileName: "",
+                    pattern: "",
+                    foldersOnly: false,
+                }}
+            >
+                {({ handleSubmit }) => {
```
ash3rz, line 215: To get the validations to run, you'll need to use this `handleSubmit` function (hopefully your IDE was showing you that this was not being used or called anywhere).  You'll note on your `Submit` button, there's no `onClick` defined.  You want this function to run when the `Submit` button is clicked

src/components/data/PathListAutomation.js
```diff
@@ -51,14 +50,22 @@ import {
...
 
 import Skeleton from "@material-ui/lab/Skeleton";
+import DialogContent from "@material-ui/core/DialogContent";
+import DialogActions from "@material-ui/core/DialogActions";
+import Dialog from "@material-ui/core/Dialog";
+import DialogTitle from "@material-ui/core/DialogTitle";
 
 export default function PathListAutomation(props) {
     const {
+        scroll,
```
ash3rz, line 60: You can remove this

src/components/data/PathListAutomation.js
flynn1999: As pointed out by Ashley this message comes from the browser: https://stackoverflow.com/questions/5392882/why-is-chrome-showing-a-please-fill-out-this-field-tooltip-on-empty-fields

src/components/data/PathListAutomation.js
```diff
@@ -193,246 +199,292 @@ export default function PathListAutomation(props) {
...
-                                    height="25vh"
-                                    label={t("suggestionSelection_any_plural")}
-                                />
-                            </Grid>
-                            <Grid item xs>
-                                <Field
-                                    id={buildId(
-                                        baseId,
-                                        ids.PATH_LIST_AUTO_FOLDERS_ONLY_SWITCH
+        <>
```
ash3rz, line 202: This fragment isn't needed

src/components/data/PathListAutomation.js
```diff
@@ -193,246 +199,292 @@ export default function PathListAutomation(props) {
...
+                    selectedPaths: [],
+                    dest: startingPath,
+                    fileName: "",
+                    pattern: "",
+                    foldersOnly: false,
+                }}
+            >
+                {({ handleSubmit }) => {
+                    return (
+                        <>
```
ash3rz, line 216: Also this fragment can be removed

src/components/data/PathListAutomation.js
```diff
@@ -193,246 +199,292 @@ export default function PathListAutomation(props) {
...
-                        >
-                            {createPathListError && (
-                                <Grid item xs>
-                                    <ErrorTypographyWithDialog
-                                        baseId={baseId}
-                                        errorMessage={errorMsg}
-                                        errorObject={createPathListError}
-                                    />
-                                </Grid>
-                            )}
```
ash3rz, line 408: Looks like this was accidentally removed?

ash3rz (approved):
Looks like one of the sections that would display an error was accidentally removed - otherwise, LGTM 👍

src/components/data/PathListAutomation.js
```diff
@@ -193,246 +199,292 @@ export default function PathListAutomation(props) {
...
-                                        {!isFetching &&
-                                            infoTypes &&
-                                            infoTypes.length > 0 &&
-                                            infoTypes.map((type) => {
-                                                const labelId = `checkbox-list-label-${type}`;
-                                                return (
-                                                    <ListItem
-                                                        key={type}
+                                    <DialogContent>
+                                        <>
```
psarando, line 239: This Fragment isn't necessary (`DialogContent` can already contain multiple children).

psarando (commented):
I also caught one more `<>` that can be removed.

src/components/data/PathListAutomation.js
flynn1999: I have this in lines 247 - 256 but I'll move it to its previous location

ash3rz: Ah, it's not that it needs to be moved - it's that it exists in 2 places intentionally, and you've removed 1 of the occurrences.

src/components/data/PathListAutomation.js
```diff
@@ -193,246 +199,276 @@ export default function PathListAutomation(props) {
...
+                                </DialogContent>
+                                <DialogActions>
+                                    {createPathListError && (
+                                        <Grid item xs>
+                                            <ErrorTypographyWithDialog
+                                                baseId={baseId}
+                                                errorMessage={errorMsg}
+                                                errorObject={
+                                                    createPathListError
+                                                }
```
ash3rz, line 441: 🤔 I see why you added this here - that appears to be how it was originally added.  I suspect it was intended to go within `DialogContent` though and was missed during review. The idea being that if there's an error, you can see it at the top and bottom of the dialog content, so it's harder to miss.

sri2k1us: Putting it along the action buttons in a `Grid`, will align the error message with the action buttons. At least that what I remember.

src/components/data/PathListAutomation.js
```diff
@@ -145,16 +151,10 @@ export default function PathListAutomation(props) {
...
-            pattern,
-            dest,
-            fileName,
-            foldersOnly,
-        } = values;
+        const { selectedPaths, pattern, dest, fileName, foldersOnly } = values;
 
         const submission = {
-            paths: selectedResources?.map((resource) => resource?.path),
+            paths: selectedPaths,
```
ash3rz, line 157: Ah no!  This file got updated while you were out and looks your rebase reverted those changes (this might be easier to see here in the PR compared to Fork?).  We'll need those old changes put back in while keeping your dialog related changes.

ash3rz (changes requested):

src/components/data/PathListAutomation.js
flynn1999: I've added the previous changes back.

src/components/data/PathListAutomation.js
```diff
@@ -213,229 +219,252 @@ export default function PathListAutomation(props) {
...
-                                </Paper>
-                            </Grid>
-                        </Grid>
-                        <Grid
-                            container
-                            direction="row"
-                            justify="flex-end"
-                            alignItems="flex-end"
-                            spacing={1}
-                        >
```
ash3rz, line 405: Looks like this `Grid` bit got missed when adding this to the `DialogActions` section

src/components/data/PathListAutomation.js
```diff
@@ -213,229 +219,252 @@ export default function PathListAutomation(props) {
...
-                                    />
+                                        color="primary"
+                                        type="submit"
+                                        onClick={handleSubmit}
+                                    >
+                                        {i18nCommon("done")}
+                                    </Button>
                                 </Grid>
-                            )}
-                            <Grid item>
```
ash3rz, line 415: This `item` got removed from the tag

src/components/data/PathListAutomation.js
```diff
@@ -213,229 +219,252 @@ export default function PathListAutomation(props) {
...
-                                    id={buildId(
-                                        baseId,
-                                        ids.PATH_LIST_AUTO_CANCEL_BTN
-                                    )}
-                                    onClick={onCancel}
-                                >
-                                    {i18nCommon("cancel")}
-                                </Button>
-                            </Grid>
-                            <Grid item>
```
ash3rz, line 426: This `item` got removed from the tag

src/components/data/PathListAutomation.js
```diff
@@ -213,229 +219,252 @@ export default function PathListAutomation(props) {
...
-                                        baseId,
-                                        ids.PATH_LIST_AUTO_DONE_BTN
-                                    )}
-                                    color="primary"
-                                    type="submit"
-                                >
-                                    {i18nCommon("done")}
-                                </Button>
-                            </Grid>
-                        </Grid>
```
ash3rz, line 438: When you add back in the `Grid` portion from above, you'll also need to add this closing tag

ash3rz (commented):
Some final stragglers to take care of, then I think this'll be good to go 👍

ash3rz (approved):
:shipit: LGTM

PR #334, opened 2021-04-23
CORE-1363 Add pending termination dlg

ash3rz:
This displays when users try to go to an output folder (from either a button or menu item in the Analyses view) that doesn't yet exist because the analysis is still running.

image


src/components/analyses/PendingTerminationDlg.js
```diff
@@ -0,0 +1,46 @@
...
+import React from "react";
+
+import { build } from "@cyverse-de/ui-lib";
+import { Button, Typography } from "@material-ui/core";
+
+import DEDialog from "components/utils/DEDialog";
+import { useTranslation } from "i18n";
+import ids from "./ids";
+
+function PendingTerminationDlg(props) {
```
ash3rz, line 17: Not sure if "Pending Termination" makes sense to anybody else - open to suggestions on the name

src/components/analyses/listing/Actions.js
```diff
@@ -49,16 +50,29 @@ const RelaunchButton = React.forwardRef((props, ref) => {
...
-            href={href}
-            onClick={onClick}
-            ref={ref}
+            href={isTerminated ? href : null}
+            ref={isTerminated ? ref : null}
+            onClick={(event) => {
+                if (isTerminated) {
+                    onClick();
+                } else {
+                    setPendingTerminationDlgOpen(true);
```
ash3rz, line 73: I would have used `stopPropagation` and `preventDefault` here, but then the row doesn't get selected, which means I don't have access to the analysis name. So I opted for disabling the `href` and `ref` instead

src/components/analyses/utils.js
```diff
@@ -210,6 +210,22 @@ const canShare = (selectedAnalyses) => {
     );
 };
 
+const ANALYSIS_TERMINAL_STATES = [
+    analysisStatus.CANCELED,
+    analysisStatus.COMPLETED,
+    analysisStatus.FAILED,
+    analysisStatus.REMOVED,
```
ash3rz, line 217: I'm not 100% sure if these are all the terminal states from [here](https://github.com/cyverse-de/sonora/blob/master/src/components/models/analysisStatus.js)

sri2k1us: LGTM 👍🏽

sri2k1us (approved):
I have 1 suggestion about the message displayed to the user. But LGTM 👍🏽

psarando (approved):
LGTM :shipit: 👍

ash3rz:
I just realized I forgot to include this change for the dashboard action as well.  I'll make that change before merging.

ash3rz:
Thanks for the reviews!

PR #335, opened 2021-04-27
CORE-1378 and CORE-1379 Completed analysis notifications take you to the output folder

ash3rz:
... and the notification message has been updated to say "Access your output files here." similar to the VICE running notification.

NOTE: This is on top of #334, so the first commit can be ignored

Screenshot (I'm hovering over the completed notification and you can see the preview URL in the bottom left is going to the data page)
image

### DIFF
--- a/src/components/notifications/Message.js
+++ b/src/components/notifications/Message.js
@@ -18,7 +18,10 @@ import {
 import NotificationCategory from "components/models/NotificationCategory";
 import SystemId from "components/models/systemId";
 
-import { getAnalysisDetailsLinkRefs } from "components/analyses/utils";
+import {
+    getAnalysisDetailsLinkRefs,
+    isTerminated,
+} from "components/analyses/utils";
 import { getAppListingLinkRefs } from "components/apps/utils";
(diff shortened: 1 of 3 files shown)


sri2k1us (approved):
:shipit:

src/components/notifications/Message.js
```diff
@@ -46,10 +49,13 @@ function AnalysisLink(props) {
     const { notification } = props;
     const { t } = useTranslation("common");
     const message = getDisplayMessage(notification);
-    const action = notification.payload?.action;
+    const payload = notification.payload;
```
psarando, line 52: Since we know this notification's payload is an analysis at this point, I would refactor this to:
```suggestion
    const analysis = notification.payload;
```
I think that would make more sense further below as well (`analysis?.status === analysisStatus.COMPLETED` rather than `payload?.status === analysisStatus.COMPLETED` etc).

psarando (approved):
LGTM 👍

ash3rz:
Thanks for the reviews!

PR #338, opened 2021-04-28
Core 716 Allow users to submit request to publish their app

sri2k1us:
- Large Screen Form
image

- iPhone 5
image


src/components/apps/PublishAppDialog.js
```diff
@@ -0,0 +1,365 @@
...
+            requestAppPublication({
+                systemId: app?.system_id,
+                appId: app?.id,
+                request,
+            });
+        }
+    };
+
+    const validateTestDataPath = (value) => {
+        if (!value.includes(communityPath)) {
```
psarando, line 99:
```suggestion
        if (!value.startsWith(communityPath)) {
```

src/components/apps/PublishAppDialog.js
```diff
@@ -0,0 +1,365 @@
...
+        }
+        return;
+    };
+
+    return (
+        <Formik
+            initialValues={{
+                name: app?.name,
+                description: app?.description,
+                system_id: app?.system_id,
```
psarando, line 110: It doesn't look like the form is using `system_id` (the `handleSubmit` function is directly using `app?.system_id`).
```suggestion
```

src/components/apps/PublishAppDialog.js
```diff
@@ -0,0 +1,365 @@
...
+                                />
+                            )}
+                            {error && (
+                                <ErrorTypographyWithDialog
+                                    errorMessage={t("publicationRequestFailed")}
+                                    errorObject={error}
+                                />
+                            )}
+                            <Field
+                                name={"name"}
```
psarando, line 169:
```suggestion
                                name="name"
```

src/components/apps/PublishAppDialog.js
```diff
@@ -0,0 +1,365 @@
...
+                                        appsConstants.APP_NAME_RESTRICTED_CHARS,
+                                        value,
+                                        t
+                                    )
+                                }
+                                required={true}
+                                component={FormTextField}
+                            />
+                            <Field
+                                name={"description"}
```
psarando, line 184:
```suggestion
                                name="description"
```

src/components/apps/PublishAppDialog.js
```diff
@@ -0,0 +1,365 @@
...
+                                acceptedType={ResourceTypes.FOLDER}
+                                label={t("testDataLocation")}
+                                component={FolderInput}
+                                required={true}
+                                validate={(value) =>
+                                    validateTestDataPath(value)
+                                }
+                            />
+                            <Field
+                                name={"inputDesc"}
```
psarando, line 205:
```suggestion
                                name="inputDesc"
```

src/components/apps/PublishAppDialog.js
```diff
@@ -0,0 +1,365 @@
...
+                                label={t("inputDescLabel")}
+                                id={build(parentId, ids.PUBLISH.INPUT)}
+                                required={true}
+                                component={FormMultilineTextField}
+                                validate={(value) =>
+                                    nonEmptyField(value, i18nUtil)
+                                }
+                            />
+                            <Field
+                                name={"parameterDesc"}
```
psarando, line 215:
```suggestion
                                name="parameterDesc"
```

src/components/apps/PublishAppDialog.js
```diff
@@ -0,0 +1,365 @@
...
+                                label={t("parameterDescLabel")}
+                                id={build(parentId, ids.PUBLISH.PARAMS)}
+                                required={true}
+                                component={FormMultilineTextField}
+                                validate={(value) =>
+                                    nonEmptyField(value, i18nUtil)
+                                }
+                            />
+                            <Field
+                                name={"outputDesc"}
```
psarando, line 225:
```suggestion
                                name="outputDesc"
```

src/components/apps/PublishAppDialog.js
```diff
@@ -0,0 +1,365 @@
...
+                                                                    index,
+                                                                    ids.PUBLISH
+                                                                        .LINK_ADD_BTN
+                                                                )}
+                                                                onClick={() =>
+                                                                    arrayHelpers.insert(
+                                                                        index,
+                                                                        ""
+                                                                    )
+                                                                } // insert an empty string at a position
```
psarando, line 331: 👍

psarando (commented):
Looks good 👍 🎉 

I have some style, clean-up, and help text suggestions, mostly minor.

Should we also import the help text that displays at the top of the dialog in the old DE? It seems like important info, especially for users that haven't published an app before (though it may need some tweaks for `sonora`):

Publish Dialog help text

https://github.com/cyverse-de/ui/blob/06c9fa859c8a366de30a8afa201231bee333eb50/de-lib/src/main/java/org/iplantc/de/theme/base/client/apps/AppsMessages.properties#L24

sri2k1us:
> Looks good 👍 🎉
> 
> I have some style, clean-up, and help text suggestions, mostly minor.
> 
> Should we also import the help text that displays at the top of the dialog in the old DE? It seems like important info, especially for users that haven't published an app before (though it may need some tweaks for `sonora`):
> 
> Publish Dialog help text
> 
> https://github.com/cyverse-de/ui/blob/06c9fa859c8a366de30a8afa201231bee333eb50/de-lib/src/main/java/org/iplantc/de/theme/base/client/apps/AppsMessages.properties#L24

Good catch. I forgot adding that text. I was saving it for the last.

public/static/locales/en/apps.json
```diff
@@ -14,6 +15,7 @@
     "appType": "Type",
     "ariaCheckbox": "{{label}} checkbox",
     "ariaTableListing": "Apps table listing",
+    "attributionLinks": "Please add any attribution link(s) to this app or for the tool(s) used by this app:",
```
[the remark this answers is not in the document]
sri2k1us: Hmm this text is not used as a field label actually. This is a text prompt using Typography.
See Screenshot attached above.

public/static/locales/en/apps.json
```diff
@@ -60,11 +62,14 @@
     "hpc": "High-Performance Computing",
     "htcondorExtraRequirements": "HTCondor Extra Requirements",
     "imageLabel": "Image",
+    "inputDescLabel": "Briefly describe input file(s) required by this app like number of file(s) required, file format accepted etc...",
```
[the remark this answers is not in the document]
sri2k1us: I like breaking it down like this 👍🏽

sri2k1us:
Note:  Nirav has requested to collect licensing info from the user when submitting app for public use. I am going to add it in this PR.

public/static/locales/en/apps.json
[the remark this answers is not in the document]
psarando: Ah yes, oops, I forgot the screenshot when reviewing this text.
Looks fine then 👍

sri2k1us:
Should be ready for review with licensing fields now.

src/components/apps/PublishAppDialog.js
psarando: Should this validation be updated with `startsWith`? Otherwise it could consider a path like `/iplant/home/psarando/iplant/home/shared/my_secret_data` as valid 🤔

src/components/apps/PublishAppDialog.js
```diff
@@ -0,0 +1,409 @@
...
+                                        <Link
+                                            href="#"
+                                            component="button"
+                                            onClick={(event) => {
+                                                // prevent form submission
+                                                event.preventDefault();
+                                                intercomShow();
+                                            }}
+                                        />
+                                    ),
```
psarando, line 189: This `Link` could probably use a `variant="body2"` prop, so that the `button` font matches the style of the surrounding text.

public/static/locales/en/util.json
```diff
@@ -29,5 +29,6 @@
     "timestamp": "Timestamp",
     "unexpectedError": "This wasn't supposed to happen. Please try again or contact support!",
     "user": "User",
-    "viewDetails": "View Details"
+    "viewDetails": "View Details",
+    "inValidUrl": "A valid URL must begin with either ftp or http or https."
```
psarando, line 33: Missed this in my last review, but because "invalid" is one word, I would name this key `invalidUrl`.

psarando (approved):
LGTM 👍 

I have a couple more minor suggestions, and I still recommend updating the `validateTestDataPath`.

src/components/apps/PublishAppDialog.js
```diff
@@ -0,0 +1,409 @@
...
+                            <Typography>{t("attributionLinks")}</Typography>
+                            <FieldArray
+                                name="references"
+                                render={(arrayHelpers) => (
+                                    <Grid container spacing={3}>
+                                        {values.references &&
+                                        values.references.length > 0 ? (
+                                            values.references.map(
+                                                (link, index) => (
+                                                    <>
```
ash3rz, line 287: This fragment can be removed - should help the components from looking so squished in the file

src/components/apps/PublishAppDialog.js
sri2k1us: 😈

src/components/apps/PublishAppDialog.js
psarando: I don't think this one can be removed, since this `map` is returning multiple `Grid` items.

sri2k1us: ✅ . May be I can extract this into its own component.

ash3rz: Ah, you're right. I think because it's squished, I thought it was all under one grid. I need vertical helper lines like in my IDE 😄

PR #339, opened 2021-04-28
Switch to a mutation for the instant launch API call.

johnworth:
SSIA.

Very similar to the suggestion @ash3rz made in the PR that added instant launches.

I attempted to simplify the `setOpen(false)` so they were in the `onSettled` handler or in a `try...finally`, but in both cases there was an unacceptable delay between the dialog closing at the `router.push()` call taking effect, leaving the user an opportunity to click around and screw up the flow. Ultimately I ended up leaving things as they are.


psarando (approved):
LGTM 👍

ash3rz (approved):
:shipit: 
Thanks for digging into this!

PR #341, opened 2021-04-29
Switch to the <Formik> tag for consistency

johnworth:
Changes the InstantLaunchMapping component from the useFormik hook to the `Formik` tag.

No other changes have been made.

### DIFF
--- a/src/components/instantlaunches/admin/InstantLaunchMapping.js
+++ b/src/components/instantlaunches/admin/InstantLaunchMapping.js
@@ -44,7 +44,7 @@ import {
 import { Add as AddIcon, Delete as DeleteIcon } from "@material-ui/icons";
 
 import { useTranslation } from "i18n";
-import { useFormik } from "formik";
+import { Formik } from "formik";
 
 const useStyles = makeStyles((theme) => ({
     title: {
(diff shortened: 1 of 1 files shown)


src/components/instantlaunches/admin/InstantLaunchMapping.js
```diff
@@ -92,140 +92,141 @@ const initialValues = {
...
-                    {infoTypes.map((infoType) => (
+                    <TextField
+                        id={buildID(baseID, ids.NAME)}
+                        name="mappingName"
+                        label={t("common:name")}
+                        className={classes.flexItem}
+                        value={values.mappingName}
+                        onChange={handleChange}
+                        error={touched.mappingName && errors.mappingName}
+                        helperText={touched.mappingName && errors.mappingName}
```
psarando, line 111: All of these `value`, `onChange`, `error`, and `helperText` props can go away if this `TextField` is replaced by a formik `FastField` with a `component={FormTextField}` prop (and `FormTextField` is imported from `ui-lib`).

The `<FastField ... component={FormTextField}>` also handles `onBlur` events, so if a field is required, and the user enters it, then exits it without making any changes (i.e. if the field is "touched"), then it can display a `required` validation error message (though I don't see any required fields in this form).

So all of the `TextField` components in this form can be refactored the same way (even the `select` field).

ash3rz (commented):
👍 Nothing to add other than what @psarando already said

johnworth:
Ill aim to work on this Monday after my appointment.

johnworth:
lol, I must have completely spaced this. Sorry. It's so out of date that I'm just going to go ahead and close it.

PR #342, opened 2021-04-30
Open IL in a new tab.

sri2k1us:


src/components/instantlaunches/index.js
```diff
@@ -101,24 +100,28 @@ const InstantLaunchButton = ({
...
     const theme = useTheme();
 
+    React.useEffect(() => {
+        if (ilUrl) {
+            window.open(ilUrl);
+            setIlUrl(null);
+            setOpen(false);
+        }
+    }, [ilUrl]);
+
```
sri2k1us, line 114: useEffect will ensure that `window` object is always available.

johnworth (approved):
LGTM. I know you're working on some text to tell users to disable their pop-up blockers for Sonora, but otherwise I think everything here is fine.

sri2k1us:
Thanks for the review. Additional reviews welcome.

PR #343, opened 2021-05-03
core-1399 Improve Analysis info UX.

sri2k1us:
- Large Screen:
   - Remove collapse feature. Use `List` instead of `Table`
   - Sort Updates in descending order of timestamp so that latest updates is on top
image

- iPhone 5
image

PS: Ran this change by Amanda and she likes this UX change.

### DIFF
--- a/src/components/analyses/details/InfoPanel.js
+++ b/src/components/analyses/details/InfoPanel.js
@@ -97,7 +97,6 @@ function Step(props) {
                 <Updates
                     updates={updates}
                     baseId={build(baseId, ids.INFO.UPDATE)}
-                    t={t}
                 />
             </AccordionDetails>
         </Accordion>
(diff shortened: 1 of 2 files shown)


slr71 (commented):
This looks good to me, although I'm not a UI expert.

src/components/analyses/details/InfoPanel.js
```diff
@@ -162,7 +94,11 @@ function Step(props) {
                         btnText={t("copyAnalysisId")}
                     />
                 </div>
-                <Updates updates={updates} baseId={baseId} t={t} />
+                <Updates
+                    updates={updates}
+                    baseId={build(baseId, ids.INFO.UPDATE)}
+                    t={t}
```
ash3rz, line 100: Looks like this `t` prop is passed to `Updates` but that component is not using it

ash3rz (approved):
:shipit:

sri2k1us:
Thanks for the reviews.

PR #344, opened 2021-05-03
CORE-1398 Fix display of App Editor Selection param default values

psarando:
This PR will fix some display issues in the App Editor Selection param default values.

The App Editor will now add unique keys to selection list items in the form, which will allow the Selection param property editor to reset its default value to the actual instance of the list item with the matching key.

Before this fix, the Selection param property editor's "Default value" field would not update as changes were made to list items (since the common `DefaultValueField` uses a `FastField`). Also, when editing the "Display", "Argument", or "Value" field of a list item that is also selected as the parameter's default value, then the preview list at the top of the form would appear to clear the selected value (it uses a form `Field` so it updates right away, but editing the selected list item causes formik to create a new instance of that item which was not getting reset in the preview field). 

For example, in the following Selection param property editor "New Argument 2" is the list item selected as the default value:
App Editor - Selection param property editor

The display bugs in this editor are shown the following screenshot, after simply editing the default value's "Display" label to "Edited Argument 2":
App Editor - Selection param property editor display bug

This is only a display bug in this editor form, and does not affect the actual default value when saving the app.

### DIFF
--- a/src/components/apps/editor/params/SelectionPropertyFields.js
+++ b/src/components/apps/editor/params/SelectionPropertyFields.js
@@ -6,12 +6,11 @@
  */
 import React from "react";
 
-import { FastField, FieldArray } from "formik";
+import { FastField, Field, FieldArray } from "formik";
 import { Trans } from "react-i18next";
 
 import { useTranslation } from "i18n";
 
-import DefaultValueField from "./common/DefaultValueField";
 import DescriptionField from "./common/DescriptionField";
 import ExcludeArgumentField from "./common/ExcludeArgumentField";
 import LabelField from "./common/LabelField";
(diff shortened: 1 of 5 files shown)


src/components/apps/editor/params/SelectionPropertyFields.js
```diff
@@ -196,37 +195,74 @@ function SelectionItemEditor(props) {
...
+    // and this default value editor gets easy access to formik's setFieldValue.
+    React.useEffect(() => {
+        const foundDefault =
+            paramArguments?.find((paramArg) => paramArg.key === value?.key) ||
+            "";
+
+        if (foundDefault !== value) {
+            setFieldValue(name, foundDefault);
+        }
+    }, [paramArguments, name, value, setFieldValue]);
```
ash3rz, line 216: 👍

ash3rz (approved):
LGTM :shipit:

psarando:
Thanks for the review! 👍

PR #345, opened 2021-05-05
CORE-1401 Reference Gemone paramter improvements in App Editor and Launcher

psarando:
This PR will allow users to clear reference genome param selections in the App Launcher, if that parameter is not marked as required. This also allows app integrators to clear default selections previously set in the App Editor.

This PR also improves the loading display of Reference Genome params by using cached data instead of relying on `onSuccess` callbacks, and increasing the query's stale time so that the list is fetched less often (since this list is no longer intended to be updated).


src/components/apps/launch/ReferenceGenomeSelect.js
```diff
@@ -53,6 +57,7 @@ const ReferenceGenomeSelect = ({ param, ...props }) => {
             margin="normal"
             {...selectProps}
         >
+            {!selectProps.required && <MenuItem value="">&nbsp;</MenuItem>}
```
psarando, line 60: This 1-liner allows users to clear a selection when the param is not required, but also allows app integrators to clear a default value setting.

It's the same strategy used to allow app integrators to clear the default selection for `List` params.

Though I just realized that users cannot clear a selection for optional `List` params in the App Launch form. I can also add that improvement to this PR.

psarando: ✅ done

psarando: ✅ done

src/components/apps/queries.js
```diff
@@ -43,17 +43,20 @@ export function useAppElementInfoTypes(enabled, onSuccess, onError) {
     });
 }
 
+export function sortReferenceGenomes(genomes) {
+    return stableSort(genomes, (a, b) => a.name.localeCompare(b.name));
+}
+
 export function useReferenceGenomes(enabled, onSuccess, onError) {
     return useQuery({
-        queryKey: [REFERENCE_GENOMES_QUERY_KEY, { deleted: false }],
```
sri2k1us, line 48: Do we want the deleted genomes as well ? or just the service returns genomes that not deleted by default ?

src/components/apps/queries.js
```diff
@@ -43,17 +43,20 @@ export function useAppElementInfoTypes(enabled, onSuccess, onError) {
...
+        queryKey: REFERENCE_GENOMES_QUERY_KEY,
         queryFn: getReferenceGenomes,
         config: {
             enabled,
+            staleTime: 1000 * 60 * 5, // 5 minutes in milliseconds.
```
sri2k1us, line 56: can't we just cache it for ever (Infinity) ?

src/components/apps/queries.js
sri2k1us: Ahh nm, i see you updated the key below.

src/components/apps/queries.js
sri2k1us: I say this because, our sci team is not actively maintaining this reference genome list. So chance of them getting updated is really really at this point.

sri2k1us (approved):
Code LGTM. I have suggestion of caching ref. genomes.

src/components/apps/queries.js
psarando: We could, and I was wondering about that. The only reason I was hesitant to use infinite cache and stale time, is that we already have an admin UI in `sonora` to allow updates.

So with this 5 minutes stale time, I think a user would only have to wait 5 minutes to see a rare update to this list (maybe a label update, or even deleting one). Otherwise, with infinite cache and stale times, I guess it would require a rolling restart of the UI to reflect the changes to users (since they are cached on the server)? If we don't care about that, we can set the cache to infinite, but I also thought a quick loading mask every 5 minutes was no big deal either. I'm fine with either option, really 🤷

psarando: Oh, but if we invalidate this query in the admin component when a ref genome is updated/deleted, then we can safely use infinite cache for this list:
https://github.com/cyverse-de/sonora/blob/1e3ca3dfcfcf5e109fe56fe24bc7597736c9c6cb/src/components/apps/admin/referenceGenomes/ReferenceGenomes.js#L65-L75

I'll make that update in the PR as well, then.

sri2k1us: My understanding of `react-query` is that caching doesn't happen on the server. It is in memory cache and doesn't require server restarts. When the browser tab is closed / refreshed, I think everything is wiped out. You can test with `bootstrap`. When you login, `bootstrap` will be fetched once but after that it is fetched from cache when needed. When you close tab and do it again, bootstrap will be fetched again.
See - https://react-query-v2.tanstack.com/docs/guides/caching
So invalidating it here admin side will no effect on the user side of things for a different user but I see no harm in invalidating it either.

ash3rz: This was also my understanding.  
As for 5 minutes or infinite, I can see the value in infinite for app editor since the user can save, refresh, and pick up the changes without losing any work. For app launch though, it might be nice to pick up the changes automatically without losing any launch configuration.  
If we feel confident it'll be rare for reference genomes to be updated though, then I don't see an issue with using infinite. We could always change it later too.

ash3rz (approved):
LGTM :shipit:

src/components/apps/queries.js
psarando: Ah, yes, I saw at https://react-query-v2.tanstack.com/docs/guides/caching that it uses in-memory caching, but it doesn't say if that's the browser's memory or the server's 😄 

I could have sworn during testing something else, that I was seeing requests in the browser's dev console every time something was getting fetched from the server, but only once in the server's logs in the terminal, for the initial fetch in the browser. So I thought it was cached in the dev server rather than the browser; but now I can't seem to reproduce it with this component and API call 🤔 

In that case, I'll just use infinite cache and stale time for this PR, and the user should be able to refresh the browser in case of a rare update. And yes, we can always change it later if needed 👍

psarando: > I can see the value in infinite for app editor since the user can save, refresh, and pick up the changes without losing any work. For app launch though, it might be nice to pick up the changes automatically without losing any launch configuration.

I'm not sure I understand the concern, but just in case this helps: the reference genome form field won't lose its selected value, even while the list is loading and it displays as a loading skeleton. That skeleton doesn't replace the entire field, only the cosmetic display of the field. In fact, the browser console will show a warning that the selected value is not "in range" of the values of the list while it's loading (since it has no options yet until the loading is done). So once the list is fully loaded, the field will then display with the correct option selected.

psarando:
Thanks for the reviews! 👍 

I think I've addressed all comments, so I'll go ahead and merge this now.

PR #346, opened 2021-05-05
CORE-843 Users can list all or my communities

ash3rz:
I just picked a placeholder icon for now. I went with a list icon since a community is mostly a curated list of things (currently apps). If we decide to rename communities, then that may help with choosing an icon as well.

Being able to view and edit communities will be in the next PR.

Desktop:
image

Help Msg:
image

Mobile:
image


public/static/locales/en/communities.json
```diff
@@ -0,0 +1,9 @@
+{
+    "allCommunities": "All Communities",
+    "description": "Description",
+    "helpText": "Communities are public groups of apps. Creating a community allows you to create a custom collection of apps, and other CyVerse members can follow that community. When following a community, that community's app listing becomes available in the Apps view.",
```
sri2k1us, line 4: Might want to rethink this wording a little bit. `Communities are public groups of apps` - Right now communities works only with `Apps`. But it might be expanded to data. Also `When following a community, that community's app listing becomes available in the Apps view` - makes me think that I have to follow a community to access a particular app.

src/components/communities/Listing.js
```diff
@@ -0,0 +1,123 @@
...
+            <Table>
+                <TableBody>
+                    <EmptyTable
+                        message={t("noCommunities")}
+                        numColumns={Object.keys(COMMUNITY_COLUMNS).length}
+                    />
+                </TableBody>
+            </Table>
+        );
+    }
```
sri2k1us, line 118: `BasicTable` can handle  `loading ` and `empty table` for you. Just pass `loading` and `emptyDataMessage` props to `BasicTable` below.

src/components/communities/Toolbar.js
```diff
@@ -0,0 +1,126 @@
...
+                        : t("myCommunities")
+                }
+                renderInput={(params) => (
+                    <TextField
+                        {...params}
+                        label={t("view")}
+                        variant="outlined"
+                    />
+                )}
+            />
```
sri2k1us, line 71: Minor - we usually do these simple selections using `https://material-ui.com/components/selects/`. Any specific reason to use `AutoComplete` ?

sri2k1us (approved):
LGTM. I had some minor comments.

src/components/communities/Toolbar.js
psarando: For simple selects, I prefer a `TextField` with `select` and `variant="outlined"` props: https://material-ui.com/components/text-fields/#select

I can understand, though, that some users would prefer to use the mouse as little as possible, so tabbing into this field and typing the filter they want may be preferable 🤷

src/components/communities/Toolbar.js
```diff
@@ -0,0 +1,126 @@
...
+                }}
+            >
+                <Typography component="div">
+                    <Trans
+                        t={t}
+                        i18nKey="helpText"
+                        components={{
+                            p: <p />,
+                        }}
+                    />
```
psarando, line 119: I don't see any `p` tags in the `helpText`:
```suggestion
                    {t("helpText")}
```

src/components/communities/Listing.js
```diff
@@ -0,0 +1,123 @@
...
+                setData(results.groups);
+            },
+        },
+    });
+
+    const {
+        isFetching: fetchAllCommunities,
+        error: allCommunitiesError,
+    } = useQuery({
+        queryKey: [ALL_COMMUNITIES_QUERY, userProfile?.id],
```
psarando, line 77: Is `userProfile?.id` required for this `queryKey`?
If so, then should it also be checked as part of the `enabled` flag?

src/components/communities/Listing.js
```diff
@@ -0,0 +1,123 @@
...
+    }, [COMMUNITY_COLUMNS]);
+
+    const {
+        isFetching: fetchMyCommunities,
+        error: myCommunitiesError,
+    } = useQuery({
+        queryKey: [MY_COMMUNITIES_QUERY, { userId: userProfile?.id }],
+        queryFn: getMyCommunities,
+        config: {
+            enabled: COMMUNITY_FILTER.MY_COMMUNITIES === filter,
```
psarando, line 66: Should `userProfile?.id` be checked as part of the `enabled` flag?

public/static/locales/en/communities.json
psarando: If we want a more open description for the future, maybe something like:
```suggestion
    "helpText": "Communities can feature a curated collection of public resources in the Discovery Environment, such as apps. Creating a community allows you to create a custom collection of resources, and other CyVerse members can follow that community. When following a community, that community's custom listing of resources becomes easily accessible, such as its own custom listing in the Apps view.",
```
We probably want this text reviewed by more non-devs 🤔

psarando (approved):
LGTM 👍 

I just had a couple of minor suggestions, comments, and questions.

ash3rz:
Thanks for the reviews!  Updated the text based on 411's suggestions.

PR #347, opened 2021-05-05
Cancel consumer when websocket is closed.

sri2k1us:
SSIA.
Notes on how to test:
* Login to RabbitMQ web console and go to your queue in the `Queues` tab. Your queue name will look like `de_notifications.sriram`. You can easily find you queue by filtering with your cyverse username. Click on the queue name.

image

* Now login to DE.  Copy the consumer tag information printed on the server console.
* Use browser find to find that consumer tag in your queue's list of consumers.
image

* Now close the DE browser tab and refresh your RabbitMQ web console browser tab. The consumer tag should have disappeared from your queue.

### DIFF
--- a/src/server/amqp.js
+++ b/src/server/amqp.js
@@ -50,6 +50,7 @@ export function setUpAmqpForNotifications() {
  * @param {Object} ws - Websocket instance
  */
 export function getNotifications(user, ws) {
+    let consumerTag = "";
     if (!user) {
         logger.error("User not found. Unable to get notifications!");
         ws.close();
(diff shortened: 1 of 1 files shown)


src/server/amqp.js
```diff
@@ -63,11 +64,23 @@ export function getNotifications(user, ws) {
...
     });
 
     ws.on("error", function (error) {
         logger.info("Websocket error: " + error);
+        msgChannel.unbindQueue(
+            QUEUE,
+            config.amqpExchangeName,
+            NOTIFICATION_ROUTING_KEY + user
+        );
+        msgChannel.cancel(consumerTag);
```
psarando, line 83: Minor suggestion, but since this logic is repeated in `ws.on("close",...)`, I would have put them into some kind of "connectionCleanup" function, and let both handlers call that.

psarando (approved):
LGTM 👍

ash3rz (approved):
LGTM :shipit:

src/server/amqp.js
```diff
@@ -63,11 +64,23 @@ export function getNotifications(user, ws) {
...
         ws.close();
```
ianmcorvidae: Since we're calling close here, do we need to duplicate the unbind/cancel, or would it get handled in the close handler, above? I suppose it may not hurt to have both, as long as channel.cancel won't error out if the consumer is already canceled, though.

sri2k1us:
Thanks for the reviews.

ianmcorvidae (commented):
one minor comment, shouldn't matter much since it's already merged, but anyway

src/server/amqp.js
psarando: Ah, good catch. I didn't notice that in my review, but it looks like it's now handled in the refactor in the last commit before this PR was merged (close and error are now calling a common cleanup function).

sri2k1us: Should I explicitly call `close` on websocket when there is an websocket error? I had that previously but removed it during refactor 🤔

PR #349, opened 2021-05-07
Core 1166 Allow users to copy metadata

sri2k1us:
- Copy Dialog
image

- Validation
image


src/components/data/SelectionDrawer.js
```diff
@@ -78,13 +78,11 @@ function SelectionToolbar(props) {
     // The path key includes the file name if it's a file
     const handleConfirm = () => {
         if (multiSelect) {
-            const resourceNames = selectedResources.map(
-                (resource) => resource.path
-            );
-            onConfirm(resourceNames);
+            onConfirm(selectedResources);
```
sri2k1us, line 81: I am returning array of disk resource object only in case of multi-select to keep the changes to minimum.

psarando (changes requested):
Looks good so far. In addition to a few minor suggestions, I think there's still a `DataSelectionDrawer` in `src/components/data/viewers/PathListViewer.js` that needs its `onConfirm` refactored.

Also there are a couple of `multiSelect={true}` fields in `stories/data/SelectionDrawer.stories.js` that may be broken now (not sure if you tested those yet)?

src/components/metadata/form/index.js
```diff
@@ -489,6 +517,13 @@ const MetadataForm = ({
                             setTemplateMetadata(metadata);
                             setTemplateListingDialogOpen(true);
                         }}
+                        onCopyMetadataSelected={() => {
+                            if (userProfile) {
+                                setCopyDialogOpen(true);
+                            } else {
+                                setCopyDialogOpen(true);
+                            }
```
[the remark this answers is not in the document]
sri2k1us: copy 🍝  issue.

sri2k1us:
> Looks good so far. In addition to a few minor suggestions, I think there's still a `DataSelectionDrawer` in `src/components/data/viewers/PathListViewer.js` that needs its `onConfirm` refactored.
> 
> Also there are a couple of `multiSelect={true}` fields in `stories/data/SelectionDrawer.stories.js` that may be broken now (not sure if you tested those yet)?

I knew I am going to miss something. Good Catch 👍🏽 .

src/components/metadata/form/index.js
```diff
@@ -160,7 +164,9 @@ const MetadataFormListing = (props) => {
 
     const saveDisabled = loadingOrSubmitting || !dirty || errors.error;
 
-    const showSaveToFile = !loadingOrSubmitting && !(dirty && editable);
+    const showSaveToFile = !loadingOrSubmitting && !dirty;
```
ash3rz, line 167: Looks like `editable` got dropped from this boolean - was that intentional?

ash3rz (approved):
LGTM - I just had one minor question 👍

src/components/metadata/form/index.js
sri2k1us: Yes. So that users can save the metadata even if they only have read-only access to the files (eg: community data)

psarando (approved):
LGTM :shipit: 👍 🎉

sri2k1us:
Thanks for the reviews.

PR #353, opened 2021-05-11
CORE-1409 Add Refresh btn to notifications listing

ash3rz:
Desktop:
image

Mobile - not sure if these should get pushed into a dot menu at this point, but I left it for now:
image


psarando (approved):
LGTM :shipit: 👍

ash3rz:
Thanks for the review!  Additional reviews still welcome!

PR #354, opened 2021-05-12
CORE-1408 Add last modified and size to default columns

ash3rz:
From the discussion, it sounded like we're not concerned about mobile view having horizontal scroll with the extra columns.  If that's not the case, I can try to make it so the default for mobile is just the checkbox, name, and dot menu like before.

NOTE: This will not affect anyone who has already customized their columns.  Local storage gets priority.

Desktop:
image

Mobile:
image


psarando (commented):
:shipit:

ash3rz:
Thanks for the review!  Additional reviews still welcome!

PR #357, opened 2021-05-13
CORE-1420 Fixes to prevent editing Workflows as regular Apps

psarando:
This PR will prevent editing Workflows as regular Apps by displaying the `ComingSoonInfo` on the apps edit page for Workflows, and it also disables the edit menu item in the apps listing row dot-menu for public workflows (since that's the current behavior in old DE).

This PR also includes fixes for menu item keys in the apps listing row dot-menu.

Workflow edit page - coming soon!


ash3rz (approved):
:shipit:

psarando:
Thanks for the review!

PR #359, opened 2021-05-14
CORE-1422 Add a `/de` page for redirecting legacy links

psarando:
This PR will add a `/de` page for redirecting legacy links, such as the following:

* Quick Launch links
  * https://de.cyverse.org/de/?type=quick-launch&quick-launch-id=ff6719c2-90ef-43d8-98bb-7608c9c0a32b&app-id=67d15627-22c5-42bd-8daf-9af5deecceab
* App links
  * https://de.cyverse.org/de/?type=apps&app-id=67d15627-22c5-42bd-8daf-9af5deecceab&system-id=de
* Data links
  * https://de.cyverse.org/de/?type=data&folder=/iplant/home/shared/iplantcollaborative/example_data/wordcount

This page will also redirect to the root URL https://de.cyverse.org/ by default (so it will also handle redirecting https://de.cyverse.org/de to the Dashboard).

An empty, loading table will display while the redirect is handled.
image

### DIFF
--- a/src/pages/de.js
+++ b/src/pages/de.js
@@ -8,7 +8,11 @@ import React from "react";
 
 import { useRouter } from "next/router";
 
-import { getAppLaunchPath, getQuickLaunchPath } from "components/apps/utils";
+import {
+    getAppLaunchPath,
+    getListingPath,
+    getQuickLaunchPath,
+} from "components/apps/utils";
 import { getFolderPage } from "components/data/utils";
 import systemId from "components/models/systemId";
 import BasicTable from "components/utils/BasicTable";
@@ -20,10 +24,25 @@ export default function DeLegacyRedirector() {
     let redirectPath;
     switch (query?.type) {
         case "apps":
-            redirectPath = getAppLaunchPath(
-                query["system-id"],
-                query["app-id"]
-            );
+            if (query["app-category"]) {
+                redirectPath = getListingPath(
+                    null,
+                    null,
+                    null,
+                    null,
+                    null,
+                    JSON.stringify({
+                        system_id: query["system-id"] || systemId.de,
+                        id: query["app-category"],
+                    })
+                );
+            } else {
+                redirectPath = getAppLaunchPath(
+                    query["system-id"],
+                    query["app-id"]
+                );
+            }
+
             break;
 
         case "data":
--- a/src/pages/de.js
+++ b/src/pages/de.js
@@ -1,0 +1,60 @@
+/**
+ * A backwards compatibility page for redirecting legacy links
+ * to their proper current locations.
+ *
+ * @author psarando
+ */
+import React from "react";
+
+import { useRouter } from "next/router";
+
+import { getAppLaunchPath, getQuickLaunchPath } from "components/apps/utils";
+import { getFolderPage } from "components/data/utils";
+import systemId from "components/models/systemId";
+import BasicTable from "components/utils/BasicTable";
+
+export default function DeLegacyRedirector() {
+    const router = useRouter();
+    const query = router.query;
+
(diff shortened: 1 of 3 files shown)


src/pages/de.js
```diff
@@ -0,0 +1,60 @@
...
+
+        case "quick-launch":
+            redirectPath = getQuickLaunchPath(
+                query["system-id"] || systemId.de,
+                query["app-id"],
+                query["quick-launch-id"]
+            );
+            break;
+
+        default:
```
ash3rz, line 41: Nice ✨ 

I did a quick check in the old DE and looks like we also had the case for an app category query string in the path.  Not sure how often that was used or if it's worth preserving at this point, but figured I'd mention it: https://github.com/cyverse-de/ui/blob/master/de-lib/src/main/java/org/iplantc/de/desktop/client/presenter/DesktopPresenterImpl.java#L812-L817

ash3rz (approved):
🎉 Nice find!  I would've tried to do something complicated in Express 😄 
LGTM, I just had a minor consideration that can probably be ignored

src/pages/de.js
psarando: Good catch. I forgot to double check the old code.

It looks like it's attempting to open the apps listing to the appropriate category, and it may have attempted to select the correct app with the given app ID as well, though it's not doing that anymore.

The new code can display the apps listing for a specific category with a query param like `selectedCategory={"system_id":"de","id":"..."}`, except the category filter drop-down will display a blank label.

If that's acceptable for these legacy links, I can add that change to this PR as well.

sri2k1us: How is this different from https://de.cyverse.org/de/?type=apps&app-id=67d15627-22c5-42bd-8daf-9af5deecceab&system-id=de? This link also filters by `system-id` and `id` ? 🤔

sri2k1us (approved):
LGTM 🚢

src/pages/de.js
psarando: The link https://de.cyverse.org/de/?type=apps&app-id=67d15627-22c5-42bd-8daf-9af5deecceab&system-id=de
will redirect users to https://de.cyverse.org/apps/de/67d15627-22c5-42bd-8daf-9af5deecceab/launch (since the old DE is opening the launch dialog for those links now), but for a link like https://de.cyverse.org/de/?type=apps&system-id=de&app-category=8835fcff-53d8-472a-8fc5-9ffc03ebe925, I was going to add a redirect to the apps listing page, such as: `https://de.cyverse.org/apps?selectedCategory={"system_id":"de","id":"8835fcff-53d8-472a-8fc5-9ffc03ebe925"}`.

That will open the apps listing for the given category, and it will also ignore any `app-id` query param (which is what the old DE is also doing now).

sri2k1us: Ahh now I see it. Somehow, I got confused with `id` for selected category with app id. 👍🏽 .

src/pages/de.js
```diff
@@ -0,0 +1,79 @@
...
+                    null,
+                    null,
+                    null,
+                    null,
+                    null,
+                    JSON.stringify({
+                        system_id: query["system-id"] || systemId.de,
+                        id: query["app-category"],
+                    })
+                );
```
psarando, line 38: Even though this returns a `{pathname, query}` object, it turns out `router.replace` won't auto-stringify the `selectedCategory` query param's value 😞

psarando:
Thanks for the reviews! 👍 

I'll merge this now for tomorrow's hotfix, but more reviews still welcome on the category listing redirect (not sure how often those links are used anyway 🤷 ).

PR #360, opened 2021-05-17
CORE-1411 Display analysis details input params as links

psarando:
This PR updates the analysis details `ParamsPanel` to display input params as links that will navigate to the target folder, or parent folder for file params. It will also display a "not found" error message below the input's display name if the input no longer exists at that location.

The links are display using a `DELink`, which includes a tooltip of the link's target, which is the input's folder location.

Input param not found
...
Input param link

### DIFF
--- a/src/components/analyses/details/ParamsPanel.js
+++ b/src/components/analyses/details/ParamsPanel.js
@@ -25,8 +25,7 @@ import AppParamTypes from "components/models/AppParamTypes";
 
 import DELink from "components/utils/DELink";
 import DETableHead from "components/utils/DETableHead";
-import DEErrorDialog from "components/utils/error/DEErrorDialog";
-import ErrorTypography from "components/utils/error/ErrorTypography";
+import ErrorTypographyWithDialog from "components/utils/error/ErrorTypographyWithDialog";
 import { ERROR_CODES, getErrorCode } from "components/utils/error/errorCode";
 import TableLoading from "components/utils/TableLoading";
 
@@ -51,10 +50,12 @@ function InputParameterValue(props) {
     const { id, param_type, path } = props;
 
     const [statError, setStatError] = React.useState();
+    const [otherError, setOtherError] = React.useState();
 
     React.useEffect(() => {
         if (!path) {
             setStatError(null);
+            setOtherError(null);
         }
     }, [path]);
 
@@ -72,14 +73,17 @@ function InputParameterValue(props) {
             enabled: path,
             onSuccess: () => {
                 setStatError(null);
+                setOtherError(null);
             },
             onError: (error) => {
                 if (ERROR_CODES.ERR_DOES_NOT_EXIST === getErrorCode(error)) {
                     setStatError(
                         t("errorInputDoesNotExist", { path: linkTarget })
                     );
+                    setOtherError(null);
                 } else {
                     setStatError(null);
+                    setOtherError(error);
                 }
             },
         },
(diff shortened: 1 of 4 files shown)


src/components/analyses/details/ParamsPanel.js
```diff
@@ -27,9 +47,72 @@ import {
...
+            },
+            onError: (error) => {
+                if (ERROR_CODES.ERR_DOES_NOT_EXIST === getErrorCode(error)) {
+                    setStatError(
+                        t("errorInputDoesNotExist", { path: linkTarget })
+                    );
+                } else {
+                    setStatError(null);
+                }
+            },
```
sri2k1us, line 84: Should we show error that happens when irods is or data-info is down ? 🤔

src/components/analyses/details/ParamsPanel.js
```diff
@@ -27,9 +47,72 @@ import {
...
+    React.useEffect(() => {
+        if (!path) {
+            setStatError(null);
+        }
+    }, [path]);
+
+    const displayValue = parseNameFromPath(path);
+    const linkTarget = getFolderPage(
+        AppParamTypes.FOLDER_INPUT === param_type ? path : getParentPath(path)
+    );
```
sri2k1us, line 64: Should we rather open the file in viewer cos the file could be buried down cos pagination although am not sure how useful it is to open a file if it is binary 🤔

sri2k1us (approved):
LGTM. I have couple of questions though. We should clear that up before merging.

src/components/analyses/details/ParamsPanel.js
psarando: Right, I'm not sure if users will want to view input contents from here.

I thought the main usefulness of this feature was the tooltip to show where the input was located, or an error message if the input no longer exists (to explain why it would be missing in a relaunch).

Maybe we can stick with the link taking the user to the folder listing for now, and if users start asking to view the contents from here instead, we can change it then. 🤔

src/components/analyses/details/ParamsPanel.js
psarando: Not sure 🤔 
This was basically copy-paste from the app launch input selectors, which displayed "not found" errors in the launch form the same way (and this is the way they were displayed for input params in the old DE too). I guess it silently ignores anything except "not found" errors, and if there is still a problem when the app launches, the analysis will fail then:
https://github.com/cyverse-de/sonora/blob/0b03f7314c717e646339f1ac681bc8058aefdec5/src/components/apps/launch/params/DiskResourceSelector.js#L30-L54

In this display, I could add a different error message if the path couldn't be checked.
Then maybe the file could still display as a link so the user can check if the error is temporary by clicking on it?

src/components/analyses/details/ParamsPanel.js
psarando: I got sci-analyst feedback preferring navigating to the parent folder, so I'll stick with that for now.

src/components/analyses/details/ParamsPanel.js
psarando: It turned out to be pretty easy to also display other errors in an `ErrorTypographyWithDialog`, so I'll show that, but also keep the link in case the error is temporary (sci-analyst feedback also said that should be fine).

Analysis params input - other errors

This coud also be updated in the app launch form's default input params (also relaunch input params) in a future PR.

psarando:
I think I've addressed all feedback so far, so I'll go ahead and merge this now, but more reviews still welcome 🙇

src/components/analyses/details/ParamsPanel.js
psarando: Actually, this should never be an issue in the app launch form, since the service will remove inputs that no longer exist (or the endpoint will return an error for other stat errors). This could still be useful when checking default inputs in the app editor, though.

PR #362, opened 2021-05-18
Fix bugs around Global and Detailed search

sri2k1us:
- Fix intercom logging so that it logs only when search results are updated.
- Fix copying DE link from details page.
- Show search results from `Community Data` and `Shared with Me` in Detailed search results page.
- Fix `Copy DE Link` tool tip.

### DIFF
--- a/src/components/search/detailed/DataSearchResults.js
+++ b/src/components/search/detailed/DataSearchResults.js
@@ -39,6 +39,7 @@ import { getHost } from "components/utils/getHost";
 import { copyStringToClipboard } from "components/utils/copyStringToClipboard";
 import { copyLinkToClipboardHandler } from "components/utils/copyLinkToClipboardHandler";
 import CopyLinkButton from "components/utils/CopyLinkButton";
+import { trackIntercomEvent, IntercomEvents } from "common/intercom";
 
 import { IconButton, Tooltip, Typography, Grid } from "@material-ui/core";
 import { Info, Label } from "@material-ui/icons";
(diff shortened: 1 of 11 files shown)


src/components/search/detailed/DataSearchResults.js
```diff
@@ -149,21 +143,25 @@ function DataSearchResults(props) {
...
+                    userHomeDir: "",
                     rowsPerPage: searchConstants.DETAILED_SEARCH_PAGE_SIZE,
                     sortField: sortField,
                     sortDir: sortOrder,
                 },
             ]);
             setDataSearchQueryEnabled(true);
         }
-    }, [searchTerm, sortField, sortOrder, userHomeDir]);
+    }, [searchTerm, sortField, sortOrder]);
```
sri2k1us, line 154: Now I think about it - this still doesn't cover 1 use case where there are no search results matching the the term entered by users. In that case we don't show the `See all results for `term` option in the drop-down. So user has to hit `enter` to see all the results. May be we should that option even when there are no matching top 10 results for `Data` 🤔

psarando: Sounds like a good idea to me 👍

sri2k1us: Ok this done.

ash3rz (approved):
LGTM :shipit:

psarando (approved):
LGTM :shipit: 👍

sri2k1us:
Thanks for the reviews.

PR #367, opened 2021-06-01
CORE-1430: accommodate the new response body format for the analysis …

slr71:
…sharing and unsharing endpoints

### DIFF
--- a/src/components/analyses/listing/Listing.js
+++ b/src/components/analyses/listing/Listing.js
@@ -277,21 +277,11 @@ function Listing(props) {
         submitAnalysisSupportRequest,
         {
             onSuccess: (responses) => {
-                const shareError =
-                    responses?.length > 1 &&
-                    responses[0].sharing.find((share) =>
-                        share.analyses.find((analysis) => !analysis.success)
-                    );
-
-                if (shareError) {
-                    showErrorAnnouncer(t("statusHelpShareError"));
-                } else {
-                    setShareWithSupportAnalysis(null);
-                    announce({
-                        text: t("statusHelpShareSuccess"),
-                        variant: AnnouncerConstants.SUCCESS,
-                    });
-                }
+                setShareWithSupportAnalysis(null);
+                announce({
+                    text: t("statusHelpShareSuccess"),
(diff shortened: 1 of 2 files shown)


src/components/analyses/listing/Listing.js
```diff
@@ -277,21 +277,11 @@ function Listing(props) {
...
-
-                if (shareError) {
-                    showErrorAnnouncer(t("statusHelpShareError"));
-                } else {
-                    setShareWithSupportAnalysis(null);
-                    announce({
-                        text: t("statusHelpShareSuccess"),
-                        variant: AnnouncerConstants.SUCCESS,
-                    });
-                }
```
slr71, line 294: In this case, the only response bodies that are being processed are the ones from the analysis sharing and unsharing endpoints. These endpoints have been updated so that if any analysis would fail validation, then the endpoint will return a 400 status, so there's no need to check the response body at all.

src/components/sharing/index.js
```diff
@@ -209,7 +209,7 @@ function Sharing(props) {
                     sharing.forEach((shares) => {
                         const updates = getShareResponseValues(shares);
                         const successFalse = updates.filter(
-                            (update) => update.success === false
```
slr71, line 212: In this case the response bodies from other sharing and unsharing endpoints are also being processed, so we still need to examine the response bodies. For the analysis sharing and unsharing endpoints, the `success` flag will not exist. (The analysis sharing and unsharing endpoints use an `ok` flag instead because they're now asynchronous, and it's impossible to predict with complete accuracy whether or not the analysis will be shared successfully.) This means that an analysis sharing or unsharing response will never match the response body scanning code here. This is okay, however, because the analysis sharing and unsharing endpoints will only return an error in the response body if a preliminary validation check fails. And if that happens then the endpoint will return a 400 status. The response bodies from other sharing and unsharing endpoints will still be processed as they were before.

sri2k1us (approved):
:shipit:

psarando (approved):
LGTM 👍

slr71:
Thanks for the reviews!

PR #369, opened 2021-06-03
Fix IDs not being set on next/link

ash3rz:
Nextjs's Link doesn't accept an ID attribute, so I refactored the options to always have MUI's Link, which does accept an ID.


sri2k1us (approved):
:shipit:

ash3rz:
Thanks for the review! Since this is kinda small and I'm iterating on tests, I'll merge it now, but additional reviews still welcome!

PR #370, opened 2021-06-03
CORE-1432 Fixes for Add/Edit Tool form

psarando:
CORE-1432 was initially for fixing a submission error where the form was submitting an `id` in `container_port` settings, which are returned in the `GET` tool details endpoints, but are not allowed in the `PATCH` tool update endpoints.

During testing, I found many "uncontrolled to controlled" warnings in the dev console for form fields that didn't have initial values set; and also other bugs when editing tools, such as missing `interactive_apps` settings on new interactive tools, or submission errors when the user attempted to remove values from number fields causing the form to submit empty strings for those values (such as in the `UID` field). 

This PR will apply the following fixes for the Add/Edit Tool form:

* Ensure at least an empty string is set for all form field values.

* Refactored the `EditTool` form's `mapPropsToValues` function to only return fields used in the form or required in submissions.

* Added a `formatSubmission` function to remove empty values, where needed. This also fixes logic for adding `interactive_apps` settings, and allows the user to remove values from number fields.

* Fixed `ToolImplementation` validation on file fields.

* Fixed Storybook `configMocks` and `EditTool` stories.

* Fixed the request `body` format in the `adminAddTool` service facade.


src/components/tools/edit/EditTool.js
```diff
@@ -150,20 +152,8 @@ function EditToolDialog(props) {
...
-        if (submission.interactive) {
-            const interactive_apps = {
-                cas_url: config.vice.defaultCasUrl,
-                name: config.vice.defaultName,
-                image: config.vice.defaultImage,
-                cas_validate: config.vice.defaultCasValidate,
-            };
-            submission.container.interactive_apps = interactive_apps;
-            submission.container.skip_tmp_mount = true;
-        }
```
psarando, line 166: The `interactive` flag was never set in the form for new tools, so these settings were never added to the submission.

src/components/tools/edit/formatters.js
```diff
@@ -0,0 +1,325 @@
...
+        memory_limit: memory_limit || undefined,
+        min_disk_space: min_disk_space || undefined,
+        pids_limit: pids_limit || undefined,
+        skip_tmp_mount: false,
+        interactive_apps: undefined,
+        container_ports: undefined,
+        container_devices: undefined,
+        container_volumes: undefined,
+        container_volumes_from: undefined,
+    };
```
psarando, line 274: The `PATCH` endpoint docs have the following note:
> If the container object is omitted in the request, then existing container settings will not be modified, but if the container object is present in the request, then all container settings must be included in it. Any existing settings not included in the request's container object will be removed.

So I'm using `undefined` to remove container settings in submissions (something in the client will remove those keys at some point in the serialization and submission to the service).

This only applies to container settings, so for top-level keys I'm still allowing empty strings (or `0`) to allow those values to be removed, which appears to be acceptable to the service.

src/components/tools/edit/ToolImplementation.js
```diff
@@ -76,7 +76,7 @@ function TestFiles(props) {
                                         label={t("fileName")}
                                         required
                                         validate={(value) =>
-                                            nonEmptyField(values, i18nUtil)
+                                            nonEmptyField(value, i18nUtil)
```
sri2k1us, line 79: 👀

stories/configMock.js
```diff
@@ -1,3 +1,5 @@
+const ONE_GiB = 2 ** 30;
```
sri2k1us, line 1: Trivial - This is already defined in `src/constants.js`. Can we reuse that ?

sri2k1us (approved):
LGTM 🚢

ash3rz (approved):
LGTM :shipit:

stories/configMock.js
psarando: Good catch 👍

psarando:
Thanks for the reviews! 👍

PR #371, opened 2021-06-04
Move forwardRef to inner OptionLink

ash3rz:
Hopefully the last fix 😅 I didn't follow [this](https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-function-component) closely enough to realize I needed to move the `forwardRef` to the component that `Link` is actually wrapping.


sri2k1us (approved):
Tried it locally. Seems to work. :shipit:

ash3rz:
Thanks for the review! Additional reviews still welcome!

PR #373, opened 2021-06-09
CORE-707 Add Copy App/Workflow menu items to Apps Listing

psarando:
This PR will add Copy App/Workflow menu items to the Apps Listing by updating the `RowDotMenu` with a Copy action, labeled as "Copy App" for 1-step apps or "Copy Workflow" for multi-step apps.

The old DE had completely separate menus for Apps and Workflows, since it was requested that users had a way to tell the difference between them somehow; but I thought simply renaming these menu items in the new dot-menu could accomplish the same thing.

The Edit menu items were also renamed for Apps and Workflows.

Only 1-step apps are supported for now, which will also redirect the user to the copied app's edit page.

Selecting the "Copy Workflow" action will display a "Coming Soon!" announcer popup.

This PR also fixes the `mockAxios` setup in the apps `Editor` stories, since another story's mock setup (probably the Listing stories?) was interfering with these stories.

Edit/Copy App menu items:
Edit/Copy App menu items

---
Edit/Copy Workflow menu items:
Edit/Copy Workflow menu items

---
Note that when the user selects the "Copy App" menu item, the copy icon is replaced with a circular loading icon and the menu item becomes disabled while the service request is processed:
Copy App menu item loading


ash3rz (approved):
LGTM :shipit:

psarando:
Thanks for the review 👍

PR #375, opened 2021-06-19
CORE-1440 Prompt users to update their user portal account info

ash3rz:
During the grace period:
image

When the profile info is expired:
image


src/components/layout/UserPortalUpdatePrompts.js
```diff
@@ -0,0 +1,149 @@
...
+            const portalUrl = `${updateUrl}&redirectUrl=${encodeURIComponent(
+                window.location.href
+            )}`;
+            window.location.replace(portalUrl);
+        }
+    };
+
+    return (
+        <>
+            <Snackbar
```
ash3rz, line 69: I created my own Snackbar here instead of using the CyVerse announcer because it seemed different enough:
1) We didn't want it to be auto-dismissed after a short period of time - the current CyVerse Announcer has the ability to set a custom duration but by default it has a duration set if one is not provided. In order to display the announcer infinitely, you have to omit the duration prop which wouldn't be possible with the CyVerse announcer since it sets a default.
2) I noticed it was possible to dismiss regular announcers by just clicking around the DE, so it was possible to miss the announcer entirely if you were clicking around while the announcer was just popping up

sri2k1us:
LGTM. But am not too happy that we have to show 3 paragraph of text in an alert. Doesn't looks good on the 👀 and not a good UX. Am going to check with 411 to see if bulk of this text can be moved to main website page and show a Learn More link here. So the alert will be: 
`We noticed that you haven't updated your user profile in a while. Learn More. Update Now` with an alert `title`  `Update Account Information`

src/server/api/cyverseRealm.js
```diff
@@ -0,0 +1,51 @@
...
+            .request({
+                method,
+                url: apiURL.toString(),
+                headers: accessToken
+                    ? { Authorization: `Bearer ${accessToken}` }
+                    : {},
+            })
+            .then((apiResponse) => {
+                // Do not set apiResponse.headers
+                // Setting them wipes out the auth for subsequent calls
```
sri2k1us, line 39: 👍🏽

ash3rz:
Update pushed with the link to the new FAQ page, and screenshots updated.

public/static/locales/en/common.json
```diff
@@ -80,9 +81,13 @@
     "timestamp": "{{timestamp}} ago",
     "tools": "Tools",
     "update": "Update",
+    "updateAccountInformation": "Update Account Information",
+    "updateNow": "Update Now",
     "useOldDEForThisFeature": "Please use the <oldDELink>previous version of the Discovery Environment</oldDELink> for access to this feature, until we can add it to this version.",
     "userMenuAriaLabel": "user menu",
     "username": "Username",
+    "userPortalExpiredText": "Your account information is outdated. For continued access to CyVerse services, update your account information now.",
+    "userPortalGracePeriodText": "Update Account Information<br/>We’ve noticed that you haven’t updated your profile in a while.",
```
sri2k1us, line 90: `Update Account Information` needs to be in `<AlertTitle>` tag. So that the heading and message is has different weights.

sri2k1us (approved):
LGTM. I had 1 more comment about `Alert Title`. Once that is taken care, :shipit:

public/static/locales/en/common.json
ash3rz: Ah, I didn't notice it was a separate component.  I was looking for a `title` prop in the `Alert` component 👍

src/components/layout/UserPortalUpdatePrompts.js
```diff
@@ -0,0 +1,165 @@
...
+                    {t("userPortalGracePeriodText")}
+                </Alert>
+            </Snackbar>
+
+            <Dialog
+                open={!!userPortalResp?.update_required}
+                id={baseId}
+                fullWidth
+                aria-labelledby={dialogTitleId}
+            >
```
psarando, line 141: Should this Dialog set the `disableBackdropClick` and `disableEscapeKeyDown` props to prevent the user from closing the dialog with those actions?
```suggestion
            <Dialog
                open={!!userPortalResp?.update_required}
                id={baseId}
                fullWidth
                disableBackdropClick
                disableEscapeKeyDown
                aria-labelledby={dialogTitleId}
            >
```

psarando (approved):
LGTM 👍 

Though I wonder if we need to completely prevent the user from closing the `update_required` Dialog.

src/components/layout/UserPortalUpdatePrompts.js
ash3rz: Ah good catch.  I tried this locally and, at least in FireFox, was unable to close the dialog by clicking outside or hitting Escape, but I'll add these to be safe.

psarando: Oh, maybe those only matter if `onClose` is also set, but it's not set here.
Oh well, doesn't hurt to add them anyway 🤷

ash3rz:
Thanks for the reviews!

src/components/layout/UserPortalUpdatePrompts.js
```diff
@@ -0,0 +1,167 @@
...
+
+    const onCloseAnnouncer = () => setShowSnackbar(false);
+
+    const onUpdateNow = () => {
+        const updateUrl = userPortalResp?.update_url;
+        if (updateUrl) {
+            const portalUrl = `${updateUrl}&redirectUrl=${encodeURIComponent(
+                window.location.href
+            )}`;
+            window.location.replace(portalUrl);
```
psarando, line 70: I didn't catch this until I tried it after this PR was merged, but a minor inconvenience of using `replace` here is that it breaks the user's browser history, since the user portal simply redirects back and does not also use replace.

So which ever page the user was viewing when they get the dialog will be replaced with the user portal URL after they are redirected back to the DE.

ash3rz: Almost forgot about this comment! 😅 
I don't think I'm seeing what you're seeing though, or maybe I'm misunderstanding.  When I tested it, I:
* Navigated to de.cyverse.org
* Got the pop-up to update
* Got sent to the user portal, updated my info, hit "Proceed to application"
* Got redirected back to the DE with de.cyverse.org as my url
* Back button takes me to user portal, back button again takes me to the DE

PR #376, opened 2021-06-22
Instant Launches should set the output directory 

johnworth:
The quick launch that is used to form the instant launch can have a output directory set that the launching user can't access. The output directory should be set to the user's default output directory during the instant launch process. Additionally work needs to be done to support the `Go to output folder` option in the analyses view.

### DIFF
--- a/src/components/instantlaunches/index.js
+++ b/src/components/instantlaunches/index.js
@@ -21,11 +21,10 @@ import { useTranslation } from "i18n";
 import { build as buildID } from "@cyverse-de/ui-lib";
 import ids from "components/instantlaunches/ids";
 import { instantlyLaunch } from "serviceFacades/instantlaunches";
-import { useBootstrapInfo } from "contexts/bootstrap";
 
 import { useMutation } from "react-query";
 import { getHost } from "components/utils/getHost";
-import { useHomePath } from "components/data/utils";
+import { useDefaultOutputDir } from "components/data/utils";
 
 const useStyles = makeStyles((theme) => ({
     progress: {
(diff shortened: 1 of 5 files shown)


src/components/instantlaunches/index.js
```diff
@@ -107,6 +110,16 @@ const InstantLaunchButton = ({
...
+    const [bootstrapInfo] = useBootstrapInfo();
+    const homePath = useHomePath();
+
+    const preferences = bootstrapInfo?.preferences;
+
+    const output_dir =
+        preferences?.default_output_folder?.path ||
+        preferences?.system_default_output_dir?.path ||
+        (homePath && `${homePath}/analyses`) ||
+        "";
```
psarando, line 122: This looks like a duplicate of the logic already in the apps launch form:
https://github.com/cyverse-de/sonora/blob/f067afa56fc0f435124955734e96637ca516d2a0/src/components/apps/launch/index.js#L74-L80

I would refactor this logic into a new hook, maybe named something like `useDefaultOutputDir`in `src/components/data/utils.js`, similar to `useSelectorDefaultFolderPath`:
https://github.com/cyverse-de/sonora/blob/f067afa56fc0f435124955734e96637ca516d2a0/src/components/data/utils.js#L167-L176

Then both this file and `src/components/apps/launch/index.js` would no longer have to reference `useBootstrapInfo` or `useHomePath`, since they are only using them to get this default output folder path.

psarando (approved):
LGTM 👍 
Though I would cleanup the DRY violation.

ash3rz (approved):
I agree with Paul's suggestion, otherwise LGTM :shipit:

src/components/instantlaunches/index.js
johnworth: Thanks for the heads up on this, I completely forgot that I had copied that code into here. That's what I get for letting a PR sit for a month...

johnworth: Okay, should be updated.

psarando (approved):
LGTM 🎉 ✨

PR #378, opened 2021-06-23
core-1456 Update sidebar menu color.

sri2k1us:
image

image

image

### DIFF
--- a/src/components/layout/UserMenu.js
+++ b/src/components/layout/UserMenu.js
@@ -30,12 +30,12 @@ const useStyles = makeStyles((theme) => ({
         padding: theme.spacing(1),
         [theme.breakpoints.down("xs")]: {
             padding: theme.spacing(0.2),
-            backgroundColor: theme.palette.info.main,
+            backgroundColor: "#052537",
         },
     },
 
     typography: {
-        color: theme.palette.info.main,
+        color: "#052537",
         [theme.breakpoints.down("xs")]: {
             color: theme.palette.info.contrastText,
         },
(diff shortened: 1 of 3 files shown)


psarando (changes requested):
LGTM 👍  but is there a reason for this change?
Is this the standard CyVerse colors now for app side-bars and toolbars?

These magic numbers are now repeated in a few locations, but they probably should come from named constants somewhere, or from fields in our theme (e.g. `theme.palette.main-side-bar` and `theme.palette.main-app-bar` perhaps):
https://github.com/cyverse-de/sonora/blob/3071bf50a9a46b3dab7d383cc47a2150c1caba34/src/components/theme/default/index.js#L1-L5

sri2k1us:
> LGTM 👍 but is there a reason for this change?
> Is this the standard CyVerse colors now for app side-bars and toolbars?
> 
> These magic numbers are now repeated in a few locations, but they probably should come from named constants somewhere, or from fields in our theme (e.g. `theme.palette.main-side-bar` and `theme.palette.main-app-bar` perhaps):
> https://github.com/cyverse-de/sonora/blob/3071bf50a9a46b3dab7d383cc47a2150c1caba34/src/components/theme/default/index.js#L1-L5

These changes were requested by Nirav for 2 reasons:
- Better contrast against highlight color
- Match AppBar color to User Portal color

But I don't think we will need to update the `primary` color. 

Adding to themes sounds like a good idea especially if we want to do a dark mode.

ash3rz (approved):
LGTM :shipit:

psarando (approved):
LGTM :shipit: 👍

sri2k1us:
Thanks for the review.

PR #380, opened 2021-06-26
CORE-726 Workflow Editor

psarando:
This PR adds the Workflow Editor, including integration with the existing apps edit page, and a new pipelines creation page.

The Apps Listing toolbar `+ Create` button is now a menu-button with 2 options, for creating an app or a workflow:
Create menu

---

The Workflow Editor looks basically the same as the App Editor (and Launcher), with 3 steps in the editor, but the App Stepper has been improved so that help text can also be included under each step's label:
Workflow Editor - Info step

Step 1 on Mobile:
Workflow Editor - Info step - mobile

---

After the user visits a step and has completed all fields in it, the step can now be marked as completed in larger screens:
Workflow Editor - Arrange step (Info step complete)

---

The step will only be marked completed after the user visits the step, which means either viewing another step, or after saving the form (with or without API errors):
Workflow Editor - IO Mapping step

Workflow Editor - All steps complete on save

---

Step info text can also now be replaced with an error message if the step has errors, but again the step has to have been visited before errors will display:
Workflow Editor - Step errors

Step error messages on mobile:
Workflow Editor - Step errors - mobile

---

If the user adds an app that uses a deprecated tool (such as anything using the `backwards-compat` image), then a warning will be displayed above the stepper (so it's displayed for all steps), and also displayed over the description of the app using the deprecated tool:
Workflow Editor - Arrange step - deprecated tool

---

### Known issues and TODO

1. So far only this Workflow Editor is using the new App Stepper features of marking steps completed and step help/error text, but the App Launch and Edit forms can use these features in a new PR.

2. The app search endpoint can only return high-level info about an app, so once the user selects an app from the App Search Drawer, the Workflow Editor will add a place holder for that app in "Select & Order" listing, but it will display a loading mask where the app description is normally displayed, and the move/delete buttons will be hidden until the editor can fetch task info for the app, which includes info about its inputs and outputs.

Workflow Editor - Arrange step - tasks loading

If the user selects Step 3 while the app's tasks are loading, then Step 3 will display empty fields for that app's inputs and outputs, even after the tasks are loaded from the service -- until the user views another step and comes back to step 3.
Workflow Editor - IO Mapping step - tasks loading

The entire Stepper should probably be masked while app tasks are loading, similar to how it's masked during a save. This would prevent the user from viewing another step until the tasks are loaded and input/output info can be displayed in step 3.


src/common/NavigationConstants.js
```diff
@@ -16,7 +16,8 @@ export default {
     LOGIN: "login",
     LOGOUT: "logout",
     MORE: "more",
-    NEW_APP: "apps/create",
+    NEW_APP: "/apps/create",
+    NEW_WORKFLOW: "/pipelines/create",
```
psarando, line 20: This allows the user to navigate to the create page from an app-by-id listing page, such as https://de.cyverse.org/apps/de/67d15627-22c5-42bd-8daf-9af5deecceab.

src/components/search/detailed/AppSearchDrawer.js
```diff
@@ -137,10 +137,7 @@ function SearchToolbar(props) {
                 value={searchText}
                 onChange={handleInputChange}
                 className={classes.searchField}
-                InputProps={{
-                    startAdornment: <SearchIcon color="primary" />,
-                    disableUnderline: true,
-                }}
+                InputProps={{ startAdornment: <SearchIcon color="primary" /> }}
```
psarando, line 140: This was throwing a warning that the DOM doesn't recognize `disableUnderline`, and removing it seemed to have no effect.

src/components/search/detailed/AppSearchResults.js
```diff
@@ -139,7 +139,7 @@ export default function AppSearchResults(props) {
                     )[1];
                     const { t } = useTranslation("common");
                     return (
-                        <Grid spacing={1}>
+                        <Grid container spacing={1}>
```
psarando, line 142: This fixed a warning that `spacing` can only be used on a `container` (which this is).

src/components/apps/AppStepper.js
```diff
@@ -128,22 +129,25 @@ const AppStepper = React.forwardRef((props, ref) => {
...
-                        </StepLabel>
-                    </StepButton>
-                </Step>
-            ))}
+            {steps.map((step, index) => {
+                const completed = stepCompleted && stepCompleted(index);
+                const hasError = stepError && stepError(index);
+
+                return (
+                    <Step key={step.label} completed={completed}>
```
psarando, line 137: Pretty simple change on the app `Step` side to get it to display as completed, though the state management in the form side is a little more complicated; such as managing step `touched` state, so it's not instantly marked as completed or with an error until the user actually does something with the step.

ash3rz (approved):
LGTM :shipit: 

I do agree that the steps should having a loading mask if still fetching the more detailed app info for tasks, especially since the UI shows a blue checkmark for the `Select and Order Apps` step (at least in that last screenshot example).  Another option (which might be harder to do) might be to change that checkmark to a loading icon and then display a message like "Changes since you last viewed" to draw the user back to that step, similar to how you're showing error messages on the stepper. Probably a loading mask should be sufficient though to avoid the need in the first place.

sri2k1us:
Code looks good to me however we need to be consistent with the naming. I see a lot of pipeline / workflow mixed up in naming the files / variables / messages. They all should be updated to use `workflow`.  (PS: I don't mind workflow being called Pipelines as well since they make more sense cos we only support linear workflows anyway but that change will have impact in other places so we need to get approval from sci-team and PIs.) 

UI/UX issues:

1. In the legacy DE, the 1st step's input field is disabled but I see an empty drop down in the new interface:
image. 
We should look into hiding that drop-down all together.

2. I think I had mentioned this comment in App Integration. The last step should include options to `Save and Exit` or `Test` the App or Workflow. The current UI leaves the users hanging in the last step. This could be a separate PR.

3. Minor - Is it possible hide the `Up` arrow button for 1st app and the `Down` arrow button for the last step ?

src/components/apps/workflows/Editor.js
```diff
@@ -0,0 +1,429 @@
...
+                            getFormError("description", touched, errors)
+                        );
+                    }
+
+                    return getFormError(
+                        `workflowSteps[${stepIndex}]`,
+                        touched,
+                        errors
+                    );
+                };
```
sri2k1us, line 278: 👍🏽

sri2k1us (commented):
I had some comments on the UX but the biggest thing that need to be addressed is the naming inconsistency between workflow and pipeline

src/common/NavigationConstants.js
psarando: I was only mirroring the API endpoints, which use `/apps/pipelines`, but I'm fine with sticking with `/workflows` in the UI.

public/static/locales/en/workflows.json
```diff
@@ -0,0 +1,23 @@
...
+    "editWorkflow": "Edit Workflow: {{name}}",
+    "mapOutputsToInputs": "Match Outputs to Inputs",
+    "outputMappingLabel": "Step {{step}}: {{label}}",
+    "outputsLabel": "Output(s)",
+    "selectAndOrderApps": "Select & Order Apps",
+    "stepErrorInfo": "Workflows require a name and a description.",
+    "stepErrorMapping": "At least one input for each app must come from an output of a previous app.",
+    "stepErrorSelectAndOrder": "You must add 2 or more apps to this workflow.",
+    "stepHelpInfo": "Currently, only linear workflows may be created. At least one input for each app must come from an output of a previous app.",
+    "stepHelpMapping": "At least one input for each app must come from an output of a previous app.",
```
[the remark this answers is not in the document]
psarando: Yep. This was the help message in the original DE, and I couldn't think of any different wording for the error message 🤷

psarando:
> 1. In the legacy DE, the 1st step's input field is disabled but I see an empty drop down in the new interface:
>        We should look into hiding that drop-down all together.

That will be no problem 👍 
Workflow Editor - IO step - hidden inputs


> 3. Minor - Is it possible hide the `Up` arrow button for 1st app step and the `Down` arrow button for the last app step ?

That will be possible:
Workflow Editor - Arrange step - hidden move btns

psarando:
> Code looks good to me however we need to be consistent with the naming. I see a lot of pipeline / workflow mixed up in naming the files / variables / messages. They all should be updated to use `workflow`.

I've updated as much as I thought was reasonable to use `workflow` instead of `pipeline`, especially in any user-facing messages (and the `create` URL/path), but I left `pipeline` references in the files and functions that directly refer to the `/apps/pipelines` endpoints.

I can still rename the file and functions in `serviceFacades/pipelines` to use `workflows` instead, if anyone has a strong preference for it (it is a facade after all 🤔 ).

sri2k1us (approved):
LGTM 🎉

psarando:
Thanks for the reviews! 👍 

I think I've addressed all feedback, and I also found a solution to the Known-Issue-2 by masking the stepper and step nav buttons with loading skeletons while step tasks load.

I think a final `Save/Exit/Test` step can be done in another PR for both workflows and regular apps.

I'll merge this now, but additional reviews will still be welcome 🙇

PR #381, opened 2021-06-29
Core 1444 - Allow users to refresh cache of their publicly linked files

sri2k1us:
image

image


public/static/locales/en/data.json
```diff
@@ -122,6 +122,10 @@
     "requestDOIPrompt": "Please read the <manual>DOI Manual</manual> and check if you meet the requirements for submitting a DOI request.",
     "requestDOIAgreement": "By clicking <b>I need a DOI</b>, I have read and agree to the terms of the <agreement> Data Commons </agreement>.",
     "refresh": "Refresh",
+    "refreshCache": "Refresh Cache",
+    "refreshCacheTooltip": "Public file(s) are cached for performance. Select `Refresh Cache` to update the cache.",
```
psarando, line 126:
```suggestion
    "refreshCacheTooltip": "Public files are cached for performance. Select `Refresh Cache` to update the cache.",
```

public/static/locales/en/data.json
```diff
@@ -18,7 +18,7 @@
     "cancel": "Cancel",
     "communityData": "Community Data",
     "copyPath": "Copy Path",
-    "copyLinks": "Copy Public Link(s)",
+    "copyLinks": "Copy",
```
psarando, line 21: I'd be tempted to leave this alone, but change the button label to use `common:copy` instead 🤔

src/components/data/PublicLinks.js
```diff
@@ -186,47 +202,80 @@ function PublicLinks(props) {
...
-                            className={classes.toolbarButtons}
-                            onClick={() => setDownload(true)}
-                            startIcon={<CloudDownload />}
-                            id={build(baseId, ids.DOWNLOAD_LINKS_BTN)}
-                        >
-                            <Hidden xsDown>{t("download")}</Hidden>
-                        </Button>
-                    </Tooltip>
-                </Toolbar>
+                <TextareaAutosize
```
psarando, line 205: 👍

src/components/data/PublicLinks.js
```diff
@@ -186,47 +202,80 @@ function PublicLinks(props) {
...
+                    rowsMin={paths?.length + 1 || 3}
```
psarando, line 206: Seems like adding logic based on the contents to the `rowsMin` prop in `TextareaAutosize` defeats its purpose?
```suggestion
                    rowsMin={3}
```

src/server/configuration.js
```diff
@@ -284,3 +285,10 @@ export const uatURL = config.get("services.unified_astronomy_thesaurus.base");
  * The base URL for the User Portal's API
  */
 export const userPortalURL = config.get("services.user_portal_api.base");
+
+/**
+ *
+ * The base ULR for Data Store anon files
```
psarando, line 291:
```suggestion
 * The base URL for Data Store anon files
```

src/serviceFacades/filesystem.js
```diff
@@ -272,6 +272,25 @@ export const getPublicLinks = (key, paths) => {
...
+                    callApi({
+                        endpoint: `/api/filesystem/refresh-cache?path=${path}`,
+                        method: "GET",
+                    })
+                )
+            )
+                .then((values) => resolve(values))
+                .catch((error) => reject(error))
+        )
+    );
```
psarando, line 291: Pretty sure `Promise.all` doesn't need to be wrapped in another Promise.
```suggestion
            Promise.all(
                paths.map((path) =>
                    callApi({
                        endpoint: `/api/filesystem/refresh-cache?path=${path}`,
                        method: "GET",
                    })
                )
            );
```

psarando (approved):
LGTM 👍 

I only had a couple of minor suggestions.

src/serviceFacades/filesystem.js
psarando: Ah, yep, found another example:
https://github.com/cyverse-de/sonora/blob/3071bf50a9a46b3dab7d383cc47a2150c1caba34/src/serviceFacades/tools.js#L153-L159

sri2k1us: Hmm. I was looking for example when I wrote this and found this way and then https://github.com/cyverse-de/sonora/blob/3071bf50a9a46b3dab7d383cc47a2150c1caba34/src/serviceFacades/analyses.js#L143 😕

ash3rz (approved):
LGTM :shipit:

sri2k1us:
Thanks for the reviews.

PR #382, opened 2021-07-13
Admin Tool Requests.

sri2k1us:
image

image


src/components/doi/TableView.js
```diff
@@ -109,7 +109,7 @@ export default function TableView(props) {
                 <EnhancedTableHead
                     baseId={baseId}
                     selectable={false}
-                    rowsInPage={listing?.analyses?.length || 0}
```
sri2k1us, line 112: Copy / Pasta Issue.

src/components/tools/NewToolRequestDialog.js
```diff
@@ -29,7 +29,7 @@ export default function NewToolRequestDialog(props) {
     const [requestError, setRequestError] = useState();
 
     const { t } = useTranslation("tools");
-    const { t: i18nUtil } = useTranslation("utils");
```
sri2k1us, line 32: This was causing `404`. There is no corresponding `utils.json` file. There is only `util.json`

src/components/tools/listing/TableView.js
```diff
@@ -260,40 +261,42 @@ function TableView(props) {
...
-                    <ToolListingTableBody
-                        columns={columns}
-                        handleClick={handleClick}
-                        t={t}
-                        multiSelect={multiSelect}
-                        selected={selected}
-                        tableId={tableId}
-                        tools={tools}
-                        isAdmin={isAdmin}
+        <PageWrapper appBarHeight={isAdmin ? 280 : 0}>
```
sri2k1us, line 264: We need a better way of getting the correct scroll behavior without this size hard-coded.

src/pages/admin/apps.js
```diff
@@ -107,24 +82,18 @@ export default function Apps() {
...
         return (
             <>
-                <Tabs
-                    value={selectedTab}
-                    onChange={onTabSelectionChange}
-                    classes={{ indicator: classes.tabIndicator }}
-                >
-                    <Tab
+                <DETabs value={selectedTab} onChange={onTabSelectionChange}>
+                    <DETab
```
sri2k1us, line 86: Use `DETabs/DETab` instead of MUI Tabs / Tab.

src/serviceFacades/tools.js
```diff
@@ -158,6 +160,49 @@ function deleteTools({ ids: toolIds }) {
...
+        params["limit"] = rowsPerPage;
+        params["offset"] = page * rowsPerPage;
+    }
+
+    return callApi({
+        endpoint: `/api/admin/tool-requests`,
+        method: "GET",
+        params: params,
+    });
+}
```
sri2k1us: `GET` `/admin/tool-requests` supports pagination but right now it doesn't return the `total` field so it is not set up to use pagination for now.

src/components/tools/requests/Details.js
```diff
@@ -0,0 +1,184 @@
...
+
+    const { isFetching: isDetailsFetching, error: detailsFetchError } =
+        useQuery({
+            queryKey: [ADMIN_TOOL_REQUEST_DETAILS_QUERY_KEY, { id: requestId }],
+            queryFn: getAdminToolRequestDetails,
+            config: {
+                enabled: requestId && open,
+                onSuccess: (resp) => {
+                    setDetails(resp);
+                },
```
psarando, line 47:
```suggestion
                onSuccess: setDetails,
```

src/components/tools/requests/Details.js
```diff
@@ -0,0 +1,184 @@
...
+                            title="Details"
+                            actions={
+                                    <Button
+                                        id={build(parentId, ids.BUTTONS.OK)}
+                                        type="submit"
+                                        color="primary"
+                                    onClick={onClose}
+                                    >
+                                        {i18nCommon("ok")}
+                                </Button>
```
psarando, line 73: Looks like a weird formatting issue here (I'm surprised the `prettier` check didn't throw an error about it).

src/components/tools/requests/Details.js
```diff
@@ -0,0 +1,184 @@
...
+                                        label={t("toolName")}
+                                        id={build(
+                                            parentId,
+                                            ids.TOOL_REQUEST.NAME
+                                        )}
+                                        component={FormTextField}
+                                        InputProps={{
+                                            readOnly: true,
+                                        }}
+                                    />
```
psarando, line 102: All of these form fields are always read-only, and Formik's `handleSubmit` function is never called, so I don't think these need to be in a form at all 🤔 

It seems like all of these `FormTextField` components can be replaced with a vanilla MUI `TextField` instead.
For example here:
```suggestion
                                    <TextField
                                        value={details.name}
                                        label={t("toolName")}
                                        id={build(
                                            parentId,
                                            ids.TOOL_REQUEST.NAME
                                        )}
                                        InputProps={{
                                            readOnly: true,
                                        }}
                                        variant="outlined"
                                        margin="dense"
                                        fullWidth
                                    />
```

src/components/utils/UpdateRequestDialog.js
```diff
@@ -39,26 +47,40 @@ import {
...
+
+    const { isFetching: isToolRequestFetching, error: toolRequestFetchError } =
+        useQuery({
+            queryKey: [ADMIN_TOOL_REQUEST_DETAILS_QUERY_KEY, { id: requestId }],
+            queryFn: getAdminToolRequestDetails,
+            config: {
+                enabled: requestId && open && requestType === RequestType.TOOL,
+                onSuccess: (resp) => {
+                    setRequestDetails(resp);
+                },
```
psarando, line 76:
```suggestion
                onSuccess: setRequestDetails,
```

src/components/utils/UpdateRequestDialog.js
```diff
@@ -136,7 +175,7 @@ export default function UpdateRequestDialog(props) {
                                     baseId={baseId}
                                 />
                             )}
-                            {isRequestFetching && (
+                            {isDOIRequestFetching && (
```
psarando, line 178: Does this need to also check `isToolRequestFetching`?

src/components/utils/UpdateRequestDialog.js
```diff
@@ -155,7 +194,7 @@ export default function UpdateRequestDialog(props) {
                                 />
                             )}
 
-                            {!isRequestFetching && (
+                            {!doiRequestFetchError && (
```
psarando, line 197: Does this also need to check `toolRequestFetchError`?

psarando (approved):
LGTM 🎉 

I just had a couple of questions, and some minor refactoring suggestions 👍

ash3rz (approved):
LGTM :shipit: 
One weird thing from the screenshots is the tool names look like they're sometimes indented?  I didn't see anything weird in the code though.
Also, very minor thing, but in the Update Dialog (which I don't think was part of this PR) the `date` column is all lowercase and should probably be `Date`, same for the `comments` column as well as the `comments` label for the form.

src/components/tools/requests/Details.js
sri2k1us: Good catch. But I guess, `Formik` provides `json` to form field initialization / value mapping and I don't see any performance issues associated with `Formik`.  So unless there is strong preference to not use `Formik`, I will leave it alone for now.

sri2k1us:
> LGTM :shipit:
> One weird thing from the screenshots is the tool names look like they're sometimes indented? I didn't see anything weird in the code though.
 I think that is due to fact that the cell is rendered as `Button` which adds some margin / padding. You can see the same thing with `Status` column.

> Also, very minor thing, but in the Update Dialog (which I don't think was part of this PR) the `date` column is all lowercase and should probably be `Date`, same for the `comments` column as well as the `comments` label for the form.

src/components/tools/requests/Details.js
sri2k1us: Weird. When I format with VS Code extension prettier, it stays like this. But when I use our `npm` task, this was fixed. 😕

sri2k1us:
Thanks for the reviews.

PR #383, opened 2021-07-14
CORE-1400 Add method for checking maintenance flag

ash3rz:
This PR is mostly so others can review the way we're integrating Unleash - we're still waiting on a maintenance page design though I included a placeholder page for now.  This requires the 2 PRs in k8s-resources to be merged before this can be merged (though Unleash is already running in our cluster). 

When the maintenance flag is enabled, no matter which URL users try to go to, they'll see the maintenance page.
When the maintenance flag is disabled, if they're still currently viewing the maintenance page, refreshing will send them to the page they were originally trying to go to.

There's a placeholder starter maintenance HTML file and a CSS file with our primary and secondary colors, plus a copy-paste of the CSS for MUI's card -- all to be replaced by whatever design Mariah comes up with.


psarando (approved):
LGTM 👍

sri2k1us (approved):
🚢

ash3rz:
Thanks for the reviews!

PR #385, opened 2021-07-21
CORE-1061 Show error details for uploads

ash3rz:
Announcer:
image

Desktop:
image

Mobile:
image

### DIFF
--- a/src/components/uploads/api.js
+++ b/src/components/uploads/api.js
@@ -6,14 +6,9 @@
  * @module uploads/api
  */
 
-import { checkForError } from "../../common/callApi";
-import { trackIntercomEvent, IntercomEvents } from "common/intercom";
+import { IntercomEvents, trackIntercomEvent } from "common/intercom";
 
-import {
-    errorAction,
-    updateStatusAction,
-    setCancelFnAction,
-} from "../../contexts/uploadTracking";
+import { errorAction, setCancelFnAction, updateStatusAction } from "../../contexts/uploadTracking";
 
 /**
  * Starts a single upload.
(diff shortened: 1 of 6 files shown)


src/components/uploads/api.js
```diff
@@ -79,18 +88,17 @@ export const startUpload = (
...
-            if (e.details) {
-                console.error(
-                    `${e.details.code} ${JSON.stringify(e.details.context)}`
-                );
-            } else {
-                console.error(e.message);
+            let errorMessage = e;
+            if (e.data) {
+                const { error_code, ...rest } = e.data;
+                errorMessage = `${error_code} ${JSON.stringify(rest)}`;
```
psarando, line 94: I wonder if we should have a friendlier error message for common, known errors, like `ERR_EXISTS`, and leave this generic message only for unexpected errors.

For example: `A file with that name already exists at that location`.
The full path could also be parsed from `e.data` and included, in case the user is uploading multiple files to different destination folders 🤔

psarando (approved):
LGTM 👍 

Though I wonder if we could have a friendlier error message just for `ERR_EXISTS` errors.

src/components/uploads/api.js
sri2k1us: Agree with @psarando . We already do something like that in Rename Dialog https://github.com/cyverse-de/sonora/blob/d37ddaedd4f96a312018dbe6e464e1d443903ff8/src/components/data/RenameDialog.js#L56 and CreateFolderDialog https://github.com/cyverse-de/sonora/blob/d37ddaedd4f96a312018dbe6e464e1d443903ff8/src/components/data/CreateFolderDialog.js#L51

sri2k1us (commented):
LGTM but I think we should fix the error message before merging this.

ash3rz:
Thanks for the reviews!  Screenshots updated with the new error message.

sri2k1us:
One final knit pick from the screenshot: To be consistent, can we call it `Error Details` instead of `Failure Details`

sri2k1us (approved):

psarando (approved):
LGTM ✨ 

`Error Details` sounds good to me too 👍

ash3rz:
Thanks for the reviews!

PR #387, opened 2021-07-24
CORE-1214 Allow App Editor and Launch steppers to be marked completed

psarando:
This PR will allow the App Editor and Launch form steps to be marked as completed after the user has visited the step, and if the step does not contain any errors, similar to the feature added to the Workflow Editor in #380.

In the following example, the user has visited the App Launch form's steps 1, 2, and 4, but skipped step 3:
App Launch - Step 3 skipped

In the following example, the user has visited every step in the App Editor without adding or filling out any fields, so all steps appear complete except for Step 1, since the app name and description are required but not filled in. Since these fields will not display an error until the user touches the field, the step also will not display an error until one of those fields displays an error:
App Editor - step 1 not complete

---

After the App Launch form is submitted, or the App Editor is saved, every step is marked as touched, so every step will be marked as completed (or with an error icon if there are errors in that step).

---

Since the App form steps can now also display help text or error messages (also added in #380), the App Info step in the App Editor now has some help text and an error message:
App Editor - Info step help text

App Info step error text:
App Editor - Info step error text

---

This PR has not added any other help text or error messages for any other App Editor or Launch form steps.


ash3rz (approved):
LGTM :shipit:

psarando:
Thanks for the review! 👍

PR #388, opened 2021-07-26
Remove react-select

sri2k1us:
- Remove react-select
- Update FormSearchField to use material-ui [autocomplete](https://material-ui.com/api/autocomplete/#autocomplete-api)


src/components/forms/FormSearchField.js
```diff
@@ -1,43 +1,67 @@
...
+        setSearchValue(val);
+    }, [value, setSearchValue, props.labelKey]);
+
+    const onOptionSelected = (event, option, reason) => {
         setFieldValue(field.name, option ? option[props.valueKey] : "");
     };
-    const onSearchBlur = (event) => {
-        setFieldTouched(field.name, true);
-    };
-
```
sri2k1us, line 28: Not sure how to do this in material-ui yet. There is `blurOnSelect`. May be use that ?

src/components/forms/FormSearchField.js
```diff
@@ -1,43 +1,67 @@
...
 }) => {
     const errorMsg = getFormError(field.name, touched, errors);
-    const onOptionSelected = (option) => {
+    const [searchValue, setSearchValue] = React.useState(value);
+
+    React.useEffect(() => {
+        const val = {};
+        val[props.labelKey] = value;
+        setSearchValue(val);
+    }, [value, setSearchValue, props.labelKey]);
```
sri2k1us, line 33: `value` will not show up in the input field without setting the `label`.

psarando (changes requested):
I think this could use some more local testing, in Storybook and in dev mode, but I found at least 1 critical issue so far, and a couple of cosmetic issues.

src/components/forms/FormSearchField.js
```diff
@@ -1,43 +1,67 @@
...
+                onInputChange={handleSearch}
                 onChange={onOptionSelected}
-                onBlur={onSearchBlur}
-                {...field}
-                {...props}
+                renderOption={(option, state) => renderCustomOption(option)}
+                renderInput={(params) => (
+                    <TextField
+                        {...params}
+                        label={helperText}
```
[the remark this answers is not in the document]
sri2k1us: Doh! I made this change and for some reason this one got discarded along with few others things I wanted to get rid of 😩

src/components/forms/FormSearchField.js
```diff
@@ -1,43 +1,67 @@
...
+                }
+                options={options}
+                size={size}
+                freeSolo={true}
+                value={searchValue}
+                onInputChange={handleSearch}
                 onChange={onOptionSelected}
-                onBlur={onSearchBlur}
-                {...field}
-                {...props}
```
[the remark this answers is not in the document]
sri2k1us: No specific reason other than I was ignorant about these since things seems to work without any noticeable issues.

src/components/forms/FormSearchField.js
sri2k1us: I was testing this like you suggested by setting `required` to true but that didn't show any error when the field was left blank. But imo, it doesn't make sense to have a search field required. We should use `FormSelectField` for those 🤔

psarando: This is just like any other text field, except it can search thousands of ontology terms the user can choose from, so they don't make a spelling mistake or typo with the term (Ramona also suggested that in the future the IRI could be saved in the metadata as well). The ontology terms in the API can also get updated from time to time. I don't think we can expect anyone to create a select field with thousands of options in our metadata templates.

The `onBlur` callback is still being removed from `field` in the destructuring at the top of this component, but that callback is how this field sets its `touched` to `true`, and the error message will only display if the field is touched.

Also `blurOnSelect` should probably be set, as you suggested earlier.

If this still does not allow the `TextField` to display the `Required` error message, then maybe the `{...field}` and `{...props}` need to be spread into the `TextField` instead (or maybe it should be a `FormTextField` which already has the logic of checking `getFormError` for its `helperText`, so we wouldn't have to do that here 🤔 ).

sri2k1us: No dice 🎲 . I have pushed all the changes I got. Further suggestions welcome.

stories/base/FormFields.stories.js
```diff
@@ -312,5 +313,103 @@ Timestamp.argTypes = {
...
+        },
+        {
+            id: "envo:class:http://purl.obolibrary.org/obo/ENVO_01000184",
+            iri: "http://purl.obolibrary.org/obo/ENVO_01000184",
+            label: "subtropical desert biome",
+            ontology_prefix: "ENVO",
+        },
+    ];
+    return (
+        <TestForm initialValue="subtropical desert biome">
```
psarando, line 383: Sorry, this `TestForm` is kind of weird, but now that the `required` boolean control is used in the Field, this `TestForm` should also use the `required` prop:
```suggestion
        <TestForm required={required} initialValue="subtropical desert biome">
```

src/components/forms/FormSearchField.js
psarando: I got it to work in Storybook by adding the `required={required}` prop to the `TestField`, though now I'm seeing weird behavior:

If I clear the value, the field is marked as an error with the `Required` message when the value is cleared (as expected), but then selecting an option from the search results doesn't clear the validation error at first. I added console logging to the validation function that shows the new value is set, but then immediately after an empty value is set, even though the selection value still displays in the field 😕 
Selecting another option then clears the validation error, but again the validation function logging shows that the new value is set, followed immediately by the old value:
```
SearchField handleSearch = index.js:114
values.test_field <empty string> index.js:114
SearchField handleSearch = desert tree line ecotone index.js:114
values.test_field desert tree line ecotone index.js:114
values.test_field <empty string> index.js:114
SearchField handleSearch = desert-alpine tree line ecotone index.js:114
values.test_field desert-alpine tree line ecotone index.js:114
values.test_field desert tree line ecotone index.js:114
```
I'm not sure what's causing this yet.

psarando: For some reason, removing `blurOnSelect` fixes this validation weirdness.

src/components/forms/FormSearchField.js
```diff
@@ -1,46 +1,96 @@
...
+                <FormTextField
+                    label={label}
+                    variant="outlined"
+                    form={form}
+                    field={field}
+                    helperText={helperText}
+                    required={required}
+                    {...props}
+                    {...params}
+                />
```
psarando, line 91: `FormTextField` already sets `variant="outlined"`, and now that all `props` are getting spread into this field, I think there's no longer a need to remove `label`, `helperText`, and `required` from the props in the destructuring above. If those props are omitted from the destructing, then they wouldn't have to be repeated here:
```suggestion
                <FormTextField
                    form={form}
                    field={field}
                    {...props}
                    {...params}
                />
```

src/components/forms/FormSearchField.js
```diff
@@ -1,46 +1,96 @@
...
     label,
     helperText,
     required,
-    form: { touched, errors, setFieldTouched, setFieldValue },
+    options,
+    size = "small",
+    renderCustomOption,
+    handleSearch,
+    form: { setFieldValue, ...form },
     ...props
```
psarando, line 23: The dev console in Storybook is now showing these warnings:
```
Warning: React does not recognize the `labelKey` prop on a DOM element.
Warning: React does not recognize the `valueKey` prop on a DOM element.
```
These could be added to the destructuring here, then they won't be spread into the `FormTextField` below.

src/components/forms/FormSearchField.js
```diff
@@ -1,46 +1,96 @@
...
-        setFieldTouched(field.name, true);
+    const [searchValue, setSearchValue] = React.useState(value);
+    const filter = createFilterOptions();
+    const { t } = useTranslation("common");
+
+    React.useEffect(() => {
+        const val = {};
+        val[props.labelKey] = value;
+        setSearchValue(val);
+    }, [value, setSearchValue, props.labelKey]);
```
psarando, line 33: Should this be using `valueKey` rather than `labelKey`?

src/components/forms/FormSearchField.js
```diff
@@ -1,46 +1,96 @@
...
+                    filtered.push({
+                        inputValue: params.inputValue,
+                        label: t("formatTermFreeTextOption", {
+                            inputValue: params.inputValue,
+                        }),
+                    });
+                }
+
+                return filtered;
+            }}
```
psarando, line 80: Minor style suggestion:
```suggestion
            filterOptions={(options, params) => {
                const filtered = filter(options, params);
                const { inputValue } = params;
                // Suggest the creation of a new value
                if (inputValue) {
                    filtered.push({
                        inputValue,
                        label: t("formatTermFreeTextOption", {
                            inputValue,
                        }),
                    });
                }

                return filtered;
            }}
```

psarando (changes requested):

sri2k1us:
Thanks @psarando for working on this 👍🏽

src/components/forms/FormSearchField.js
```diff
@@ -1,46 +1,91 @@
...
+                const { inputValue } = params;
+                // Suggest the creation of a new value
+                if (params.inputValue !== "") {
+                    filtered.push({
+                        inputValue: params.inputValue,
+                        label: t("formatTermFreeTextOption", {
+                            inputValue: inputValue,
+                        }),
+                    });
+                }
```
psarando, line 76: No need to reference `params.inputValue` anymore, and `inputValue: inputValue` is now redundant:
```suggestion
                const { inputValue } = params;
                // Suggest the creation of a new value
                if (inputValue) {
                    filtered.push({
                        inputValue,
                        label: t("formatTermFreeTextOption", { inputValue }),
                    });
                }
```

src/components/forms/FormSearchField.js
```diff
@@ -1,46 +1,91 @@
...
-                defaultInputValue={value}
-                onChange={onOptionSelected}
-                onBlur={onSearchBlur}
-                {...field}
-                {...props}
-            />
-            <FormHelperText>{errorMsg || helperText}</FormHelperText>
-        </FormControl>
+        <Autocomplete
+            id="searchField"
```
psarando, line 51: One (last?) critical issue left in this PR:
In a metadata template like `Environment Ontology Terms`, all 3 ontology search fields will have this same `searchField` ID, rather than an ID like `metadataTemplate.metadata.avus[2]`.

Either this `id` should be removed so that the correct `id` gets spread into the `FormTextField` from `...props`, or else this `id` should be set to the `id` from `props`.

sri2k1us: I set it on the `Autocomplete` itself as recommended by the MUI.

ash3rz (approved):
I didn't catch anything 👍 (though I'm not as knowledgeable on all the features that we were getting out of the old component)

psarando (approved):
Thanks for tackling this upgrade 🎉 :shipit: 👍

sri2k1us:
> Thanks for tackling this upgrade 🎉 :shipit: 👍

Thanks for helping out.

PR #389, opened 2021-07-28
CORE-1400 Add newly designed maintenance page

ash3rz:
This is pretty much copy-paste of Mariah's codepen, but with some changes:
* Updated link URLs (some were duplicates)
* Added some fallback fonts 
* Intercom
* The `Enable Maintenance Reminders` button

TO NOTE: I wasn't sure what URLs to put for `Data Store` and `Science APIs` so I linked to the cyverse.org pages for those.  I'm open to suggestions.

Desktop:
image

Mobile:
image
image
image

### DIFF
--- a/public/maintenance.html
+++ b/public/maintenance.html
@@ -38,15 +38,15 @@
                 back online.</p>
             <div class="d-grid gap-2 d-md-block mx-auto my-5">
                 <a class="btn btn-lg btn-maintenance" href="https://status.cyverse.org/"
-                   target="_blank">
+                   target="_blank" rel="noopener noreferrer">
                     View Status of All Services
                 </a>
                 <a class="btn btn-lg btn-maintenance"
-                   href="https://cyverse.org/maintenance" target="_blank">
+                   href="https://cyverse.org/maintenance" target="_blank" rel="noopener noreferrer">
                     View Maintenance Calendar
                 </a>
                 <a class="btn btn-lg btn-maintenance"
-                   href="https://user.cyverse.org/account" target="_blank">
+                   href="https://user.cyverse.org/account" target="_blank" rel="noopener noreferrer">
                     Enable Maintenance Reminders
                 </a>
             </div>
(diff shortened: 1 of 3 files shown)


public/maintenance.js
```diff
@@ -0,0 +1,35 @@
+window.intercomSettings = {
+    app_id: "tpwq3d9w",
```
sri2k1us, line 2: wonder if this can read from query param in the page url.
e.g: de.cyverse.org/maintenance.html?app_id=tpwq3d9w.

If this works, then we don't have hard code the `app_id` here but can read from config file on server side.

sri2k1us (approved):
LGTM. We should  into reading intercom app-id from url query-params.

public/maintenance.js
ash3rz: So with the current setup I wouldn't be able to do that.  I'm not modifying/redirecting the requested URL to the maintenance page (although, you can go directly to the maintenance page if you really wanted to).  
The URL that they requested (e.g. de.cyverse.org/apps/systemId/appId/) should stay the same, but they get sent back as a response the maintenance page instead. 
This was to allow them to hit refresh without being stuck on the maintenance page once we're out of maintenance (e.g. we turn off maintenance, they hit refresh, and the maintenance page becomes the app launch page they were trying to visit originally). That was a feature we had in the legacy DE, so I kept that, but we can go a different route if needed.
I did timebox looking into [templating](https://expressjs.com/en/guide/using-template-engines.html), but ran into issues where it wouldn't load the CSS file and the JS file couldn't access those variables. I could give it another go if we want to explore that more though.

ash3rz: Oops, to be more specific - this was a feature we had when a user ended up on one of our error pages (/de/error I think).

public/maintenance.html
```diff
@@ -1,32 +1,92 @@
...
+            <div class="d-grid gap-2 d-md-block mx-auto my-5">
+                <a class="btn btn-lg" style="background-color:#0971ab;color:#ffff;" href="https://status.cyverse.org/"
+                   target="_blank">
+                    View Status of All Services
+                </a>
+                <a class="btn btn-lg" style="background-color:#0971ab;color:#ffff;"
+                   href="https://cyverse.org/maintenance" target="_blank">
+                    View Maintenance Calendar
+                </a>
+                <a class="btn btn-lg" style="background-color:#0971ab;color:#ffff;"
```
psarando, line 46: Maybe these colors can be added to a class rather than repeated in the styles?

public/maintenance.html
```diff
@@ -1,32 +1,92 @@
...
+        src="https://cyverse.org/sites/default/files/inline-images/social_icon-01.png" class="icon" alt="twitter" /></a>
+    <a href="https://www.facebook.com/CyVerse.org/" target="_blank"> <img
+        src="https://cyverse.org/sites/default/files/inline-images/social_icon-02.png" class="icon"
+        alt="facebook" /></a>
+    <a href="https://www.youtube.com/channel/UC-gvdjTz9rq6RovZ57LoDDA/featured" target="_blank"><img
+        src="https://cyverse.org/sites/default/files/inline-images/social_icon-03.png" class="icon" alt="youtube" /></a>
+    <a href="https://www.linkedin.com/company/cyverse.org/" target="_blank"> <img
+        src="https://cyverse.org/sites/default/files/inline-images/social_icon-04.png" class="icon"
+        alt="linkedin" /></a>
+    <a href="https://github.com/cyverse/" target="_blank"> <img
```
psarando, line 85: All of these links that use the `target="_blank"` attribute should also set a `rel="noopener noreferrer"` attribute (see https://mathiasbynens.github.io/rel-noopener/ for example).

All `target="_blank"` links could include this safety measure, though it's probably not as crucial to include `noreferrer` in links to other cyverse.org pages.

psarando (changes requested):
The UX LGTM 🎉 ✨ 

Though I have some standards and security suggestions.

psarando (approved):

ash3rz:
Thanks for the reviews!

PR #390, opened 2021-07-31
CORE-1465 Add `tools.admin.max_*` resource config support

psarando:
This PR will add `tools.admin.max_*` resource config support, and also allow arbitrary max resource limits in app/tool forms.

The app and tool resource request/limit select fields will still be populated with powers of 2, but the configs can set a max value that is not a power of 2 (such as our actual limits in the cluster), and that max value will still display as the upper limit in the selection list.

For example, if `tools.admin.max_memory_limit` is set to 244GiB, then the tool admin form list would be populated like this:
Tool Admin Max Memory list

This PR will also fix a bug in the Edit Tool form that could display a power of 2 beyond the actual limit set in the configs.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -23,6 +23,15 @@ module.exports = withBundleAnalyzer({
         SESSION_POLL_INTERVAL_MS: config.has("sessions.poll_interval_ms")
             ? config.get("sessions.poll_interval_ms")
             : 5000,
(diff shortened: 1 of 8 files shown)


ash3rz (approved):
LGTM :shipit:

sri2k1us (approved):
:shipit:

psarando:
Thanks for the reviews! 👍

PR #397, opened 2021-08-11
Fix req body parsing

ash3rz:
I was digging into this because I noticed that the exact same request body for a POST request when used in a DELETE request would return an error from the service that the body was not a map and was nil. This appears to be the first DELETE request with a request body added to the new DE. 

Reading through [the express docs](https://expressjs.com/en/4x/api.html#req.body), I saw that they recommend using `express.json()` to parse the req body otherwise it's undefined.  Not sure how it was working so well before this, but either way, this seems to be the recommended way to do things.

I also bumped the axios version since we keep getting dependabot PRs for it.  I tested several different requests in the DE and didn't notice any issues.

### DIFF
--- a/src/server/api/terrain.js
+++ b/src/server/api/terrain.js
@@ -27,18 +27,16 @@ import { replaceNamedPathSegments } from "../../common/functions";
 export const handler = ({ method, pathname, headers }) => {
     return async (req, res) => {
         const accessToken = req?.kauth?.grant?.access_token;
-        call(
-            {
-                method,
-                pathname,
-                headers,
-                query: req.query,
-                params: req.params,
-                userID: accessToken?.content?.preferred_username,
-                accessToken: accessToken?.token,
-            },
-            req
-        )
+        call({
+            method,
+            pathname,
+            headers,
+            query: req.query,
+            params: req.params,
+            userID: accessToken?.content?.preferred_username,
+            accessToken: accessToken?.token,
+            data: req?.body,
+        })
             .then((apiResponse) => {
                 res.set(apiResponse.headers);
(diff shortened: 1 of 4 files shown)


src/server/api/terrain.js
```diff
@@ -27,18 +27,16 @@ import { replaceNamedPathSegments } from "../../common/functions";
...
-            {
-                method,
-                pathname,
-                headers,
-                query: req.query,
-                params: req.params,
-                userID: accessToken?.content?.preferred_username,
-                accessToken: accessToken?.token,
-            },
-            req
```
ash3rz, line 40: Previously we were sending the entire request object, which below gets renamed to `inStream`, and was being set as the request body

src/server/api/terrain.js
```diff
@@ -90,20 +94,19 @@ export const call = (
         apiURL.search = querystring.stringify(query);
     }
 
-    logger.info(`TERRAIN ${userID} ${method} ${apiURL.href}`);
+    logger.info(
+        `TERRAIN ${userID} ${method} ${apiURL.href} ${JSON.stringify(data)}`
```
ash3rz, line 98: I added the request body to the log, but we don't need it if others think this will be noisy or is unnecessary

sri2k1us: Logging request body is helpful 👍🏽

sri2k1us (approved):
🚢

src/server/api/terrain.js
psarando: Let's try it out for a while, we can always revert it if it gets too noisy 👍

ash3rz:
Thanks for the reviews!

PR #402, opened 2021-08-14
CORE-1475: updated the VICE loading page to work when the CSI driver …

slr71:
…is enabled


ash3rz (approved):
LGTM :shipit:

slr71:
Thanks for the review!

PR #403, opened 2021-08-17
CORE-1485 Rename Quick Launch to Saved Launch

ash3rz:
To be honest, before this ticket I completely forgot we had 2 types of launches (quick and instant) and I found myself getting confused on which was which. So I renamed everything instead of only updating the display text.

* I also removed some files related to AppTile which wasn't in use
* Removed some i18n strings that weren't actually being used
* I did NOT rename the quick launches URL
* I left server side code completely alone


public/static/locales/en/launch.json
```diff
@@ -37,7 +37,7 @@
     "resourceRequirements": "Resource Requirements",
     "resourceRequirementsForStep": "Resource Requirements for Step {{step}}",
     "reviewAndLaunch": "Review and Launch",
-    "saveAsQuickLaunch": "Save as Quick Launch",
+    "saveAsSavedLaunch": "Save as Saved Launch",
```
ash3rz, line 40: This one and the one above sound a little strange.  Not sure if `Create as Saved Launch` or just `Save Launch` might be better?

sri2k1us (approved):
LGTM. I have couple of suggestions but we can also discuss that in a UX meeting.

psarando (changes requested):
Looking good 👍 

I agree with changing the app launch step label to just "Launch or Save".

I also really think we should only be providing links with the `saved-launch-id` query param, since I consider these links as also user-facing, and we can still accept both `quick` and `saved` in the launch page.

public/static/locales/en/launch.json
ash3rz: I started to go with `Save`, but then with App Launch that looked a little confusing next to `Launch Analysis` on which action I should take next.
image

I'll go with `Create Saved Launch` for now. `Create` makes it sound like I'm generating something new where `Save` sounds like I'm saving my progress on this form while continuing to work on it.  But I'm still open to ideas on this one.

psarando (approved):
LGTM 👍

ash3rz:
Thanks for the reviews!

PR #404, opened 2021-08-18
CORE-1483 Show running time for analyses in dashboard

ash3rz:
The first running analysis under `Recent Analyses` is the example in this screenshot. It shows the amount of time that has passed (the text is coming from [date-fns](https://date-fns.org/v2.23.0/docs/formatDistanceToNow)) and the date in parentheses. Completed analyses only show the date.

image

Also removed some unused code.

### DIFF
--- a/src/components/dashboard/dashboardItem/ItemBase.js
+++ b/src/components/dashboard/dashboardItem/ItemBase.js
@@ -1,6 +1,5 @@
-import React, { useEffect, useState } from "react";
+import React from "react";
 import clsx from "clsx";
-import { formatDistanceToNow } from "date-fns";
 import { Trans, useTranslation } from "i18n";
 import ReactPlayer from "react-player/youtube";
 
@@ -16,8 +15,8 @@ import {
     useTheme,
 } from "@material-ui/core";
 
+import useAnalysisRunTime from "components/analyses/useAnalysisRunTime";
 import buildID from "components/utils/DebugIDUtil";
-import { isTerminated } from "components/analyses/utils";
 
 import ids from "../ids";
 import * as constants from "../constants";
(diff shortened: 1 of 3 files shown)


psarando (approved):
LGTM 👍

src/components/dashboard/dashboardItem/ItemBase.js
```diff
@@ -45,6 +45,45 @@ const DashboardLink = ({ target, kind, children }) => {
...
+function AnalysisSubheader(props) {
+    const { analysis, date: formattedDate } = props;
+    const { t } = useTranslation(["dashboard", "apps"]);
+    const terminated = isTerminated(analysis);
+
+    const [runningTime, setRunningTime] = useState(null);
+
+    useEffect(() => {
+        const handleUpdateRunningTime = () => {
+            setRunningTime(formatDistanceToNow(new Date(analysis.start_date)));
```
sri2k1us, line 57: So I was talking to @slr71 about this feature the other day and she said this is the job submission time 🤔  and not the actual the time when job actually started running. She suggested we query job status history to get the actual time when the job started running.

src/components/dashboard/dashboardItem/ItemBase.js
```diff
@@ -45,6 +45,45 @@ const DashboardLink = ({ target, kind, children }) => {
...
+        };
+
+        let interval;
+        if (!terminated) {
+            interval = setInterval(handleUpdateRunningTime, 60000);
+            handleUpdateRunningTime();
+        }
+
+        return () => clearInterval(interval);
```
sri2k1us, line 66: Also we might want to move some of this logic to some where like `utils` since this is requested in `Analyses` view as well.

sri2k1us (commented):
I think we should get clarification on analysis submission time vs running time

src/components/dashboard/dashboardItem/ItemBase.js
ash3rz: Sure, I'll take a look at that endpoint

sri2k1us:
LGTM. 
1 minor UX change (since that part of the code is touched) I would like to suggest is to add highlight color to Analysis status.
For E.g: `Completed` in `palette.cobalt`, `Running` - `palette.grass` and `Failed` - `palette.alertRed`

sri2k1us (approved):
:shipit:

ash3rz:
Thanks for the reviews!

PR #406, opened 2021-08-31
Add collections to apps and update collections icon

ash3rz:
NOTE: This branch is on top of #405. The first commit can be ignored.

The collections endpoints can take a while, so I opted to start caching the My Collections query to make opening the dialog with the list of your collections a little less annoying.

New icon in the navigation drawer and showing up in the Apps view dropdown:
image

Dialog to pick a collection:
image
image

Collection listing in apps view:
image

### DIFF
--- a/src/components/collections/Listing.js
+++ b/src/components/collections/Listing.js
@@ -5,10 +5,10 @@
  * collections they are currently following
  */
 
-import React, { useMemo, useState } from "react";
+import React, { useEffect, useMemo, useState } from "react";
 
 import buildID from "components/utils/DebugIDUtil";
-import { useQuery } from "react-query";
+import { queryCache, useQuery } from "react-query";
 
 import BasicTable from "components/table/BasicTable";
 import WrappedErrorHandler from "components/error/WrappedErrorHandler";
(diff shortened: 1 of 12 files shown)


src/components/collections/Listing.js
```diff
@@ -0,0 +1,129 @@
...
+            queryFn: getMyCollections,
+            config: {
+                enabled:
+                    !myCollectionsCache &&
+                    COLLECTION_FILTER.MY_COLLECTIONS === filter,
+                onSuccess: (results) => {
+                    setData(results.groups);
+                },
+            },
+        });
```
sri2k1us, line 95: For caching, usually I follow these steps:
 - Increase cache / stale time for the related query (or set it to infinity)
 - Try to get the query data from cache when needed. if thats not available, then fetch it 
 - Update query data when new data is available or simply invalidate that query. 

Not sure if am missing any thing but I didn't see the step 1 happening here. Also, Not sure how things are will RQ3 as well 🤔 since queryCache is gone and I saw some weirdness with bags 😖 in RQ3 .

ash3rz: For step 1, I think we're increasing the cache/stale time to Infinity in cases where we only want to fetch the query once, and if it's successful, never fetch it again (e.g. bootstrap, tool types). 
In this case, I want it to stay cached across the 2 pages (collections and apps) until the user (un)follows, creates, or deletes a new collection.  I might have missed something, but when I tried locally adding the cache/stale time and setting them to Infinity, and invalidated the query during those mutations, the query wouldn't refresh. I'd still get back the old list that was previously cached.
If this ends up being an issue with RQ3, we can either remove the caching (I just thought it was more convenient to have it cached so the user doesn't have to keep waiting for the list to populate) or we can switch to using a context.

sri2k1us: Ok. We can revisit again after the RQ3 update.

sri2k1us (approved):
:shipit:

psarando (approved):
I didn't catch any issues 👍

ash3rz:
Thanks for the reviews!

PR #408, opened 2021-09-02
Add completed analyses total run time to dashboard

ash3rz:
image

### DIFF
--- a/src/components/analyses/useAnalysisRunTime.js
+++ b/src/components/analyses/useAnalysisRunTime.js
@@ -21,13 +21,14 @@ import {
     ANALYSIS_HISTORY_QUERY_KEY,
     getAnalysisHistory,
 } from "serviceFacades/analyses";
+import { isTerminated } from "./utils";
 
 function useAnalysisRunTime(
     analysis,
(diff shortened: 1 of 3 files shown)


src/components/analyses/useAnalysisRunTime.js
```diff
@@ -26,15 +27,17 @@ function useAnalysisRunTime(
     stepFilterFn = (step) => step.step_number === 1
 ) {
     const isRunning = analysis?.status === analysisStatus.RUNNING;
+    const isComplete = analysis?.status === analysisStatus.COMPLETED;
```
ash3rz, line 30: I went with just completed as opposed to terminated (which includes cancelled, failed, and removed) because I wasn't sure if there would be interest in the run time for all those states, but let me know if others should be added.

sri2k1us: I vote to add this feature for all terminal states. 👍🏽

sri2k1us (approved):
:shipit:

psarando (approved):
LGTM 👍

ash3rz:
Thanks for the reviews!

PR #409, opened 2021-09-03
Add notification categories to i18n

ash3rz:
Switching from using the text in the `NotificationCategory` model as what's displayed to the user, to what we have in the notification i18n strings. Also display the Community category as Collection instead. Since the model is unchanged, all the service calls which use the model to create filters should still work as expected.

image

image

From storybook to help show at least one of each type:
image


sri2k1us (approved):
🚢

ash3rz:
Thanks for the review!  I'll merge this now so I can re-run the QA tests before the weekend, but additional reviews welcome.

PR #413, opened 2021-09-13
CORE-1482 Add Featured Apps category to Apps view

ash3rz:
image

image


src/constants.js
```diff
@@ -16,6 +16,7 @@ export default {
     APPS_UNDER_DEV: "Apps under development",
     FAV_APPS: "Favorite Apps",
     MY_PUBLIC_APPS: "My public apps",
+    FEATURED_APPS: "Featured Apps",
     HPC: "High-Performance Computing",
     BROWSE_ALL_APPS: "Browse All Apps",
     BROWSE_ALL_APPS_ID: "pppppppp-pppp-pppp-pppp-pppppppppppp",
```
sri2k1us, line 22: While we are in it, can we fix the casing. For example, `Apps under development` - > `Apps Under Development`, `My public apps` -> `My Public Apps` etc..

sri2k1us (commented):
I think Nirav / Tyson wanted `Featured Apps` to be the default selection. Can you please check with them ?

src/components/apps/toolbar/AppNavigation.js
```diff
@@ -200,6 +202,10 @@ function AppNavigation(props) {
         setAnchorEl(null);
     };
 
+    const categoryI18nName = (name) => {
+        return name.toLowerCase().replaceAll(" ", "_");
+    };
```
psarando, line 207: ✨

psarando (approved):
LGTM 👍

sri2k1us (approved):
LGTM 🚢

src/constants.js
```diff
@@ -16,9 +16,11 @@ export default {
...
+    FEATURED_APPS_ID: "cccccccc-cccc-cccc-cccc-cccccccccccc",
```
ash3rz, line 23: Wasn't sure if this should be in a config instead - I'll gladly add it if that's the preference

psarando: My instinct is to say that it should be in a config, but we already have `BROWSE_ALL_APPS_ID` as a constant. Plus I don't think there's a way to update configs without rebuilding the entire UI right now anyway 🤷

psarando (approved):
LGTM 👍

PR #415, opened 2021-09-21
Core 1494 - Fix data search for logged out users.

sri2k1us:
- Fetch bootstrap info if it not available cache
- Restrict logged out data search results to `Community Data` folder.


src/components/search/dataSearchQueryBuilder.js
```diff
@@ -9,12 +11,13 @@ export const getDataSimpleSearchQuery = (
...
         query: {
-            all: userHomeDir
+            all: !isDetailed
                 ? searchClauses.concat({
                       type: "path",
-                      args: { prefix: userHomeDir },
+                      args: { prefix: pathPrefix },
                   })
                 : searchClauses,
         },
```
sri2k1us, line 23: For detailed search page, we don't want restrict results to any particular path.

ianmcorvidae: Is this right? I thought that for logged-out users we still wanted to limit to community data only.

ianmcorvidae (changes requested):
I think the main bug isn't fixed (since logged-out users can still see things shared with anonymous on the detailed search page, even though they now won't be able to in the global search field), but everything else looks like it's correct/an improvement.

ianmcorvidae (approved):
Looks good to me!

src/components/search/GlobalSearchField.js
```diff
@@ -365,10 +370,33 @@ function GlobalSearchField(props) {
...
+                homeDir = homeDir + "/";
+                setUserHomeDir(homeDir);
+            }
+        },
+        [bootstrapCache]
+    );
+
+    const { isFetching: isBootStrapFetching } = useBootStrap(
+        bootstrapQueryEnabled,
+        (respData) => preProcessData(respData),
```
psarando, line 387:
```suggestion
        preProcessData,
```

psarando (approved):
LGTM 👍

sri2k1us:
Thanks for the reviews.

PR #416, opened 2021-09-22
Core 1498 Fix preferences validation

sri2k1us:


src/components/preferences/Preferences.js
```diff
@@ -383,8 +384,14 @@ function Preferences(props) {
...
     }
-    const validateShortCuts = (values, props) => {
+    const validate = (values, props) => {
         const errors = {};
+        if (values?.webhook?.url) {
+            const type = values?.webhook?.type?.type;
+            if (type === null || type === undefined || type === '') {
+                errors["webhook.type.type"] =  t("webhookTypeError");
+            }
+        }
```
sri2k1us, line 394: This validation will run but will not mark errors on the form 😭

src/components/preferences/Webhooks.js
```diff
@@ -82,11 +90,10 @@ export default function Webhooks(props) {
                         name="webhook.type.type"
                         id={buildID(baseId, ids.WEBHOOK_TYPES_SELECT)}
                         label={t("type")}
-                        required
                         select
-                        variant="outlined"
+                        variant="standard"
                         size="small"
-                        component={FormTextField}
+                        component={FormSelectField}
```
sri2k1us, line 96: If I add field level validation, it will work but I have to touch the field first 😕 . Also I had to change it `FormSelectField` instead of `FormTextField`

src/serviceFacades/users.js
```diff
@@ -94,7 +93,7 @@ function updateWebhooks(webhooks) {
     return callApi({
         endpoint: `/api/webhooks`,
         method: "PUT",
-        body: webhooks,
+        body: webhooks ? webhooks : { "webhooks": [] },
```
sri2k1us, line 96: Allow users to delete webhooks.

src/components/preferences/Webhooks.js
```diff
@@ -61,6 +63,12 @@ export default function Webhooks(props) {
         },
     });
 
+    React.useEffect(() => {
+        const hasURL = values?.webhook?.url ? true : false;
+        setEnableTestButton(hasURL && !isTesting);
+    }, [isTesting, values]);
+
```
sri2k1us, line 70: Enable `Test` button only when there is url or another test is not in progress.

src/components/preferences/Preferences.js
ash3rz: Going off memory, I think `errors` is an object with the same format as the model you provided. 
So `errors["webhook.type.type"]` would be accessing an object like `{"webhook.type.type": "someValue"}` instead of `{webhook: {type: {type: "someValue"}}}`.  So I think if you update this to `errors.webhook.type.type` you'll get the result you want.

src/components/preferences/Webhooks.js
ash3rz: Having to touch the field first sounds right to me. If the user opens a blank form, for example, we don't want every field to automatically be red just because they haven't filled it out yet. So validation should run after a field has been touched or when the user tries to submit.

src/components/preferences/Preferences.js
sri2k1us: Ahh that makes sense although `errors.webhook.type.type = err msg` resulted in error. So I guess I have to construct the error object manually 🤔

ash3rz: Actually that reminds me! I don't think it's advertised anywhere, but Formik has a `setIn` function (similar to the `getIn` function that we're using in our helper form fields) that would be useful in this case.

```
errors = setIn(errors, "webhook.type.type", t("webhookTypeError"))
```

sri2k1us: `errors = { webhook: { type: { type: t("webhookTypeError") } } };` worked for me 🎉

src/components/preferences/Webhooks.js
```diff
@@ -61,6 +63,12 @@ export default function Webhooks(props) {
         },
     });
 
+    React.useEffect(() => {
+        const hasURL = values?.webhook?.url ? true : false;
+        setEnableTestButton(hasURL && !isTesting);
+        setFieldTouched("webhook.type.type", hasURL, false);
+    }, [isTesting, setFieldTouched, values]);
+
```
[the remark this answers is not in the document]
ash3rz: This seems ok to me, but I'll defer to @psarando since he probably has the most experience with Formik.  

An alternative might be to add a [validate](https://formik.org/docs/api/field#validate) prop to the url field which does this same logic. That way, the logic should only run when the url value changes instead of when any value in `values` changes.

sri2k1us:
Also I think we should `hide` KB shortcuts we can implement them.

src/components/preferences/Preferences.js
```diff
@@ -383,8 +384,14 @@ function Preferences(props) {
             </>
         );
     }
-    const validateShortCuts = (values, props) => {
-        const errors = {};
+    const validate = (values, props) => {
+        let errors = {};
+        if (values?.webhook?.url) {
+            const type = values?.webhook?.type?.type;
+            if (type === null || type === undefined || type === "") {
```
psarando, line 391: I think this is a bit simpler:
```suggestion
            if (!values?.webhook?.type?.type) {
```

src/components/preferences/Webhooks.js
```diff
@@ -82,7 +90,6 @@ export default function Webhooks(props) {
                         name="webhook.type.type"
                         id={buildID(baseId, ids.WEBHOOK_TYPES_SELECT)}
                         label={t("type")}
-                        required
```
psarando, line 85: This could actually be set to `true` if there is a value entered for the URL field:
`required={!!(values?.webhook?.url)}`

src/components/preferences/Webhooks.js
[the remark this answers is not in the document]
psarando: I think it's acceptable to only show the validation error message if the user actually touches the Type field, or attempts to save the form without the Type set (and as far as I can tell the Type isn't required for testing the URL).

If we set the `required` prop on the Type field based on the URL field's value, then I don't think we need to manually set this field as "touched" here.

src/components/preferences/Webhooks.js
sri2k1us: Actually, doing this only partially solved the problem 🤔 . Now the `formik` built in `Required` validation marker shows up but our custom error message still doesn't show until the field is touched.

sri2k1us: image

sri2k1us: After field is touched:
image

psarando: This might be because every field in the form should have an initial value, even if it's an empty string: https://github.com/formium/formik/issues/445#issuecomment-366952762

So the `mapPropsToValues` might need to return an empty webhook.

Looking at these form field names, maybe it needs to return something like this:
```javascript
return {
    ...bootstrap.preferences,
    webhook: { url: "", type: { type: "" } },
}
```

sri2k1us: Ahh yes. I keep forgetting to check initial values.

src/components/preferences/Webhooks.js
```diff
@@ -82,11 +88,15 @@ export default function Webhooks(props) {
...
                         select
                         variant="outlined"
                         size="small"
                         component={FormTextField}
+                        validate={(value) => {
+                            if (values?.webhook?.url && !value) {
+                                return t("webhookTypeError");
+                            }
+                        }}
                     >
```
sri2k1us, line 100: Finally got the validation to work by removing `required` and adding field level validation.

src/components/preferences/Preferences.js
```diff
@@ -405,20 +405,45 @@ function Preferences(props) {
...
                     }
                 }
             }
         }
         return errors;
     };
-
+ */
     const mapPropsToValues = (bootstrap) => {
+        const emptyPref = {
```
psarando, line 421: ✨

src/components/preferences/Preferences.js
```diff
@@ -527,14 +553,20 @@ function Preferences(props) {
...
                                 requireAgaveAuth={requireAgaveAuth}
                                 resetHPCToken={resetHPCToken}
+                            />
+                            <Divider className={classes.dividers} />
+                            <Webhooks
+                                baseId={buildID(baseId, ids.WEBHOOK_PREF)}
                                 webhookTopics={webhookTopics}
                                 webhookTypes={webhookTypes}
                                 values={props.values}
+                                setFieldTouched={props.setFieldTouched}
```
psarando, line 563: Looks like this is no longer needed:
```suggestion
```

psarando (approved):
LGTM 👍

ash3rz (approved):
LGTM :shipit:

sri2k1us:
Thanks for the reviews.

PR #417, opened 2021-09-23
Fix unexpected error when refreshing Data page

ash3rz:
There might be a race condition between when the user is considered logged in vs out.  The case I saw in the console was where the path was `/cyverse/home/aramsey`, but the `dataRoots` were all set for the anonymous user.  This resulted in the `selectedIndex` being set to `-1` which threw an error in the render function when trying to access `dataRoots[selectedIndex]`.

### DIFF
--- a/src/components/data/toolbar/Navigation.js
+++ b/src/components/data/toolbar/Navigation.js
@@ -416,27 +416,25 @@ function Navigation(props) {
          * SHARED_WITH_ME_PATH: "/iplant/home",
          */
         if (dataRoots.length > 0 && path) {
+            let index = 0;
             if (path.startsWith(userHomePath)) {
-                setSelectedIndex(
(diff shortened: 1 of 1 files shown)


ianmcorvidae:
This all looks good to me, but the one question I have is if we can also help with the race condition somehow -- maybe whatever calls this code needs to be guarded behind something that ensures the data roots call isn't in a `loading` type state? (And I guess we might want to explicitly invalidate/purge the data roots when logging in or logging out, if we aren't. but we might be already, I didn't look

psarando (approved):
LGTM 👍

sri2k1us (approved):
:shipit:

ash3rz:
Thanks for the reviews!  I'll merge this for now as it's causing some fallout in the tests but I agree this deserves more investigation.

PR #418, opened 2021-09-23
Fixing misc warnings in the console

ash3rz:


ianmcorvidae (approved):
gotta love that sweet sweet "make the console stop yelling about things" project

src/components/layout/UserPortalUpdatePrompts.js
```diff
@@ -140,7 +140,6 @@ function UserPortalUpdatePrompts() {
                 open={!!userPortalResp?.update_required}
                 id={baseId}
                 fullWidth
-                disableBackdropClick
```
psarando, line 143: Part of the deprecation warning for this prop says:
> Use the onClose prop with the `reason` argument to filter the `backdropClick` events.

We should test this dialog to ensure that users can not close it by clicking outside of it.

psarando (approved):
LGTM 👍 ✨ 

Though we should ensure that users can't close the dialog after removing the `disableBackdropClick` prop.

sri2k1us (approved):
:shipit:

src/components/layout/UserPortalUpdatePrompts.js
ash3rz: Good catch 👍 I did test it within FireFox, Chrome, and Safari and the dialog doesn't close when clicking outside of it.  I think having `open` based on the user portal response and not having an `onClose` defined is probably preventing it from closing without that explicit setting.

ash3rz:
Thanks for the reviews!

PR #420, opened 2021-09-30
CORE-1484 Add "first class" instant launches to navigation drawer

ash3rz:
NOTE: The terrain and k8s PR will need to be merged prior to this one.  Also may require some additional testing once the QA environment is back up.

image

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -89,3 +89,6 @@ unleash:
 
 metadata:
     community_attr: cyverse-community
+    instant_launch_location_attr: ui_location
+    instant_launch_dashboard: dashboard
+    instant_launch_nav_drawer: nav_drawer
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -87,8 +87,3 @@ unleash:
     client_secret: null
     maintenance_flag: DE-Maintenance
 
-metadata:
-    community_attr: cyverse-community
-    instant_launch_location_attr: ui_location
-    instant_launch_dashboard: dashboard
-    instant_launch_nav_drawer: nav_drawer
--- a/src/components/instantlaunches/InstantLaunchButtonWrapper.js
+++ b/src/components/instantlaunches/InstantLaunchButtonWrapper.js
@@ -1,0 +1,73 @@
+/**
+ * @author aramsey
+ *
+ * A component meant to wrap any other clickable element that should launch
+ * an instant launch.
+ *
+ * The render component will be passed an `onClick` function.
+ * The `onClick` function will handle launching the instant launch and
+ * briefly displaying the Submission dialog to the user.
+ */
+import React from "react";
+
+import { useMutation } from "react-query";
+
+import { useDefaultOutputDir } from "components/data/utils";
+import withErrorAnnouncer from "components/error/withErrorAnnouncer";
+import { getHost } from "components/utils/getHost";
+import constants from "constants.js";
+import { InstantLaunchSubmissionDialog } from "./index";
+import { instantlyLaunch } from "serviceFacades/instantlaunches";
+
+function InstantLaunchButtonWrapper(props) {
+    const { instantLaunch, resource = {}, render, showErrorAnnouncer } = props;
+    const output_dir = useDefaultOutputDir();
+
+    const [open, setOpen] = React.useState(false);
+    const [ilUrl, setIlUrl] = React.useState();
+
+    React.useEffect(() => {
+        if (ilUrl) {
+            window.open(`${getHost()}${ilUrl}`);
+            setIlUrl(null);
+            setOpen(false);
+        }
+    }, [ilUrl]);
+
+    const { mutate: launch } = useMutation(instantlyLaunch, {
+        onSuccess: (listing) => {
+            if (listing.analyses.length > 0) {
+                const analysis = listing.analyses[0];
+                if (analysis.interactive_urls.length > 0) {
+                    setIlUrl(
+                        `${constants.VICE_LOADING_PAGE}/${encodeURIComponent(
+                            analysis.interactive_urls[0]
+                        )}`
+                    );
+                } else {
+                    setOpen(false);
+                }
+            } else {
+                setOpen(false);
+            }
+        },
+        onError: (err) => {
+            setOpen(false);
+            showErrorAnnouncer(err.message, err);
+        },
+    });
+
+    const onClick = () => {
+        setOpen(true);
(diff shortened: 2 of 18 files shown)


src/components/instantlaunches/InstantLaunchButtonWrapper.js
```diff
@@ -0,0 +1,73 @@
...
+    const onClick = () => {
+        setOpen(true);
+        launch({ instantLaunch, resource, output_dir });
+    };
+
+    return (
+        <>
+            {render(onClick)}
+            <InstantLaunchSubmissionDialog open={open} />
+        </>
```
ash3rz, line 69: I extracted all of the above logic from the `InstantLaunchButton` in `src/components/instantlaunches/index.js` so I could reuse it in the navigation drawer.

sri2k1us:
Nirav wants to use 
image
 for Cloud Shell app. Can we hard-code/ use some constants to get that icon in place for now until we get custom app icon feature implemented ?

config/default.yaml
```diff
@@ -89,3 +89,6 @@ unleash:
 
 metadata:
     community_attr: cyverse-community
+    instant_launch_location_attr: ui_location
+    instant_launch_dashboard: dashboard
+    instant_launch_nav_drawer: nav_drawer
```
sri2k1us, line 94: Am not entirely sure if these needs to be in config. I don't see them changing in different envs / external deployments.  They could live in some constants file and can be moved to configs when it is determined this will in fact be changing often. Also there is no real benefit to adding them in configs since we still have to redeploy the UI once the config changes ?

sri2k1us (approved):
Code looks good to me. I had some thoughts on config changes and icon needs to updated.

config/default.yaml
ash3rz: Yeah, Paul made a similar comment about configs in another PR. It seems like we need to make a decision on how we want to handle them going forward.  It might be good to have @slr71 weigh in since she's been handling all the external deployments.

ash3rz:
> Can we hard-code/ use some constants to get that icon in place for now until we get custom app icon feature implemented ?

Sure, we can hard-code it for now. I checked MUI's icons and didn't see anything that looked like a terminal but I saw you asked Mariah to make us one 👍

psarando:
> > Can we hard-code/ use some constants to get that icon in place for now until we get custom app icon feature implemented ?
> 
> Sure, we can hard-code it for now. I checked MUI's icons and didn't see anything that looked like a terminal but I saw you asked Mariah to make us one 👍

In the meantime, maybe another thing to try would be an `Avatar` with just some `>_` text?
```jsx
<Avatar><Typography variant="button">{">_"}</Typography></Avatar>
```

ash3rz:
Here's what I've got with Paul's idea:
image
on hover:
image

psarando:
> Here's what I've got with Paul's idea...

I was imagining the background color of the Avatar would be the lighter color and the text would be the darker color, like the `?` Help icon.

I guess that might require changing the color/backgroundColor CSS of the `Avatar`.

config/default.yaml
psarando: I agree that these metadata attrs and values seem like they could be just integral to the DE system, and I'm not sure they would need to change between environments or deployments, like our system-use-only `ipc_*` iRODS metadata.

That said, maybe the metadata attributes could have a more specific name, in order to prevent any potential conflicts with future user defined metadata (like the `cyverse-community` attr above). So maybe `cyverse_de_ui_location`, for example?

In general I think configs are at least conceptually useful, and we should put values in configs that can (or should) change between environments or deployments, even if it requires a rebuild and redeploy for now. I just wish it were easier to change a config value in an environment and rolling-restart the ui service to pick up the new value, without having to push the change to a git repo and rebuild everything (especially if it's supposed to be just a temporary change, like disabling a feature for a little while).

psarando: > That said, maybe the metadata attributes could have a more specific name, in order to prevent any potential conflicts with future user defined metadata (like the `cyverse-community` attr above). So maybe `cyverse_de_ui_location`, for example?
> 

Oh, I just noticed `ui_location` was already in use by older code, so scratch that suggestion 😅

src/components/layout/AppBar.js
```diff
@@ -239,6 +240,8 @@ function DEAppBar(props) {
         cacheTime: Infinity,
     });
 
+    const { instantLaunches } = useFirstClassInstantLaunch();
```
psarando, line 243: 👍

src/components/instantlaunches/admin/InstantLaunchList.js
```diff
@@ -149,7 +149,7 @@ const InstantLaunchList = ({ showErrorAnnouncer }) => {
     const allILs = useQuery(ALL_INSTANT_LAUNCHES_KEY, listFullInstantLaunches);
     const dashboardILs = useQuery(
         [DASHBOARD_INSTANT_LAUNCHES_KEY, "ui_location", "dashboard"],
-        () => listInstantLaunchesByMetadata("ui_location", "dashboard")
+        () => adminListInstantLaunchesByMetadata("ui_location", "dashboard")
     );
```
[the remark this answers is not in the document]
psarando: or a constant, if we go with that instead...

psarando (approved):
LGTM 👍 ✨

ash3rz:
How about this?
image
There's also slightly rounded too:
image

config/default.yaml
slr71: Hmm, I’m not sure I see a reason to make that configurable. I suppose that language differences might mean that users from other countries would want to user terms that make sense to them, but probably only if they had to set the attributes directly. Since we’re planning to build a UI for it, I would think we could use constants instead.

psarando:
> There's also slightly rounded too

Nice, the rounded version LGTM 👍

config/default.yaml
ash3rz: Cool - I'll refactor to use constants then.  Thanks for the input everybody!

ash3rz:
OK, I think I've got all the changes in.  The terminal icon is in and I refactored to use constants.  I also added a sign in dialog for when users click on the `cli` instant launch when not signed in.

ash3rz:
Here's Mariah's version with an svg: 
image

psarando:
> Here's Mariah's version with an svg

Also LGTM 👍

ash3rz:
Thanks for the reviews!

PR #421, opened 2021-10-01
CORE-1484 Allow admins to edit instant launches in nav drawer

ash3rz:
I don't think we're anticipating admins wanting to add more instant launches to the navigation drawer right away, so this doesn't need to be reviewed or merged prior to the release.

Adds the new `Add to / Remove from Navigation Drawer` button for each instant launch:
image


public/static/locales/en/instantlaunches.json
```diff
@@ -1,10 +1,13 @@
 {
     "addedOn": "Added On",
     "addToDashboard": "Add to Dashboard",
+    "addToNavDrawer": "Add to Navigation Drawer",
     "addedToDashboard": "Instant launch added to the dashboard.",
+    "addedToNavDrawer": "Instant launch added to the navigation drawer.",
     "addMapping": "Add Mapping",
     "addToMapping": "Add Pattern Mapping",
     "cannotAddILToDashboard": "Cannot add IL to the dashboard, a required input parameter is present.",
+    "cannotAddILToNavDrawer": "Cannot add IL to the navigation drawer, a required input parameter is present.",
```
sri2k1us, line 10: Lets avoid acronyms and try to be explicit.

src/components/instantlaunches/admin/InstantLaunchList.js
```diff
@@ -209,8 +271,10 @@ const InstantLaunchList = ({ showErrorAnnouncer }) => {
         },
     });
 
-    const isLoading = allILs.isLoading || dashboardILs.isLoading;
-    const isError = allILs.isError || dashboardILs.isError;
+    const isLoading =
+        allILs.isLoading || dashboardILs.isLoading || navDrawerILs.isLoading;
```
sri2k1us, line 275: Should we also check for `removeFromDrawerMutation` and `addToNavDrawerMutation` status ?

src/serviceFacades/instantlaunches.js
```diff
@@ -221,6 +244,31 @@ export const removeFromDashboardHandler = async (id) => {
...
+            value !== instantLaunchNavDrawer
+    );
+
+    if (filtered.length > 0) {
+        return await resetInstantLaunchMetadata(id, filtered);
+    }
+
+    return new Promise((resolve, reject) => resolve(ilMeta));
+};
+
```
sri2k1us, line 271: While this works, I wonder if we can simplify (and standardize) by using react-queries - https://react-query.tanstack.com/guides/dependent-queries#_top
Same for `addToNavDrawer` function as well.

src/components/instantlaunches/admin/InstantLaunchList.js
ash3rz: Yes - good catch. There were others missing too.  I'll add those.

src/serviceFacades/instantlaunches.js
ash3rz: I agree - I saw some opportunities to refactor for consistency with what we're doing in other components, but I also wanted to keep the changes short for this PR.  I can definitely do some refactoring in a separate PR.

PR #423, opened 2021-10-08
Core 1527 Allows users to terminate analysis from Dashboard Analysis section

sri2k1us:
image

image

PR #424, opened 2021-10-08
Core 1528 Allow users to connect to running vice analysis from dashboard

sri2k1us:
- On Top #423 

image


src/components/dashboard/dashboardItem/AnalysisItem.js
```diff
@@ -113,6 +124,22 @@ class AnalysisItem extends ItemBase {
                     </IconButton>
                 </ItemAction>
             ),
+            isVICE && !isTerminatedAnalysis && (
+                <ItemAction
+                    key={`${constants.KIND_ANALYSES}-${props.content.id}-vice`}
+                    tooltipKey="viceAction"
```
ash3rz, line 130: I don't think this tool tip key has been added to the dashboard i18n json file

PR #429, opened 2021-10-21
CORE-1544 Allow admins to set higher app launch resource requirements

psarando:
This PR will update the App Launch Resource Requirements form to allow users to select a higher limit in a requirements request list if a tool admin has set a min value higher than the max default.

### DIFF
--- a/src/components/apps/launch/ResourceRequirements.js
+++ b/src/components/apps/launch/ResourceRequirements.js
@@ -55,8 +55,11 @@ function buildLimitList(startValue, minValue, maxValue) {
         value *= 2;
     }
 
-    if (limits[limits.length - 1] !== maxValue) {
-        limits.push(maxValue);
+    // Finally check the case where either maxValue is not a power of 2,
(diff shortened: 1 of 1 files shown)


ianmcorvidae (approved):
looks good to me! I guess before, the maximum values for the limit lists were getting set too low based on the default maximums?

psarando:
> looks good to me! I guess before, the maximum values for the limit lists were getting set too low based on the default maximums?

Right. For example, even if an admin set the `min_disk_space` requirement to 1TiB for an app's tool, users would not be able to request more than 512GiB minimum disk space when launching that app, since that is the default max setting for disk space, and the `buildLimitList` function would stop populating the list once it hit that max, even though the min is higher.

ianmcorvidae (approved):
Still looks reasonable to me. A lot smaller of a change too, I guess!

sri2k1us (approved):
:shipit:

psarando:
Thanks for the reviews, all! 👍

PR #430, opened 2021-10-26
Core-1555 Analysis Landing page.

sri2k1us:
- Redirect users to the new landing page after an app is successfully launched.

- Large Screen
image

-iPhone 5

image

- Batch Analysis

image

- Back button on drill down page
image


src/components/analyses/landing/AnalysisSubmissionLanding.js
```diff
@@ -0,0 +1,620 @@
...
+            >
+                {view === BATCH_DRILL_DOWN && <BackButton />}
+                <Grid container spacing={1}>
+                    <Grid item xs={isMobile ? 0 : 6}>
+                        <Grid container>
+                            <Grid item>
+                                <Typography variant="h6" color="primary">
+                                    {analysis?.name}
+                                </Typography>
+                                <AnalysisSubheader analysis={analysis} date={formatDate(analysis?.startdate)} />
```
ash3rz, line 356: Might just be me, but when I was looking at the screenshots to see the running time, I initially couldn't find it.  I think I was expecting it to be part of the lower grid / form layout.  Not a blocker obviously, just something I thought I'd point out from my experience.

src/components/data/DataPathLink.js
```diff
@@ -0,0 +1,82 @@
...
+    const { isFetching: isFetchingStat } = useDataDetails({
+        paths: [path],
+        enabled: !!path,
+        onSuccess: () => {
+            setStatError(null);
+            setOtherError(null);
+        },
+        onError: (error) => {
+            if (ERROR_CODES.ERR_DOES_NOT_EXIST === getErrorCode(error)) {
+                setStatError(t("errorInputDoesNotExist", { path: linkTarget }));
```
ash3rz, line 45: Minor, from the screenshot this path in the error message seems to include the prefix `/data/ds`.  I wonder if that might be confusing.  It's sort of a blend between the DE URL and the path in iRods, but not completely either.

src/components/analyses/landing/AnalysisSubmissionLanding.js
```diff
@@ -0,0 +1,620 @@
...
+                            baseId,
+                            ids.STATUS_HISTORY_PANEL
+                        )}
+                        id={buildID(baseId, ids.STATUS_HISTORY_PANEL_HEADER)}
+                    >
+                        <Typography variant="subtitle2" color="primary">
+                            {t("statusHistory")}
+                        </Typography>
+                    </AccordionSummary>
+                    <div style={{ padding: theme.spacing(1) }}>
```
psarando, line 525: I'm guessing `AccordionDetails` already sets some appropriate `padding`.
Also, should this be the panel with the `STATUS_HISTORY_PANEL` ID?
```suggestion
                    <AccordionDetails id={buildID(baseId, ids.STATUS_HISTORY_PANEL)}>
```

src/components/analyses/landing/AnalysisSubmissionLanding.js
```diff
@@ -0,0 +1,620 @@
...
+                            info={history}
+                            isInfoFetching={isInfoFetching}
+                            infoFetchError={infoFetchError}
+                            baseId={baseId}
+                        />
+                    </div>
+                </Accordion>
+                <Accordion
+                    expanded={expandParams}
+                    onChange={() => setExpandParams(!expandParams)}
```
psarando, line 536: No need for us to track this state:
```suggestion
                    defaultExpanded
```

src/components/analyses/landing/AnalysisSubmissionLanding.js
```diff
@@ -0,0 +1,620 @@
...
+                >
+                    <AccordionSummary
+                        expandIcon={<ExpandMore />}
+                        aria-controls={buildID(baseId, ids.PARAMETERS_PANEL)}
+                        id={buildID(baseId, ids.PARAMETERS_PANEL)}
```
psarando, line 541: These IDs should be split up (@ash3rz previously suggested this, but it got lost/outdated), so this `AccordionSummary` should probably have something like `buildID(baseId, ids.PARAMETERS_PANEL_SUMMARY)`, and the panel it controls should get `buildID(baseId, ids.PARAMETERS_PANEL)`.

psarando (approved):
There are some nice refactors in this PR ✨ 
I had some fairly minor suggestions, but overall LGTM 👍

src/components/data/DataPathLink.js
sri2k1us: Agreed. It should be just the path.

src/components/analyses/landing/BatchResults.js
```diff
@@ -0,0 +1,158 @@
...
+                    loading={isFetchingNextPage}
+                    sortable
+                />
+            </AccordionDetails>
+            <AccordionActions>
+                <Button
+                    variant="outlined"
+                    color="primary"
+                    style={{ flex: 1 }}
+                    onClick={() => fetchNextPage()}
```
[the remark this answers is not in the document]
sri2k1us: This won't work. This actually a gotcha! See - https://react-query.tanstack.com/guides/infinite-queries

src/components/analyses/landing/AnalysisSubmissionLanding.js
```diff
@@ -0,0 +1,620 @@
...
+    }
+
+    return (
+        <PageWrapper appBarHeight={75}>
+            <div
+                style={{ margin: theme.spacing(1), padding: theme.spacing(1) }}
+            >
+                {view === BATCH_DRILL_DOWN && <BackButton />}
+                <Grid container spacing={1}>
+                    <Grid item xs={isMobile ? 0 : 6}>
```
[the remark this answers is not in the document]
sri2k1us: i think it can be some thing  like  `<Grid` item xs={12} sm={6}>`. I got this from - https://v4.mui.com/components/grid/#grid-with-breakpoints

src/components/analyses/landing/BatchResults.js
```diff
@@ -0,0 +1,158 @@
...
+            <AccordionSummary
+                expandIcon={<ExpandMore />}
+                aria-controls={buildID(baseId, ids.BATCH_PANEL)}
+                id={buildID(baseId, ids.BATCH_PANEL_HEADER)}
+            >
+                <Typography variant="subtitle2" color="primary">
+                    {t("batchDetails")}
+                </Typography>
+            </AccordionSummary>
+            <AccordionDetails>
```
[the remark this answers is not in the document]
sri2k1us: Good question but I don't see that in example docs - https://v4.mui.com/components/grid/#grid-with-breakpoints 🤔

src/components/analyses/landing/AnalysisSubmissionLanding.js
sri2k1us: With icons
image

sri2k1us: image

ash3rz: Green for completed makes sense to me and it's probably easier to change the status text color anyway 👍

sri2k1us: Mobile view with updated colors
image

sri2k1us:
Thanks for the reviews.

src/components/analyses/landing/AnalysisSubmissionLanding.js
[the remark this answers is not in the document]
psarando: But the logic here looks like `<Grid item xs={0} sm={6}>` to me, and according to https://v4.mui.com/api/grid/, the `xs` prop defaults to `false`, which I'm guessing is treated the same as `xs={0}`?

So I think `<Grid item sm={6}>` is the same as `<Grid item xs={0} sm={6}>`, which is the same as `<Grid item xs={isMobile ? 0 : 6}>` (I think, since `isMobile = useMediaQuery(theme.breakpoints.down("xs"))`).

src/components/analyses/landing/BatchResults.js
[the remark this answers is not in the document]
psarando: Must be a typo in their example (I assume you meant at https://v4.mui.com/components/accordion/), because it's setup like this in the examples at https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role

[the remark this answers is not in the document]
sri2k1us: Doh. Wrong link.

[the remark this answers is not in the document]
psarando: Oh, I just inspected their example, and it looks like MUI is auto-generating a `div` that wraps the `AccordionDetails` content, and that auto-generated `div` gets the attributes `id="panel1a-content" aria-labelledby="panel1a-header"`.

I guess we only need to manually set the content panel's ID if we're not using an `AccordionDetails` panel.

PR #433, opened 2021-11-02
Notification fixes

ash3rz:
This seems to work 🤞 I tested this with multiple tabs and windows in the same browser as well as multiple browsers. `express-ws` seems to keep its own internal list of connected clients and I'm using that.  I will be curious to see if any problems or delays arise however with many users, each with multiple tabs open.

Change notes copied from my commit messages:

* Having the useEffect depend on userProfile and currentNotification was causing up to 3 connections due to those values in the dependency array changing. 1 when userProfile first gets set, 1 when userProfile is updated with the admin key, and 1 when currentNotification is updated from undefined to null.

* See https://github.com/HenningM/express-ws#wsinstancegetwss
This solves the issue where notifications only show up in the first tab/window.
express-ws keeps an internal list of all the connected websocket clients.  In order to make sure a user who has multiple tabs/windows open receives notifications on each instance, we tag each websocket connection with the user's id, then cycle through the list of clients and send the notification to each client with the same user id tag.


psarando (approved):
LGTM :thumbsup: :tada:

sri2k1us (approved):
:shipit: May be even hot fix it in prod 🤔

ash3rz:
Thanks for the reviews!  Let's hope this works 🤞

PR #434, opened 2021-11-03
Change to using an exclusive and autodeleting queue with a randomized name.

ianmcorvidae:
So I still need to test this to make sure I have it right (in particular, I'm not sure if I need to use promises) but I thought I'd put it up and mark it draft first, and explain what I'm doing; it differs a little from how I described it in Slack anyway.

The extremely short version: this makes it so nothing about notifications shares queues ever, and they clean themselves up.

The slightly more in-depth version: by adding a random UUID to the queue name, each connection gets its own queue with the appropriate bindings. By setting up the queue with the `exclusive` flag, only the connection that created the queue is allowed to use it, and by setting the `autoDelete` flag, the queue is automatically deleted whenever the number of consumers drops from 1 to 0 (which means it won't go away immediately after declared, but will when the lone websocket that uses it closes). Since queues now clean themselves up, we don't need to unbind anything any more. I also added the queue name to the logging, though, because it might be useful or relevant.

### DIFF
--- a/src/server/amqp.js
+++ b/src/server/amqp.js
@@ -7,6 +7,7 @@
 import amqplib from "amqplib";
 import * as config from "./configuration";
 import logger from "./logging";
+import UUID from "uuid/v4";
 
 const NOTIFICATION_ROUTING_KEY = "notification.";
 const NOTIFICATION_QUEUE = "de_notifications.";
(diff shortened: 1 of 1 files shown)


ash3rz (approved):
Based on our convo in Slack, this makes sense and LGTM :shipit: 
thanks for looking into this!

ianmcorvidae:
Did some testing and it looks like this works nicely! New queues with UUIDs in their names appear when I open tabs and disappear when I close them, and there's no reason it shouldn't work across multiple replicas exactly the same. I'll go ahead and  merge this now.

PR #436, opened 2021-11-05
core-1598 Allow ht path list file for folder inputs.

sri2k1us:


src/components/data/SelectionDrawer.js
```diff
@@ -86,12 +87,21 @@ function SelectionToolbar(props) {
...
+            invalidTotal = 1;
+        }
+    } else {
+        invalidTotal =
+            ResourceTypes.ANY !== acceptedType
+                ? selectedResources.filter(
+                      (resource) => resource.type.toLowerCase() !== acceptedType
+                  ).length
+                : 0;
+    }
```
psarando, line 104: I think this could be simplified a bit, if we store some of these boolean checks into variables.
For example
```suggestion
    const allowAny = ResourceTypes.ANY !== acceptedType;
    const allowPathList = ResourceTypes.FOLDER === acceptedType && !multiSelect;
    let invalidTotal = 0;
    if (!allowAny) {
        invalidTotal = selectedResources.filter((resource) =>
                       resource.type.toLowerCase() !== acceptedType
                       && !(allowPathList && resource.infoType === InfoTypes.HT_ANALYSIS_PATH_LIST)
                  ).length;
    }
```

psarando (approved):
The logic looks right to me, though I found it a little tricky to parse at first.
I added a minor style suggestion that might help readability (though I haven't tested it).

:shipit: when ready 👍

src/components/data/SelectionDrawer.js
sri2k1us: thanks the suggestion although I am more inclined to use code I wrote since I spent some time coming up with the logic and testing them.

PR #438, opened 2021-11-09
CORE-1601: made the username suffix configurable

slr71:


johnworth (approved):
I didn't see anything obviously wrong with this. :shipit:

slr71:
Thanks for the review! I'll wait until I can merge the `apps` service changes before merging this one.

sri2k1us (changes requested):
So far so good but I think this needs some more changes. First, for the config to be available on the client side, the config must be read in using [next.config.js](https://github.com/cyverse-de/sonora/blob/master/next.config.js) and set up in [pages/_apps.js](https://github.com/cyverse-de/sonora/blob/625be0bae467f622efaa079efca6f69de210058a/src/pages/_app.js#L206). 
Also I found a typo that breaks the server start-up.

config/default.yaml
```diff
@@ -87,3 +87,5 @@ unleash:
     client_secret: null
     maintenance_flag: DE-Maintenance
 
+usernames:
```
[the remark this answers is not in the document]
slr71: D'oh! Thanks for catching that.

slr71:
I'm a little confused about the requested change for reading the configuration. It was working when I tested it yesterday using `npm run dev` is there another test that I should be doing to make sure that the configuration will always be available? Or were you just saying that there should be a constant available for this setting rather than accessing the configuration object?

sri2k1us:
> I'm a little confused about the requested change for reading the configuration. It was working when I tested it yesterday using `npm run dev` is there another test that I should be doing to make sure that the configuration will always be available? Or were you just saying that there should be a constant available for this setting rather than accessing the configuration object?

I used `nmp run dev` too but I set up the config for suffix to be something like `foo.org` and it didn't show up in the `getUserNameSuffix` function. I can show you what I did in zoom call.

slr71:
> > I'm a little confused about the requested change for reading the configuration. It was working when I tested it yesterday using `npm run dev` is there another test that I should be doing to make sure that the configuration will always be available? Or were you just saying that there should be a constant available for this setting rather than accessing the configuration object?
> 
> I used `nmp run dev` too but I set up the config for suffix to be something like `foo.org` and it didn't show up in the `getUserNameSuffix` function. I can show you what I did in zoom call.

The Zoom call helped. Thanks!

The problem was that I didn't realize that the JSON object for the configuration is being reconstructed after being read in from the configuration file. The reason it was working was that I had the code in place for making the configuration setting optional, and that was in place. I'm a little embarrassed for not thinking about that, but that was what happened.

sri2k1us:
> > > I'm a little confused about the requested change for reading the configuration. It was working when I tested it yesterday using `npm run dev` is there another test that I should be doing to make sure that the configuration will always be available? Or were you just saying that there should be a constant available for this setting rather than accessing the configuration object?
> > 
> > 
> > I used `nmp run dev` too but I set up the config for suffix to be something like `foo.org` and it didn't show up in the `getUserNameSuffix` function. I can show you what I did in zoom call.
> 
> The Zoom call helped. Thanks!
> 
> The problem was that I didn't realize that the JSON object for the configuration is being reconstructed after being read in from the configuration file. The reason it was working was that I had the code in place for making the configuration setting optional, and that was in place. I'm a little embarrassed for not thinking about that, but that was what happened.

No need to be embarrassed. As I said in the call, When we set up nextjs, we had a lot of difficulty figuring this out. Any way, constructing the config object `_app.js`  is one aspect but the main thing is reading the config file in `next.config.js` is step we all forget.  Here is the link to nextjs documentation - https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration

sri2k1us (approved):
LGTM :shipit:

src/components/analyses/listing/TableView.js
```diff
@@ -181,6 +183,9 @@ function TableView(props) {
     const theme = useTheme();
     const { t } = useTranslation("analyses");
 
+    const [config] = useConfig();
+    console.error(`and the config is.... ${config}`);
```
psarando, line 187: Looks like debug logging that can be removed now.

src/components/analyses/toolbar/AnalysesDotMenu.js
```diff
@@ -295,6 +297,9 @@ function AnalysesDotMenu({
     const { baseId, isSingleSelection } = props;
     const { t } = useTranslation("common");
 
+    const [config] = useConfig();
+    console.error(`and the config is... ${config}`);
```
psarando, line 301: Looks like debug logging that can be removed now.

src/components/utils/getUserName.js
```diff
@@ -1,15 +1,32 @@
...
  * @param {string} userNameWithDomain - user name with domain
+ * @param {object} config - the app configuration.
  *
  * @returns {string} - short username.
  */
-export const getUserName = (userNameWithDomain) => {
-    return userNameWithDomain && userNameWithDomain.includes(constants.IPLANT)
+export const getUserName = (userNameWithDomain, config) => {
+    const suffix = getUserNameSuffix(config);
+    return userNameWithDomain && userNameWithDomain.includes(suffix)
```
psarando, line 29: Since we're naming this as a `suffix` explicitly now, I wonder if we should use `endsWith` instead of `includes`?
```suggestion
    return userNameWithDomain && userNameWithDomain.endsWith(suffix)
```

psarando (approved):
LGTM 👍

slr71:
Thanks for the reviews!

PR #439, opened 2021-11-18
Core 1611 Fix Analysis Landing Page infinite loop

sri2k1us:
SSIA although am not entirely sure what was causing this infinite loop of `History` and `stat` service call. When I removed the duplicate check of  `isInfoFetching`, it fixed the issue. 😕

### DIFF
--- a/src/components/analyses/landing/AnalysisSubmissionLanding.js
+++ b/src/components/analyses/landing/AnalysisSubmissionLanding.js
@@ -99,8 +99,8 @@ export default function AnalysisSubmissionLanding(props) {
     const isMobile = useMediaQuery(theme.breakpoints.down("xs"));
     const [analysis, setAnalysis] = React.useState();
     const [helpOpen, setHelpOpen] = React.useState(false);
-    const [history, setHistory] = React.useState();
-    const [parameters, setParameters] = React.useState();
+    const [history, setHistory] = React.useState(null);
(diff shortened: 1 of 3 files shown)


src/components/analyses/landing/AnalysisSubmissionLanding.js
```diff
@@ -132,38 +132,32 @@ export default function AnalysisSubmissionLanding(props) {
...
+                    setAnalysis({
+                        ...analysis,
+                        resultfolderid,
+                        status,
+                        enddate,
+                    });
+                }
+            }
+        }
+    }, [analysis, currentNotification, id]);
```
sri2k1us, line 160: Mostly a refactor cos `React.useCallback` or having the logic inside `useEffect` is the same.

src/serviceFacades/analyses.js
```diff
@@ -283,7 +282,6 @@ function useAnalysisParameters({ id, enabled, onSuccess, onError }) {
         onSuccess: (data) => {
             onSuccess(preProcessData(data));
         },
-        onError,
```
sri2k1us, line 286: remove unused params.

psarando (approved):
LGTM 👍 

Though I don't really spot the fix either; all I can think of is the callback may have been causing issues? 🤷

sri2k1us:
Thanks for the reviews.

PR #440, opened 2021-11-18
core-1615 Fix move destination folder selection.

sri2k1us:
SSIA

### DIFF
--- a/src/components/apps/launch/params/FolderInput.js
+++ b/src/components/apps/launch/params/FolderInput.js
@@ -7,9 +7,14 @@ import React from "react";
 
 import DiskResourceSelector from "./DiskResourceSelector";
 import ResourceTypes from "components/models/ResourceTypes";
+import InfoTypes from "components/models/InfoTypes";
 
 export default function FolderInput(props) {
     return (
-        <DiskResourceSelector acceptedType={ResourceTypes.FOLDER} {...props} />
+        <DiskResourceSelector
+            acceptedType={ResourceTypes.FOLDER}
+            acceptedInfoTypes={[InfoTypes.HT_ANALYSIS_PATH_LIST]}
+            {...props}
+        />
     );
 }
--- a/src/components/apps/launch/params/FolderInput.js
+++ b/src/components/apps/launch/params/FolderInput.js
@@ -10,10 +10,13 @@ import ResourceTypes from "components/models/ResourceTypes";
 import InfoTypes from "components/models/InfoTypes";
 
 export default function FolderInput(props) {
+    const { multiSelect } = props;
     return (
         <DiskResourceSelector
             acceptedType={ResourceTypes.FOLDER}
-            acceptedInfoTypes={[InfoTypes.HT_ANALYSIS_PATH_LIST]}
+            acceptedInfoTypes={
+                !multiSelect ? [InfoTypes.HT_ANALYSIS_PATH_LIST] : null
+            }
             {...props}
         />
     );
--- a/src/components/data/SelectionDrawer.js
+++ b/src/components/data/SelectionDrawer.js
@@ -27,7 +27,6 @@ import { useTranslation } from "react-i18next";
 import ids from "./ids";
 import Listing from "components/data/listing/Listing";
 import ResourceTypes from "../models/ResourceTypes";
-import InfoTypes from "components/models/InfoTypes";
 
 import styles from "./styles";
 import constants from "../../constants";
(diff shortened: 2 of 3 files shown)


src/components/data/SelectionDrawer.js
```diff
@@ -89,17 +89,20 @@ function SelectionToolbar(props) {
 
     let invalidTotal = 0;
     if (ResourceTypes.FOLDER === acceptedType && !multiSelect) {
+        const selectedInfoType = selectedResources[0]?.infoType;
```
ash3rz, line 92: A comment might be good here - when I was first reviewing this, I was curious why we're checking for HT path list files if the user is supposed to select a single folder.

ash3rz (approved):
LGTM :shipit:

psarando (commented):
It looks like these changes fix folder selections in app inputs and move dialogs, but I don't see how it will prevent HT Path selections in move dialogs.

We should not allow HT Path list files as move destinations.

psarando:
I also noticed in QA that the row-style when selecting an HT Path list as a folder input for an app displays with an error color, even though it's an allowed selection:
image

ash3rz:
> It looks like these changes fix folder selections in app inputs and move dialogs, but I don't see how it will prevent HT Path selections in move dialogs.
> 
> We should not allow HT Path list files as move destinations.

Ah, I didn't realize we were also having that as an issue.  One thing we could maybe do is modify the `acceptedType` prop in the selection drawer to be a list of resource types (file/folder/any) and/or info types. That way the parent component that's rendering the drawer can decide if in that particular scenario it's appropriate to allow folders  + path lists or not.

src/components/data/SelectionDrawer.js
```diff
@@ -41,6 +40,64 @@ const useStyles = makeStyles(styles);
...
+                    }
+                } else {
+                    // in other places like Move or Preferences default output folder selection
+                    // we only allow folders
+                    if (selectedType !== ResourceTypes.FOLDER) {
+                        invalidTotal = 1;
+                    }
+                }
+            } else {
+                //we don't support info type based selections for multi-select folders
```
ash3rz, line 88: I think this works, but I also think it makes sense to have the responsibility be on the parent component to not provide a list of acceptable info types if they aren't actually acceptable.  

If that were the case, then I think this `switch` could be replaced with something like:
```
invalidTotal = acceptedType !== ResourceTypes.ANY 
    ? selectedResources.filter((resource) => resource.type.toLowerCase() !== acceptedType && !acceptedInfoTypes.includes(resource.infoType)) 
    : 0;
```

src/components/apps/launch/params/FolderInput.js
```diff
@@ -7,9 +7,14 @@ import React from "react";
...
 import ResourceTypes from "components/models/ResourceTypes";
+import InfoTypes from "components/models/InfoTypes";
 
 export default function FolderInput(props) {
     return (
-        <DiskResourceSelector acceptedType={ResourceTypes.FOLDER} {...props} />
+        <DiskResourceSelector
+            acceptedType={ResourceTypes.FOLDER}
+            acceptedInfoTypes={[InfoTypes.HT_ANALYSIS_PATH_LIST]}
+            {...props}
```
ash3rz, line 17: This might cover setting the `acceptedInfoTypes` correctly, I'm not 100% sure and it would need to be tested:

```
export default function FolderInput(props) {
    const {multiSelect, ...rest} = props;
    return (
        <DiskResourceSelector
            acceptedType={ResourceTypes.FOLDER}
            acceptedInfoTypes={!multiSelect ? [InfoTypes.HT_ANALYSIS_PATH_LIST] : null}
            multiSelect={multiSelect}
            {...rest}
```

src/components/data/SelectionDrawer.js
psarando: This also makes sense to me.

I would also prefer to check each item in the selected resources for both `acceptedType` and `acceptedInfoTypes`, rather than checking a count of invalid InfoTypes separately from accepted Types; otherwise I think there's a risk that the invalid selection count might be off.

Suppose in the future we allow app integrators to restrict file inputs by InfoType. If the user selects both a folder and a file with the wrong InfoType, then I think the logic in the `case ResourceTypes.FILE` will only report 1 invalid selection.

Putting the checks on each item in a single `filter` loop should correct that.

src/components/apps/launch/params/FolderInput.js
psarando: Looks good to me, although there is no need to keep `multiSelect` separate from the rest of the `props` since it still needs to be passed into the `DiskResourceSelector`.
```jsx
export default function FolderInput(props) {
    const { multiSelect } = props;
    return (
        <DiskResourceSelector
            acceptedType={ResourceTypes.FOLDER}
            acceptedInfoTypes={!multiSelect ? [InfoTypes.HT_ANALYSIS_PATH_LIST] : null}
            {...props}
```

src/components/data/SelectionDrawer.js
sri2k1us: Actually we will still need to do a `switch` even with this suggested change:
- for the case described by @psarando, we must check if the selected item is of type `FILE` `and` it conforms to the accepted infoType array if provided.
- for a `FOLDER` input type,  we must check if the selected item is of type `FOLDER` `or` it conforms to the accepted infoType array if provided. 
Beyond this, the filtering logic must still account of `null` or `undefined` `acceptedInfoTypes`.

Let me know if this makes sense.

psarando (approved):
LGTM 👍

sri2k1us:
Thanks for the reviews.

PR #442, opened 2021-12-04
Refine the ticks on the data consumption graph display

ianmcorvidae:
Since I spent a chunk of the day thinking about it while off doing appointments and errands, I think this works out pretty nicely!

Here's how it works:
1. we set a max number of ticks (6, which seems like a nice enough amount for this small display).
2. we turn the quota into a number of GiB for comparison
3. we attempt to find a nice divisor of the number of GiB in the quota. We try this in an order that hopefully means we get nicer numbers of ticks in the most common display, when a user is under quota -- 4, then 3, then 5, then 2. If nothing works (which is pretty unlikely, but we needed some sort of default), use 4. This divisor is the number of ticks up to the quota value (so if it's 4, the graph will show a tick at 0, at 1/4 of quota, 1/2 of quota, 3/4 of quota, and at quota).
4. The step size is then set based on the divisor, quota, and usage: if setting a step size of `quota / divisor` would result in `maxTicks` (6) or fewer ticks, set the step size to that. Otherwise, set it to increments of the quota.
5. Set the maximum extent of the graph to correspond to an increment of the stepSize, but at the minimum, it should be set to the quota.

This process should work pretty well even with changing/dynamic quotas later on. The only thing that might be weird is if we get any quotas that aren't divisible by 2/3/4/5 GiB, but unless we start doing special sales if you buy prime-numbered data storage quota or something we should be fine. With the current 100GiB quota, it means we get 25GiB ticks when under quota, and also up through 150GiB of usage, and then 100GiB ticks afterwards.

### DIFF
--- a/src/components/dashboard/dashboardItem/DataConsumption.js
+++ b/src/components/dashboard/dashboardItem/DataConsumption.js
@@ -44,6 +44,22 @@ ChartJS.register(
 );
 
 const options = (usage, quota, date, distance, title, theme, t) => {
+    let quotaGiB = quota / constants.ONE_GiB;
+    let divisor =
+        quotaGiB % 4 === 0
+            ? 4
+            : quotaGiB % 3 === 0
+            ? 3
+            : quotaGiB % 5 === 0
+            ? 5
+            : quotaGiB % 2 === 0
+            ? 2
+            : 4;
+    let maxTicks = 6;
+    let stepSize =
+        usage <= (quota / divisor) * maxTicks
+            ? Math.ceil(quota / divisor)
+            : quota;
     return {
         indexAxis: "y",
         plugins: {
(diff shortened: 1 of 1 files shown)


src/components/dashboard/dashboardItem/DataConsumption.js
```diff
@@ -44,6 +44,22 @@ ChartJS.register(
...
+            ? 3
+            : quotaGiB % 5 === 0
+            ? 5
+            : quotaGiB % 2 === 0
+            ? 2
+            : 4;
+    let maxTicks = 6;
+    let stepSize =
+        usage <= (quota / divisor) * maxTicks
+            ? Math.ceil(quota / divisor)
```
ianmcorvidae, line 61: I'm not 100% sure this `Math.ceil` is the right thing to do. But, it'll only come up if the `divisor` calculation falls through to the default value of 4, which only happens if the quota's prime factorization doesn't include anything below 7, and it seems unlikely that'd happen (in fact, I'd say it's pretty unlikely we'd do any that aren't multiples of 100, which is already a multiple of 2,4, and 5).

ianmcorvidae: well, or if the quota's not a multiple of a GiB, but that seems even less likely

sri2k1us:
Worked mostly well except when usage gets into `TiB`.

sri2k1us (approved):

ash3rz (approved):
LGTM :shipit:

PR #447, opened 2021-12-10
CORE-1609 Advanced Data Search Form

ash3rz:
The advanced search form is accessible from the Data page:
image

And the data search tab:
image

Here's the empty form:
image

Here's a form with everything filled out:
image

Sample mobile look:
image


src/components/forms/FormTimestampField.js
```diff
@@ -15,18 +15,26 @@ import TextField from "@material-ui/core/TextField";
...
-    const dateStr = formatDateObject(date, dateConstants.DATE_FORMAT);
+    if (newValue) {
+        const date = prevDate ? prevDate : new Date();
+        const dateStr = formatDateObject(date, dateConstants.DATE_FORMAT);
 
-    setFieldValue(fieldName, `${dateStr} ${newValue}`);
+        setFieldValue(fieldName, `${dateStr} ${newValue}`);
+    } else {
+        setFieldValue(fieldName, "");
+    }
```
ash3rz, line 37: These changes made sense to me, but I wanted to point it out since we've had this component for a while and it hasn't been an issue. 
The issue I was facing was basically that if you filled in a value, changed your mind and cleared the value, part of the value would still remain and would be present in the `values` from Formik.

sri2k1us:
- filters not updated: 
image

- Subject search for logged out users:
image
Either we should hide this and permission field for logged out users search or make the error message clear that they should log in to use this search field.

- Units different from listing:
image

sri2k1us:
Also I am not able to browse community data when logged out. Prod / QA seems to work fine:
image

ash3rz:
> Also I am not able to browse community data when logged out. Prod / QA seems to work fine: image

🤔 it's working for me in my branch: 
image

sri2k1us:
> Also I am not able to browse community data when logged out. Prod / QA seems to work fine: image

Oops. Bad config. Sorry about the confusion.

public/static/locales/en/search.json
```diff
@@ -1,16 +1,40 @@
...
     "appsSearchTab": "Apps ({{count}})",
+    "attribute": "Attribute",
+    "created": "Created",
     "dataSearchTab": "Data ({{count}})",
-    "teamSearchTab": "Teams ({{count}})",
+    "endDate": "End Date",
     "errorSearch": "Unable to search. Please try again.",
-    "noResults": "No results to display.",
-    "matchingResults": "matching results.",
+    "fileNamePlaceholder": "myAnalysisFile",
```
sri2k1us, line 10: May be include a tip saying `glob` patterns are supported.

src/components/search/form/clauses/FileSize.js
```diff
@@ -0,0 +1,124 @@
...
+const sizesList = ["KiB", "MiB", "GiB", "TiB"];
+
+function SizeUnit(props) {
+    const {
+        field,
+        form: { setFieldValue, ...form },
+        ...rest
+    } = props;
+    return (
+        <FormSelectField form={form} field={field} fullWidth={false} {...rest}>
```
psarando, line 111: I don't see why `setFieldValue` should be removed from the `FormSelectField` props.

Alternatively, rather than using a `FormSelectField`, you could also use a `FormTextField` here with the `select` prop set (I'd also suggest setting `variant="outlined"` too).

For example:
https://github.com/cyverse-de/sonora/blob/45d876316a3bf625f3a832cb61f7cea5f8e73b23/src/components/apps/launch/params/Selection.js#L12-L23

psarando (approved):
LGTM 👍 ✨

src/pages/search.js
```diff
@@ -28,12 +29,13 @@ export default function Search() {
...
                 baseId="search"
-                searchTerm={router.query?.searchTerm}
+                searchTerm={searchTerm}
+                advancedDataQuery={advancedDataQuery}
                 filter={filter}
                 selectedTab={tab}
                 onTabSelectionChange={(event, selection) => {
                     router.push(
-                        `/${NavigationConstants.SEARCH}?searchTerm=${searchTerm}&filter=${filter}&selectedTab=${selection}`
+                        `/${NavigationConstants.SEARCH}?searchTerm=${searchTerm}&advancedDataQuery=${advancedDataQuery}&filter=${filter}&selectedTab=${selection}`
```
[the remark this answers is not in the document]
ash3rz: Thanks for the reminder - I did forget this was an option.  I'll refactor

src/components/search/form/clauses/FileSize.js
sri2k1us: While this works 
image

It will nice to have something like what google drive does for Advanced search / filter:

image


May be to the left of `Advanced Data Search` button 🤔

sri2k1us:
While this works:
image

It will be nice to list the filters like how google drive does:
image


May be to the left of the `Advanced Data Search` button 🤔

sri2k1us (approved):

ash3rz:
> While this works for now: image
> 
> It will be nice to list the filters like how google drive does: image
> 
> May be to the left of the `Advanced Data Search` button 🤔

I got something basic working for now.  Since the queries can end up pretty lengthy, I made the text a bit smaller to start off with:
image

ash3rz:
Thanks for the reviews!

ianmcorvidae:
So uh... there's actually something in querydsl that's designed for summarizing queries (it's used for logging in the search service but at present nothing else). By way of example,: `All:[label~"test",path="/cyverse/home/aramsey"]`. I'm not sure if it's an ideal format or anything, but is this something that we should expose via an endpoint?

ianmcorvidae:
(noting that this might be for later and somebody else, but anyway)

ash3rz:
> So uh... there's actually something in querydsl that's designed for summarizing queries (it's used for logging in the search service but at present nothing else). By way of example,: `All:[label~"test",path="/cyverse/home/aramsey"]`. I'm not sure if it's an ideal format or anything, but is this something that we should expose via an endpoint?

I had no idea!  That sounds like a great idea to me.

src/components/search/detailed/DataSearchResults.js
```diff
@@ -380,4 +370,38 @@ function DataSearchResults(props) {
...
+        return clauses
+            ?.map((clause) => {
+                const args = clause.args;
+                return Object.entries(args).map(([key, value]) => {
+                    return `${key}:${value}`;
+                });
+            })
+            .flat()
+            .join(" ");
+    };
```
psarando, line 390: 👍

PR #448, opened 2022-01-10
Add user transition survey to dashboard.

sri2k1us:
SSIA


src/components/dashboard/constants.js
```diff
@@ -44,3 +44,5 @@ export const SM_NUM_COLUMNS = 2;
 export const SM_PIXELS = 600;
 export const XS_NUM_COLUMNS = 1;
 export const XS_PIXELS = 0;
+
+export const USER_SURVEY = "https://cyver.se/srvy";
```
ianmcorvidae, line 48: Is there a particular reason that we're using the short-url here instead of the fully-expanded link? (I guess there might be some sort of analytics being done via the short-url provider?). If there isn't one, I'd rather we had the full link, most likely, but I don't feel strongly about it if you'd prefer not.

src/components/dashboard/index.js
```diff
@@ -256,6 +257,7 @@ const Dashboard = (props) => {
     return (
         <div ref={dashboardEl} id={baseId} className={classes.gridRoot}>
             {!userProfile?.id && <Banner />}
+            {<UserSurvey />}
```
ianmcorvidae, line 260: If it's much of a bother then don't, but if it's easy, it would be nice to let this card be hidden by users if they've already done the survey.

ianmcorvidae (approved):
usual not-much-of-a-UI-dev caveats but looks reasonable to me, a couple comments that you can take or leave

slr71 (approved):
I didn't catch any issues, but my UI code reviews should always be taken with a grain of salt. 😄

src/components/dashboard/constants.js
sri2k1us: The survey is in google forms and has a pretty long url. So we shortened it.

src/components/dashboard/index.js
sri2k1us: That will be nice but this there is not an easy way to track if the user has taken the survey and I don't users to simply hide and not take it.

src/components/dashboard/constants.js
ianmcorvidae: I guess I was figuring the length of the URL is pretty irrelevant if you're just clicking a button to go to it, especially as we have it here where it won't even show in the browser on hover (pretty sure). It's a short-term thing so there's not much risk of the shortlink disappearing, but it's still an extra DNS lookup and redirect for no practical reason I can think of.

But it's still a short-term/temporary thing so I'm not that bothered by it. It just seems a little untidy, haha.

sri2k1us: True that links won't show up on hover. Now that I think about it, we will also be sending this link out to users in email and intercom messages.

src/components/dashboard/index.js
psarando, line 260:
```suggestion
            <UserSurvey />
```

src/components/dashboard/constants.js
ianmcorvidae: Yeah -- I don't know if we're using the rebrandly analytics (appears to be what the cyver.se domain is pointing to). If we are, definitely keep the short link, in any case, and it's worth sending it out in email/intercom where the link will actually appear for sure.

psarando (approved):
LGTM 👍

sri2k1us:
Updated header and button color:

image

sri2k1us:
Thanks for the reviews.

PR #451, opened 2022-01-21
core-1666 Add max cpu resources to advanced settings in app launch.

sri2k1us:
- Validation

image

- Review
image


sri2k1us:
Known issue - When relaunching, the Max cores are not set cos am not getting that back in relaunch config from the service.

src/components/apps/launch/ResourceRequirements.js
```diff
@@ -116,52 +116,84 @@ const StepResourceRequirementsForm = ({
...
+                    <Typography variant="subtitle2">
+                        {t("selectMaxs")}
+                    </Typography>
+                </Grid>
+                <Grid item xs={12}>
+                    <FastField
+                        id={buildID(baseId, "MAX_CPU")}
+                        name={`requirements.${index}.max_cpu_cores`}
+                        label={t("maxCPUCores")}
+                        component={FormSelectField}
```
psarando, line 186: Should we add some `helperText` to this field explaining to the user that the max setting here is what will count as their actual usage for this analysis?

public/static/locales/en/launch.json
```diff
@@ -38,6 +39,8 @@
     "resourceRequirementsForStep": "Resource Requirements for Step {{step}}",
     "reviewAndLaunch": "Review and Launch",
     "saveAsSavedLaunch": "Create Saved Launch",
+    "selectMins": "Select Minimums",
+    "selectMaxs": "Select Maximums",
```
psarando, line 43:
```suggestion
    "selectMaxes": "Select Maximums",
```

src/components/apps/launch/ResourceRequirements.js
```diff
@@ -116,52 +116,84 @@ const StepResourceRequirementsForm = ({
...
-                        <MenuItem key={index} value={size}>
-                            {formatGBListItem(size)}
-                        </MenuItem>
-                    ))}
-                </FastField>
+            <br />
+            <Grid container spacing={1}>
+                <Grid item xs={12}>
+                    <Typography variant="subtitle2">
+                        {t("selectMaxs")}
```
psarando, line 178:
```suggestion
                        {t("selectMaxes")}
```

psarando (changes requested):
So far so good 👍 

I think we need to set the initial value of `max_cpu_cores` here:
https://github.com/cyverse-de/sonora/blob/c87af1639671390b52ddfae38c1e8458039042ce/src/components/apps/launch/formatters.js#L38-L50

The actual max value should be pre-selected, so the user knows just how much CPU this analysis will use. That should be the `max_cpu_cores` field in the original requirements JSON from the service (destructured in that map callback).

I know you already mentioned that the user setting is not returned in the relaunch JSON, but once it's added in the service, that should be used before the actual max.

If the user setting from relaunch is named `default_max_cpu_cores`, then this function should look something like this:
```javascript
    const reqInitValues = requirements?.map(
        ({
            step_number,
            max_cpu_cores,
            default_max_cpu_cores = 0,
            default_cpu_cores = 0,
            default_memory = 0,
            default_disk_space = 0,
        }) => ({
            step_number,
            max_cpu_cores: default_max_cpu_cores || max_cpu_cores,
            min_cpu_cores: default_cpu_cores,
            min_memory_limit: default_memory,
            min_disk_space: default_disk_space,
        })
    );
```

src/components/apps/launch/ResourceRequirements.js
psarando: Alternatively, since we currently have a large block of text above the minimum select lists explaining what effect these settings have on submission wait times, then we could have another block of text above the max settings explaining the effect of the max CPU setting.

sri2k1us: @psarando  - Do you have any suggestions for the text ? I turning up blank 😞

psarando: We should get input from @johnworth and 411?

psarando (approved):
LGTM 👍

sri2k1us:
Thanks for the reviews.

PR #455, opened 2022-02-04
Core 1671 Add CPU consumption chart

sri2k1us:
image

image

image


src/components/dashboard/dashboardItem/CPUConsumption.js
```diff
@@ -0,0 +1,196 @@
...
+                padding: 1,
+            },
+        },
+        responsive: false,
+        scales: {
+            x: {
+                barThickness: "flex",
+                stacked: false,
+                min: 0,
+                max: Math.max(Math.ceil(usage / stepSize) * stepSize, quota),
```
psarando, line 103: This logic, plus all the logic for calculating `divisor`, `maxTicks`, and `stepSize` is just a copy of the logic from the `DataConsumption` component:
https://github.com/cyverse-de/sonora/blob/1fc3baf1611485cc0e07508668b5793d1f691014/src/components/dashboard/dashboardItem/DataConsumption.js#L48-L63

I think this is complicated enough that it deserves to be refactored into its own util function, which only needs `usage` and `quota` params, and it can return this `Math.max` calculation.

psarando (commented):
Looks pretty straight forward, except I suggest refactoring the complicated chart-ticks calculations into a common util function.

src/components/dashboard/dashboardItem/CPUConsumption.js
psarando: Oh wait, I didn't notice before that `DataConsumption` is using `quotaGiB` for `divisor`, but just `quota` for `stepSize` 🤔 

In that case, we can pass on trying to refactor this logic in this PR. We can revisit it later, especially if we have another chart to add in the future that will need this logic as well.

psarando (approved):

src/components/dashboard/dashboardItem/CPUConsumption.js
ianmcorvidae: Yeah, if we wanted to refactor it out, what we'd probably want to do is:

 * instead of `constants.ONE_GiB`, pass in a unit size as an argument (for CPUs, it might just be 1, but for data it's 1 GiB)
 * calculate divisor using `quota/unitSize`
 * calculate step size using divisor & max ticks

ianmcorvidae: I guess if we really wanted to get fancy we could also pass in divisor preferences (list of prefs + default, rather than the implicit `[4, 3, 5, 2], 4` we have here), haha

sri2k1us:
Thanks for the review.

PR #456, opened 2022-02-10
core-1679 Add undefined check before trying to update app search resu…

sri2k1us:
…lt count.


psarando (approved):
:shipit: 👍

sri2k1us:
Thanks for the review.

PR #462, opened 2022-03-17
CORE-1721 Fix new tool submissions in EditTool form

psarando:
This PR will fix a new tool submission bug introduced in #454, and also includes some formatting updates from prettier.

### DIFF
--- a/src/components/tools/edit/EditTool.js
+++ b/src/components/tools/edit/EditTool.js
@@ -58,7 +58,7 @@ import {
     Paper,
     Typography,
 } from "@material-ui/core";
-import LabelIcon from '@material-ui/icons/Label';
+import LabelIcon from "@material-ui/icons/Label";
 import { withStyles } from "@material-ui/core/styles";
 import { Skeleton } from "@material-ui/lab";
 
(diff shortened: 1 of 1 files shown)


src/components/tools/edit/EditTool.js
```diff
@@ -133,8 +133,8 @@ function EditToolDialog(props) {
     });
 
     const { mutate: addNewTool, status: newToolStatus } = useMutation(
-        ({ submission }) =>
-            isAdmin ? adminAddTool(submission) : addTool(submission),
+        () =>
+            isAdmin ? adminAddTool(toolSubmission) : addTool(toolSubmission),
```
psarando, line 137: The submission values were moved into state in this file's last PR.

ianmcorvidae:
As far as I can tell this looks good -- I guess the bit you commented on is the fix for the bug?

psarando:
> I guess the bit you commented on is the fix for the bug?

Yep. I also kept it in its own commit, separate from the formatting commit, if you want to look at it that way too 👍

slr71 (approved):
Looks good to me! :shipit:

psarando:
Thanks for the reviews! 👍

PR #464, opened 2022-04-01
CORE-1687 Replace Subscription Buy URL with config value

psarando:
This PR will replace the subscription `Buy` button's URL with a config value instead of using a constant.


johnworth (approved):
LGTM :shipit: 

Thanks for working on this.

stories/configMock.js
```diff
@@ -42,10 +42,8 @@ export default {
         defaultCasUrl: "https://olson.cyverse.org/cas",
         defaultCasValidate: "validate",
     },
-    metadata: {
-        communityAttr: "cyverse-community",
-        instantLaunchLocationAttr: "ui_location",
-        instantLaunchDashboard: "dashboard",
-        instantLaunchNavDrawer: "nav_drawer",
```
ianmcorvidae, line 49: Hopefully removing these from the configMock is fine? doesn't seem like they moved somewhere else

ianmcorvidae (approved):
Looks basically reasonable to me! One small question, as I usually end up doing, heh.

stories/configMock.js
psarando: Yep, I happened to notice that the actual configs were removed in a previous PR, so I thought I'd clean these up while I was making changes here.

psarando:
Thanks for the reviews! 👍

PR #465, opened 2022-04-04
Fix image viewer by adding a wrapping div that doesn't impose flexbox on its children

ianmcorvidae:
Small fix, discussed it a little in slack the other month. Before/after:

flexboxfix-nodiv
flexboxfix-div


psarando (approved):
LGTM 👍 🎉 
Also, cute example pic 😹

ianmcorvidae:
Haha cool, thanks for the review. It was the one image I'd uploaded to QA already :joy_cat:

PR #466, opened 2022-04-04
Add opentelemetry tracing, exporting to Jaeger over HTTP

ianmcorvidae:
I'm not sure if there's a way to make this work with the dev server. Maybe we can just add the same `-r` argument there, though?

Anyway putting it up for review because I'm less comfortable in JS, but this did work locally with port-forwarded jaeger-collector and the appropriate env variables set with `npm run start`. I'm just not sure about other ways of running it, or if there's anything else I should look at.


package.json
```diff
@@ -5,12 +5,12 @@
...
     "scripts": {
         "dev": "nodemon",
-        "dev-server": "cross-env NODE_ENV=development babel-node ./src/server/",
+        "dev-server": "cross-env NODE_ENV=development babel-node -r './src/server/tracing.js' ./src/server/",
         "build-frontend": "next build",
         "build-server": "babel ./src/i18n.js ./src/constants.js --out-dir dist/ && babel ./src/server --out-dir dist/server && babel ./src/common --out-dir dist/common",
         "build-docs": "jsdoc -r -d generated-docs ./src",
         "build": "run-p build-frontend build-server build-docs",
-        "start": "cross-env NODE_ENV=production node dist/server/index.js",
+        "start": "cross-env NODE_ENV=production node -r './dist/server/tracing.js' dist/server/index.js",
```
psarando, line 13: I was wondering why the tracing code to register the SDK couldn't be called inside `server/index.js`, but the couple of opentelemetry docs I came across both recommend that this tracing code should be loaded before the app loads 🤷

psarando (approved):
LGTM 👍

package.json
ianmcorvidae: It has to do with the automatic instrumentation, I think -- it needs to be loaded before the module it instruments is loaded, otherwise it won't be able to override the behavior to add the instrumentation. I think that it can sorta work if you can put it right at the top of the first thing you load... but this is much clearer (since it's more explicit about the intent) and probably less brittle. So I figured this was the way to go.

psarando: Sounds good to me 👍

ianmcorvidae:
Cool, thanks for the review. I'll get this merged, but I realize it might be best if I do it after the release tomorrow, just for a little extra safety.

PR #468, opened 2022-04-20
CORE-1731: made the user portal URL used by the client side code configurable

slr71:


config/default.yaml
```diff
@@ -1,6 +1,7 @@
 base_url: http://localhost:3000
 terrain_url: http://terrain
 listen_port: 3000
+user_portal_url: https://user.cyverse.org
```
ianmcorvidae, line 4: I noticed we have a different URL for the portal API further down in this file. I don't know if we should have both things, but if we do the defaults should probably match (and that one uses `portaldev`, looks like)

src/components/layout/AppBar.js
```diff
@@ -415,10 +415,7 @@ function DEAppBar(props) {
                                 profile={userProfile}
                                 onLogoutClick={onLogoutClick}
                                 onManageAccountClick={() =>
-                                    window.open(
-                                        constants.CYVERSE_USER_PORTAL,
-                                        "_blank"
-                                    )
+                                    window.open(config?.userPortalURL, "_blank")
```
ianmcorvidae, line 418: Not sure what this would do if config is null -- I think it'd call `window.open(null, "_blank")`, which testing in my browser appears to try to open that as though null is a string and a relative path. Which is probably not what we want. Not sure what makes sense to put here instead -- possibly just make it no-op if `config` is null?

src/components/layout/AppBar.js
```diff
@@ -456,7 +453,7 @@ function DEAppBar(props) {
                     profile={userProfile}
                     onLogoutClick={onLogoutClick}
                     onManageAccountClick={() =>
-                        window.open(constants.CYVERSE_USER_PORTAL, "_blank")
+                        window.open(config?.userPortalURL, "_blank")
```
ianmcorvidae, line 456: (same here, I guess)

ianmcorvidae (approved):
Seems pretty reasonable, couple of small comments but nothing that sticks out as any real blocker.

config/default.yaml
slr71: Ah, yes, the defaults should match. Good catch. As for whether or not we should have both, that's another question. I guess I was being lazy and I thought I'd ensure that they were the same when generating the configuration file. On second thought, however, I think it does make sense to have just one configuration setting for both. I'll fix that.

src/components/layout/AppBar.js
slr71: This was an odd case. I wasn't expecting `config` ever to be null, but when the page was rendered the first time, it was always null, then it was immediately re-rendered with the configuration values available. Several other files simply return null if the configuration value is missing in similar cases, though. My hope was that a redirect to null would never come up in production, but there's probably a better way to accomplish that.

slr71:
I ended up choosing a solution that uses our own user portal installation as the default until the the correct URL can be loaded from the configuration file. This seemed like a reasonable option, and it works locally. I'm going to merge this now because I'm going to need this for CyVerse Austria tomorrow. I'd still like a review or two, however, in case someone with more UI development experience has a better idea.

PR #471, opened 2022-05-18
CORE-1740 Allow CPU min to equal CPU max in app launch form

psarando:
Added a `validateNotAbove` check to the app launch form for use in the `min_cpu_cores` requirement field (users cannot add it as a rule in their app params, since there is no support for it in the apps service).


psarando:
Thanks for the review 👍

PR #474, opened 2022-07-13
CORE-1754 Add 409 ERR_CONFLICT ErrorHandler message

psarando:
This PR will update the `ErrorHandler` (which displays in error detail dialogs) to display a custom message for 409 `ERR_CONFLICT` service responses, including the list conflicting paths, if provided.

For example, if a user attempts to move 1 or more items with the `Move` dialog, then the usual `request failed` message will display at the top of the dialog:
Move request failed message

The `View Details` button will open the error details dialog, which will now display a custom message for 409 responses:
Conflict Error Dialog


src/components/data/MoveDialog.js
```diff
@@ -53,14 +53,13 @@ function MoveDialog(props) {
...
-        onError: (error) => {
-            setMoveError(error);
-        },
+        onError: setMoveError,
     });
 
     const handleMove = ({ dest }) => {
         const sources = selectedResources?.map((res) => res.path);
         if (status !== constants.LOADING && sources?.length > 0) {
+            setMoveError(null);
```
psarando, line 62: If the user does attempt the move action again, this will clear the previous error message display while the next request is processing.

ianmcorvidae:
This all looks reasonable to me, the only thing I'd ask is if this applies to everything that's a move under the covers. To be specific, it looks like this applies to the "Move" option, but this would also apply to rename, move to trash, restore from trash, and empty trash actions (which are also the 5 things that can return ERR_CONFLICT in data-info).

Regardless, looks good to merge to me, if we need to propagate it to more sections of the code later or anything.

psarando:
Thanks for the reviews! 👍

PR #476, opened 2022-08-01
Fix reporting of URL import submission errors

johnworth:
If the URL import dialog runs into an unknown error type, then it effectively hangs and makes it appear that it's taking forever to submit. This PR fixes that by allowing the logic to return a generic error message if the error type passed into onError() isn't in the format that we expect.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
     "requires": true,
     "packages": {
         "": {
-            "name": "sonora",
             "version": "0.0.1",
             "license": "BSD-3-Clause",
             "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
     "requires": true,
     "packages": {
         "": {
+            "name": "sonora",
             "version": "0.0.1",
             "license": "BSD-3-Clause",
             "dependencies": {
--- a/src/components/urlImport/index.js
+++ b/src/components/urlImport/index.js
@@ -137,17 +137,28 @@ const URLImportTextField = (props) => {
             errorCallback(null);
             setUploadURL("");
         },
+
         onError: (error) => {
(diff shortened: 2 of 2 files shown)


package-lock.json
```diff
@@ -5,7 +5,6 @@
     "requires": true,
     "packages": {
         "": {
-            "name": "sonora",
```
psarando, line 8: I don't think this needs to be removed.
I'm using `node` v18 now, which I think is the version that started adding this package name (or maybe it was `npm` v8 🤷 ).

psarando (changes requested):

package-lock.json
johnworth: Sorry about that, didn't realize my node install was so out of date. I upgraded it and npm and re-ran `npm install` and that line re-appeared. I've pushed it back up.

psarando: :np:

src/components/urlImport/index.js
```diff
@@ -137,17 +137,28 @@ const URLImportTextField = (props) => {
...
+                text =
+                    getErrorCode(error) === ERROR_CODES.ERR_EXISTS
+                        ? t("fileExists", {
+                              name: duplicateName,
+                              path: path,
+                          })
+                        : t("fileImportFail");
+            } else {
+                text = t("fileImportFail");
+            }
```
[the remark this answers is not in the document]
johnworth: Gah, I new the `var` didn't look right when I wrote it. It's been too long since I wrote anything in JS...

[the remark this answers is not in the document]
johnworth: I've pushed up the changes you suggested, thanks for catching that.

psarando (approved):
LGTM 👍

PR #479, opened 2022-09-07
CORE-1793: implement compute quota limit enforcement checks

slr71:
My main concern here is that I might have missed one or more places where apps can be launched. Any feedback would be greatly appreciated, of course, but it would be helpful if others can verify that I didn't miss anything.

### DIFF
--- a/src/components/apps/launch/AppInfo.js
+++ b/src/components/apps/launch/AppInfo.js
@@ -37,6 +37,10 @@ import { Info, MenuBook } from "@material-ui/icons";
 
 import { Skeleton } from "@material-ui/lab";
 
+import { appUnavailable } from "../utils";
+
+import { useConfig } from "contexts/config";
+
 const useStyles = makeStyles(styles);
 
 const LoadingErrorDisplay = ({ baseId, loadingError }) => {
@@ -60,9 +64,16 @@ const LoadingErrorDisplay = ({ baseId, loadingError }) => {
     );
 };
 
-const UnavailableMsg = ({ app, hasDeprecatedParams, baseId }) => {
+const UnavailableMsg = ({
+    app,
+    hasDeprecatedParams,
+    computeLimitExceeded,
+    baseId,
+}) => {
     let message = "";
     const { t } = useTranslation("launch");
+    const [config] = useConfig();
+
     if (app?.deleted) {
         message = (
             <Trans
@@ -97,6 +108,28 @@ const UnavailableMsg = ({ app, hasDeprecatedParams, baseId }) => {
                 }}
             />
         );
+    } else if (computeLimitExceeded) {
+        console.log("formatting the message");
+        message = (
+            <Trans
+                t={t}
+                i18nKey="computeLimitExceeded"
+                components={{
+                    buy: (
+                        <Link
+                            id={buildID(baseId, ids.BUTTONS.CONTACT_SUPPORT)}
+                            component="button"
+                            onClick={() => {
+                                window.open(
+                                    config?.subscriptions?.checkout_url,
+                                    "_blank"
+                                );
+                            }}
+                        />
+                    ),
+                }}
+            />
+        );
     }
 
     return (
@@ -107,7 +140,14 @@ const UnavailableMsg = ({ app, hasDeprecatedParams, baseId }) => {
 };
 
 const AppInfo = (props) => {
-    const { app, baseId, hasDeprecatedParams, loading, loadingError } = props;
+    const {
+        app,
+        baseId,
+        hasDeprecatedParams,
+        computeLimitExceeded,
+        loading,
+        loadingError,
+    } = props;
     const { t } = useTranslation("apps");
     const classes = useStyles();
     const theme = useTheme();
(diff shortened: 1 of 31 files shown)


src/components/apps/launch/AppInfo.js
```diff
@@ -97,6 +108,28 @@ const UnavailableMsg = ({ app, hasDeprecatedParams, baseId }) => {
                 }}
             />
         );
+    } else if (computeLimitExceeded) {
+        console.log("formatting the message");
```
ianmcorvidae, line 112: This can probably go away before merge, I guess

src/pages/apps/[systemId]/[appId]/launch.js
```diff
@@ -22,7 +23,19 @@ import {
...
 import { useUserProfile } from "contexts/userProfile";
 
-export default function Launch() {
+import {
+    getResourceUsageSummary,
+    RESOURCE_USAGE_QUERY_KEY,
+} from "serviceFacades/dashboard";
+import { useConfig } from "contexts/config";
+
+import { getUserQuota } from "../../../../common/resourceUsage";
```
Comment (ianmcorvidae) [inline src/pages/apps/[systemId]/[appId]/launch.js:32]: This is a totally random spot in the file, but I don't think I see changes in the `analyses/[analysisId]/relaunch` page, which reimplements a lot of what's in this file, so at a guess we'll need to.

ianmcorvidae (commented):
This is just a first-pass check, I'll try to get this running locally to check it further (as I mentioned in chat)

src/components/apps/launch/AppInfo.js
slr71: Ah, I missed that one. Thanks!

src/pages/apps/[systemId]/[appId]/launch.js
Comment (slr71) [reply src/pages/apps/[systemId]/[appId]/launch.js:32]: Thanks for catching that! I'll fix it.

slr71:
It looks like I need to update the dashboard to prevent running instant launches from there as well.

slr71:
I think that this is ready for review now.

psarando (commented):
LGTM 👍 

But I think I caught an issue where the `usageSummaryError` message won't display correctly from the data listing and the app bar.

src/components/apps/launch/AppInfo.js
```diff
@@ -97,6 +108,27 @@ const UnavailableMsg = ({ app, hasDeprecatedParams, baseId }) => {
...
             />
         );
+    } else if (computeLimitExceeded) {
+        message = (
+            <Trans
+                t={t}
+                i18nKey="computeLimitExceeded"
+                components={{
+                    buy: (
+                        <Link
```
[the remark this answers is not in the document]
slr71: I haven't been able to get that to work. When I try to use `ExternalLink`, I keep getting this error in Storybook:

```
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
```

Is there anything that I'm missing that might fix that error while still using `ExternalLink`?

src/components/data/listing/Listing.js
```diff
@@ -102,7 +102,7 @@ function Listing(props) {
     } = props;
     const [config] = useConfig();
 
-    const { t } = useTranslation("data");
+    const { t } = useTranslation(["data", "dashboard"]);
```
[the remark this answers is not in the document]
slr71: Ah, good catch! I'll fix that.

src/components/apps/launch/AppInfo.js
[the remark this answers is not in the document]
psarando: I'd probably have to see your changes, but did you mix up `import { ExternalLink } from "components/utils/ExternalLink";` instead of `import ExternalLink from "components/utils/ExternalLink";` as in the second suggestion for the error?

Here's an example of how it's used in a `Trans` component like this:
https://github.com/cyverse-de/sonora/blob/6d19ea18a979979d7b942485c2c05b5a097e95b0/src/components/utils/ComingSoonInfo.js#L32-L40

So here `onClick` can be replaced with `href={config?.subscriptions?.checkout_url}`.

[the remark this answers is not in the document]
slr71: My mistake. I'd forgotten that the input was different. I'll give that a try. Thanks!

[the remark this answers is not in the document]
slr71: s/input/import/

src/components/instantlaunches/listing/index.js
```diff
@@ -26,12 +26,24 @@ import {
...
-    const { baseId } = props;
+    const { baseId, showErrorAnnouncer } = props;
     const [showAppDetails, setShowAppDetails] = useState(null);
+    const [config] = useConfig();
+    const [computeLimitExceeded, setComputeLimitExceeded] = useState(
+        !!config?.subscriptions?.enforce
+    );
 
-    const { t } = useTranslation("instantlaunches");
+    const { t } = useTranslation("instantlaunches", "dashboard");
```
psarando, line 46: This should be an array:
```suggestion
    const { t } = useTranslation(["instantlaunches", "dashboard"]);
```

slr71: 🤦‍♀️ I should have caught that. Thanks for catching it.

slr71:
I think all of the problems are resolved now. @psarando would you mind another quick look at it just to make sure everything looks okay? I'm not confident in my UI prowess at all. 😆

psarando (approved):
LGTM 👍 🎉

slr71:
Thanks for the reviews!

PR #480, opened 2022-09-22
CORE-1531 Add Create App Version page and Editor support

psarando:
This PR will add a page for creating new app and workflow versions, and update the app and workflow editors to support creating new versions.

The app edit page also now requires a version ID, and has been renamed to include `versions/[versionId]/` in its path, and the UI will now always use a version ID when updating and copying apps and workflows.

Creating a new app or workflow version will fetch the UI for the latest version, but leave the version label blank in the editor to force the user to enter a new version (which is a required field):
Create App Version - Info Step

---

Create Workflow Version - Info Step

---

The editor will also validate that the version label does not conflict with any other version labels:
App Editor Version Validation

The editor allows the user to update any fields in a new app version, including changing the tool, or adding and removing any parameters or groups; and in workflows, the user may change app steps and mappings.

### DIFF
--- a/public/static/locales/en/app_editor.json
+++ b/public/static/locales/en/app_editor.json
@@ -9,6 +9,7 @@
     "appSaved": "App Saved",
     "appSavedSuggestions": "The app is saved, so you can test it out by launching the app, or you can exit the editor now.",
     "appSaveErr": "Unable to save app. Please try again.",
+    "appVersion": "Version",
     "argumentOption": "Argument option",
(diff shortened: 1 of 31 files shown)


ianmcorvidae (approved):
I haven't actually pulled down the code to try to run, but reading through it, nothing sticks out as potentially problematic to me.

slr71 (commented):
Looks good to me. :shipit:

slr71 (approved):
I meant to mark the PR as approved with the last review. 😖

PR #481, opened 2022-09-22
Add Version Selection to App Launch, Editors, and Detail views

psarando:
This PR will add Version Selection to the app details drawer, app and workflow editors, and app launch forms.

An app version launch page was added for handling specific app versions, but the original app launch page still needs to support saved launches and non-DE apps, in addition to launching the latest version of DE apps.

Also, now that the app details endpoint includes the same info as returned in the app listing endpoint, the details panel now only needs to call the details endpoint.

### Version Selector in the App Details Drawer:
App Details Version Selector
Most of this info will not change between versions, since it's associated with the overall app, except for the tools used and integrator info, which are associated with each version (and app documentation in a future PR).

---

### Version Selector in the App Editor:
App Editor Version Selector
The selector is at the top, just above the editor steps, and the version label can be updated in the App Info step, just below the app name field.

---

### Version Selector in the App Launch Form:
App Launch Version Selector


slr71 (approved):
Looks good to me! :shipit:

psarando:
Thanks for the review! 👍

PR #482, opened 2022-09-22
CORE-1486 Add Documentation tab to App Details Drawer

psarando:
This PR will add a Documentation tab to the App Details Drawer.

The documentation panel was refactored as a separate component from its enclosing dialog, so that it could be reused in the app details panel.

This also includes a bug fix where the documentation dialog would not allow the user to close the dialog without saving if any changes were made. Now the dialog (and the details drawer) will ask the user if they want to discard their changes when attempting to close the dialog/drawer.

App Details Documentation tab


slr71 (approved):
I didn't catch any problems 👍

psarando:
Thanks for the review! 👍

PR #483, opened 2022-09-30
CORE-1534 Update AdminAppDetailsDialog and app listings for versions

psarando:
This PR will update the `AdminAppDetailsDialog` with the `VersionSelection` component and update admin endpoints with a version ID.

Admin App Details Dialog

Also fixes a bug where adding and then deleting documentation in the form for an app without existing docs would attempt to update the non-existent docs.

---

Since the `AdminAppDetailsDialog` now only updates specific app versions, delete/restore and disable/enable menu items have been added to the admin app listing row dot-menu, to allow admins to delete/restore or disable/enable the entire app (all versions) at once.

Admin App Listing row dot-menu

Admin App Restore confirmation dialog

Admin App Delete confirmation dialog


slr71 (approved):
I didn't catch any problems. 👍

psarando:
Thanks for the review! 👍

PR #484, opened 2022-10-05
Hide resource usage when subscriptions not enforced

psarando:
This PR will hide data and CPU resource usage, plan info, and the buy button when the `subscriptions.enforce` config flag is set to `false`, which disables fetching info from the resource usage summary endpoint; currently resulting in empty or blank info displayed in these components:

Empty Resource Summaries

---

This PR will hide these components when this info is empty, except for the Analyses Stats, since those are fetched from a different endpoint:

Hidden Resource Summaries

Note that this screenshot is in QA, where I have no analyses in the QA db yet.


slr71 (approved):
Looks good to me! ✨

psarando:
Thanks for the review! 👍

PR #485, opened 2022-10-11
CORE-1801 Add unique Resource Query Keys to launch pages

psarando:
This PR will add unique Resource Query Keys to the app launch and analysis relaunch pages, which will prevent the `Browse` button for analysis output or inputs from resetting the form, since the data listing also uses the `RESOURCE_USAGE_QUERY_KEY` to look up data usage, which caused the App Launch page query to also enter a fetching state, which reset the launch form's `loading` prop, causing the entire `AppLaunch` component to re-render.


slr71 (approved):
Looks good to me!

ianmcorvidae (approved):

psarando:
Thanks for the reviews! 👍

PR #486, opened 2022-10-19
CORE-1804 Fix app listing navigation bug when apps are selected

psarando:
The `formatSharedApps` function would throw an uncaught exception if `selected` contained an ID that was not found in the current `data.apps` list, which would display `An unexpected error has occurred` message over the apps listing.

This could happen if the user selects an app, then attempts to navigate to another page in the listing; or if a user changes app categories, selects an app, then uses the browser's back button to navigate to the previous category.

To prevent this, the `getSelectedApps` callback was updated to only return apps included in the current `data.apps` list.

Also, the `selected` state is now reset to an empty list in `handleChangePage` and `handleViewAllApps` (it was already done for other user navigation actions), which appears to clear the selection immediately on these user actions (updating the select-all checkbox and other toolbar buttons).

The `selected` state is also reset in a `useEffect` whenever `data` changes, which also covers cases where the browser's back/forward actions are used, but the selection update is delayed until the `data` query completes and the loading mask clears.

### DIFF
--- a/src/components/apps/listing/Listing.js
+++ b/src/components/apps/listing/Listing.js
@@ -93,7 +93,10 @@ function Listing(props) {
     const [navError, setNavError] = useState(null);
 
     const getSelectedApps = useCallback(() => {
-        return selected.map((id) => data?.apps.find((app) => app.id === id));
+        // Sometimes selected gets out of sync
+        // and contains IDs not in the current data set.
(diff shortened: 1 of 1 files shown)


slr71 (approved):
Looks good to me! 👍

psarando:
Thanks for the review! 👍

PR #487, opened 2022-10-21
CORE-1533 Add admin app sharing support

psarando:
This PR will allow admins to open the apps sharing dialog from the admin apps listing, so they may grant any user (including themselves) `write` permissions on apps with the new admin apps (un)sharing endpoints.

This is required in order to create new versions for public apps.


slr71 (approved):
I didn't catch any problems 👍

PR #488, opened 2022-10-21
CORE-1531 Restrict new public app versions to public tools

psarando:
This PR will restrict the app editor's tool selector to public tools for new versions of public apps, and also restrict the workflow editor's app search dialog to public apps for new versions of public workflows.

For the app editor's tool selector, this is accomplished by hiding the tool `PermissionsFilter` and setting the tool search query `permFilter` to `public`.

Public Tool Search dialog

---

Since the app search endpoint does not support filtering by public apps, the workflow editor's app search dialog instead adds a validation check that displays a validation error if a public app is selected for public workflows.

---

Public Workflow App Validation


slr71 (approved):
I didn't catch any problems 👍

psarando:
Thanks for the review 👍

PR #489, opened 2022-10-26
CORE-1536 Submit version IDs in app/saved/instant launches

psarando:
This PR will update the app launch form `initAppLaunchValues` and `formatSubmission` functions to include the app version ID in returned values, and update the `instantlyLaunch` service facade function to ensure the app version ID is included in launch submissions.


psarando:
This is a pretty straight-forward (but essential) change, so I'm going to go ahead and merge this now; but reviews still welcome 🙇

PR #490, opened 2022-11-04
CORE-1536 Update App Launch Version Select helper text

psarando:
This PR will update the `error` prop on the App Launch's `VersionSelection` help text, so that the other/newer-version-available text is more noticeable.

Deleted App Version help text

Newer App Version Available help text


psarando:
This is pretty straight forward, so I'll merge this now, but reviews still welcome.

PR #495, opened 2022-12-01
CORE-1809 Add app version labels to Admin Instant Launch List

psarando:
This PR will add app version labels to the Admin Instant Launch List, which required some refactoring to allow space for the app name and version labels, replacing row action buttons with a `RowDotMenu`, and also to use the `BasicTable` component for consistency with the user instant launch listing.

Icons were also added in an initial column to indicate where instant launches are currently displayed to users.

Admin Instant Launch List - Row Menu - Add to Nav

Admin Instant Launch List - Row Menu - Add to all


slr71 (approved):
I didn't catch any problems 👍

psarando:
Thanks for the review 👍

PR #497, opened 2022-12-07
CORE-1734 Make analysis termination more prominent

psarando:
This PR will make analysis termination more prominent by adding a `Terminate` action to each running analysis in the Analyses Listing, move the `Terminate` action out of the Analyses Listing toolbar's dot menu and out into the toolbar itself, and make the color of the `Terminate` action icons red, including in the dashboard.

Analyses Listing

---

Dashboard Running Analyses


psarando:
Another suggestion included in the description for CORE-1734 was to "Add the ability to stop all running analyses for a user with one click".

I'm not sure if we want to still add something like that even after these changes, but it could still be possible (by having the `Terminate` button always showing in the listing, and it could behave like this when analyses are running and nothing is selected in the listing, for example).

psarando:
> I'm not sure if we want to still add something like that even after these changes, but it could still be possible (by having the `Terminate` button always showing in the listing, and it could behave like this when analyses are running and nothing is selected in the listing, for example).

Actually, I think this suggestion may be confusing to users, since the `Terminate` button would behave differently depending on whether running analyses are selected or not.

psarando:

PR #501, opened 2022-12-16
CORE-1806 Analysis Listing Improvements

psarando:
This PR will update the Analysis Listing to make it a little easier to use.

The analysis listing columns will be reordered as requested:

* Analysis Name
* Status
* Start Date
* Duration
* Action Icons

In addition, the analysis names are now links to their landing pages, and their display now allows browsers to word-break at underscores in the analysis name, if necessary.

Notice that before these changes, if analysis names are very long, they usually include underscores which do not normally wrap in browser layouts. These long, unwrapped names can cause the last couple of columns in the listing to scroll out of view, including the analysis action buttons, which now allow users to quickly terminate running analyses:

Analysis Listing - before

---

Analysis Listing - before - scrolled right

---

After these changes:

Analyses Listing - after

---

The App name and Owner columns were also removed to prevent the Action icons from scrolling off screen in the listings, so the details drawer was updated to include this info, along with additional details that are also included in the Analysis Landing page:

Analyses Listing - details

Analysis Landing page

---

In a wider laptop/desktop display:

Analyses Listing - wide


psarando:
I'm going to put this PR into `Draft` until we can get more feedback for it in the next UX meeting.

slr71 (approved):
I didn't catch any issues.

src/components/analyses/details/Drawer.js
```diff
@@ -108,16 +99,22 @@ function DetailsDrawer(props) {
                 variant: "outlined",
             }}
         >
-            <div className={classes.drawerHeader}>
-                <Typography variant="h6">{analysisName}</Typography>
-            </div>
+            <Typography variant="h6" className={classes.drawerHeader}>
+                {analysisName}
+            </Typography>
```
slr71, line 104: Just for my own edification, did removing the div have any effect on the page display format?

src/components/analyses/details/Drawer.js
```diff
@@ -108,16 +99,22 @@ function DetailsDrawer(props) {
...
                     label={t("info")}
                     id={infoTabId}
                     aria-controls={buildID(infoTabId, ids.PANEL)}
                 />
+                <DETab
+                    value={TABS.analysisHistory}
+                    label={t("statusHistory")}
+                    id={historyTabId}
+                    aria-controls={buildID(historyTabId, ids.PANEL)}
+                />
```
slr71, line 117: 👍

src/components/analyses/landing/AnalysisSubmissionLanding.js
```diff
@@ -480,71 +470,7 @@ export default function AnalysisSubmissionLanding(props) {
...
-                    <GridLabelValue label={t("endDate")}>
-                        <InfoGridValue>
-                            {formatDate(analysis?.enddate)}
-                        </InfoGridValue>
-                    </GridLabelValue>
-                    <GridLabelValue label={t("user")}>
-                        <InfoGridValue>{username}</InfoGridValue>
-                    </GridLabelValue>
-                </Grid>
+                <DetailsPanel baseId={baseId} analysis={analysis} spacing={3} />
```
slr71, line 473: 👍

src/components/analyses/listing/TableView.js
```diff
@@ -18,61 +19,67 @@ import DETableHead from "components/table/DETableHead";
...
     const analysis = props.analysis;
     const name = analysis.name;
     const baseId = props.baseId;
+    const [href, as] = getAnalysisDetailsLinkRefs(analysis.id);
+
+    // Inserting the <wbr> tag at underscores allows browsers to wrap long
+    // analysis names at those underscores.
+    const linkText = name
+        ?.split("_")
+        ?.reduce((acc, cur, i) => [...acc, "_", <wbr key={i} />, cur]);
```
slr71, line 63: ✨

slr71 (approved):
I don't know if all of the changes are in place yet, but what's there so far looks good to me.

src/components/analyses/details/Drawer.js
psarando: Nope, in this case, the wrapping `div` was extraneous, and these `drawerHeader` styles can be applied directly to the `Typography` component (which will render as an `h6` HTML element here).

psarando:
> I don't know if all of the changes are in place yet, but what's there so far looks good to me.

I made a couple more changes. Notably I refactored the `Duration` column in the listing to use the same formatted values that are used for the Analysis Landing page and the Dashboard.

This PR should be ready for review now 👍

slr71 (approved):
I didn't catch any issues. 👍 :shipit:

psarando:
Thanks for the review! 👍 

I'll merge this now so it makes the cut for next release, but additional reviews still welcome 🙇

PR #506, opened 2023-01-21
CORE-1867 Migrate to CodeMirror v6

psarando:
This PR will remove the `react-codemirror2` wrapper library dependency by integrating the UI directly with [CodeMirror v6](https://codemirror.net/).

Removing the `react-codemirror2` wrapper library dependency will allow the UI to use React v17, allowing us to update our `next.js` dependency.

These updates also include refactoring some state management between the CodeMirror Editor component and viewer components, such as updating the TextViewer components to manage their own `dirty` state (I can't find an equivalent to the previous `isClean` API), and allowing the CodeMirror Editor view to set its own height.

I'm not including any screenshots in this PR since the editors should look almost exactly the same.


slr71 (approved):
I didn't catch any issues. 👍 :shipit:

psarando:
Thanks for the review 👍

PR #509, opened 2023-02-02
Core 1870

sboleyn:
These are the edits made to go up during the Feb. maintenance to add the subscriptions UI. Storybooks is currently missing from this PR and is a WIP.


src/components/subscriptions/edit/EditQuotas.js
```diff
@@ -0,0 +1,203 @@
...
+    useQuery({
+        queryKey: [
+            SUBSCRIPTIONS_QUERY_KEY,
+            { parentId: parentId, user: subscription?.user.username },
+        ],
+        queryFn: () =>
+            getSubscriptions({ searchTerm: subscription?.user.username }),
+        enabled: true,
+        onSuccess: (data) => {
+            setSelectedSubscription(data.result.subscriptions[0]);
```
psarando, line 43: I thought this was the same info that also came from the subscription listing, which is already being passed to this component in the `subscription` prop.

So maybe there is no need for this query after all?

If so, then all references to `selectedSubscription` in this component could be replaced with the `subscription` prop instead.

psarando (approved):
Looks good! 👍 

I had few suggestions for clean up and refactoring ✨ , but nothing too pressing (although I'm a little worried about the nested Formik fields, since I haven't had a chance to try out these changes on my box in dev mode).

Since we want these changes in for the next release, you can :shipit: when you feel it's ready and working (and anything left can be addressed in a future PR).

src/components/subscriptions/edit/EditQuotas.js
sboleyn: Initially I had left this query out, but I found that the component wasn't updating the quota values after an update. At least, it wasn't updating consistently. I spent some time debugging this but wasn't able to resolve it by invalidating the query.

PR #511, opened 2023-02-08
fixed a response body field name and renamed some variables

slr71:
This fixes a problem with the subscription details not being displayed correctly on the dashboard.


slr71:
This problem can erroneously prevent users from launching jobs or uploading data, so I'm going to merge it now. Reviews are still welcome and appreciated.

psarando (commented):
LGTM :shipit:

slr71:
Thanks for the review!

PR #512, opened 2023-02-09
Add storybooks to Subscriptions UI

sboleyn:
This is a request to add React storybooks to the Subscriptions UI

### DIFF
--- a/src/components/subscriptions/edit/EditSubscription.js
+++ b/src/components/subscriptions/edit/EditSubscription.js
@@ -102,7 +102,7 @@ function EditSubscriptionDialog(props) {
             }}
             enableReinitialize={true}
         >
-            {({ handleSubmit, ...props }) => {
+            {({ handleSubmit, resetForm }) => {
                 return (
                     <Form>
                         <DEDialog
@@ -110,7 +110,7 @@ function EditSubscriptionDialog(props) {
                             open={open}
                             onClose={() => {
                                 onCloseForm();
-                                props.resetForm();
+                                resetForm();
                             }}
                             fullWidth={true}
                             title={
@@ -127,7 +127,7 @@ function EditSubscriptionDialog(props) {
                                         )}
                                         onClick={() => {
                                             onCloseForm();
-                                            props.resetForm();
+                                            resetForm();
                                         }}
                                     >
                                         {t("cancel")}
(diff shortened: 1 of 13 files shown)


src/components/subscriptions/edit/EditSubscription.js
```diff
@@ -192,31 +192,24 @@ function EditSubscriptionForm(props) {
...
-                        size="small"
-                        {...props}
-                        {...selectProps}
+            <FastField
+                name="plan_name"
+                component={FormTextField}
+                variant="outlined"
+                select
+                size="small"
+                {...selectProps}
```
psarando, line 201: I don't think `selectProps` exists in this context, so this line can be removed.

stories/subscriptions/Listing.stories.js
```diff
@@ -0,0 +1,126 @@
...
+        />
+    );
+}
+/**
+ * The subscription listing for a parameterized API call. This function retrieves the
+ * query parameters from the Axios config and adjusts the result so that we
+ * can test sorting and paging in Storybook.
+ * @param {Object} config - the Axios config
+ */
+function parameterizedSubscriptionListing(config) {
```
psarando, line 38: 👍

stories/subscriptions/EditSubscription.stories.js
```diff
@@ -0,0 +1,53 @@
...
+    } else {
+        return (
+            <EditSubscriptionDialog
+                open={true}
+                onClose={onClose}
+                parentId={parentId}
+                subscription={subscription}
+            />
+        );
+    }
```
psarando, line 39: This might also work:
```suggestion
        return (
            <EditSubscriptionDialog
                open={true}
                onClose={onClose}
                parentId={parentId}
                subscription={!newSubscription && subscription}
            />
        );
```

psarando (approved):
LGTM 👍 ✨

src/components/subscriptions/edit/EditSubscription.js
sboleyn: I believe that is in use to declare id and label. I left that from the reference I used because I think it cleans up that component a bit.

psarando: Oops, I must have overlooked that in the last PR, but I see it now 👍

PR #515, opened 2023-02-17
CORE-1876 Upgrade to next.js v12

psarando:
This PR will upgrade the UI to `next.js` v12, as well as `next-i18next` to v13, and `react` to v17.

This required some updates to the way `i18n` is configured for all of our `next.js` pages, as well as the `I18nProviderWrapper` for `Storybook` and `Jest` tests.

There should not be any visible changes to the UI, but one of the big changes in this upgrade is that translation namespaces are loaded server-side, so the client should not have to download as many JSON files for i18n messages.


next-i18next.config.js
```diff
@@ -0,0 +1,26 @@
+const path = require("path");
+
+const isProduction = process.env.NODE_ENV === "production";
+const localePath =
+    typeof window === "undefined"
+        ? path.resolve("./public/static/locales")
+        : "/static/locales";
+
+module.exports = {
+    debug: !isProduction,
```
psarando, line 10: Turning on the debug flag in dev helps catch i18n key typos and missing i18n messages, which already helped me catch these: 8415b299a226ab229030dbe378e4c2e037e005a8

src/components/data/SelectionDrawer.js
```diff
@@ -155,29 +155,30 @@ function SelectionToolbar(props) {
...
-                                    : theme.spacing(1),
-                            }}
-                            size="small"
-                        >
-                            {t("selCurrentFolder")}
-                        </Button>
-                    </>
-                )}
+                {acceptedType === ResourceTypes.FOLDER &&
+                    selectedTotal === 0 && (
```
psarando, line 159: My editor wanted to format this section with prettier 🤷

src/__mocks__/i18nProviderWrapper.js
```diff
@@ -0,0 +1,35 @@
...
+ * @returns {JSX.Element}
+ * @constructor
+ */
+export function I18nProviderWrapper({ children }) {
+    return (
+        <Suspense fallback={"Loading i18n..."}>
+            <I18nextProvider i18n={I18NextClient}>{children}</I18nextProvider>
+        </Suspense>
+    );
+}
```
slr71, line 35: 👍

public/static/locales/en/common.json
```diff
@@ -90,6 +90,7 @@
     "update": "Update",
     "updateAccountInformation": "Update Account Information",
     "updateNow": "Update Now",
+    "usageSummaryError": "Unable to get usage summary.",
```
psarando, line 93: Moving this message to the `common` ns prevents other pages from requiring the entire `dashboard` ns for only this message.

src/components/dashboard/dashboardItem/ResourceUsageItem.js
```diff
@@ -71,7 +70,7 @@ export default function ResourceUsageItem(props) {
             />
             {resourceUsageError && (
                 <ErrorTypographyWithDialog
-                    errorMessage={t("usageSummaryError")}
+                    errorMessage={t("common:usageSummaryError")}
```
slr71, line 73: 👍

src/pages/dashboard.js
```diff
@@ -1,13 +1,25 @@
...
-DashboardPage.getInitialProps = async () => ({
-    namespacesRequired: ["dashboard", "apps"],
-});
+export async function getStaticProps(context) {
+    // The `locale` prop is undefined in this page for some reason.
+    const { locale, defaultLocale } = context;
+
+    return {
+        props: {
+            ...(await serverSideTranslations(locale ?? defaultLocale, [
```
psarando, line 16: The `locale` prop is undefined in this page for some reason (but not the `defaultLocale`). I think due to the way we render this page for the root path:
https://github.com/cyverse-de/sonora/blob/961672814c603ce020b46293e70e44985b7882ec/src/server/index.js#L174-L176

slr71 (approved):
Looks good to me! :thu

psarando:
Thanks for the review! 👍 

I'll go ahead and merge now, but additional reviews still welcome 🙇

PR #516, opened 2023-02-21
Add link from Sub UI details to User Portal

sboleyn:
To see this feature, please click on the row dot menu for a user subscription. This feature is under Details. The Verified User badge is a button that will navigate to the user's page in the User Portal.

### DIFF
--- a/src/components/subscriptions/details/Drawer.js
+++ b/src/components/subscriptions/details/Drawer.js
@@ -229,7 +229,11 @@ function SubscriptionHeader(props) {
         <div className={classes.drawerHeader}>
             {portalId && (
                 <Tooltip title={linkToUserPortal}>
-                    <IconButton href={linkToUserPortal} target="_blank">
+                    <IconButton
+                        href={linkToUserPortal}
+                        target="_blank"
+                        rel="noopener noreferrer"
+                    >
                         <VerifiedUser fontSize="small" color="primary" />
                     </IconButton>
                 </Tooltip>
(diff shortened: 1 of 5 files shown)


src/components/subscriptions/details/Drawer.js
```diff
@@ -201,10 +220,21 @@ function SubscriptionDrawer(props) {
...
+    const { portalId, username } = props;
     const classes = useStyles();
+    const baseURL = constants.DEFAULT_USER_PORTAL_URL;
+    const subURL = navigationConstants.ADMIN_USER_PORTAL_USERS;
+    const linkToUserPortal = `${baseURL}${subURL}/${portalId}`;
     return (
         <div className={classes.drawerHeader}>
+            {portalId && (
+                <Tooltip title={linkToUserPortal}>
+                    <IconButton href={linkToUserPortal} target="_blank">
```
psarando, line 232: When using `target="_blank"`, the `rel="noopener noreferrer"` attribute should also be used.

For example:
https://github.com/cyverse-de/sonora/blob/682a20b94486a97008b5b053f14f2512dd241751/src/components/utils/ExternalLink.js#L10-L12
See also https://v4.mui.com/components/links/#security

psarando (approved):
LGTM 👍 

Though I would add the `rel` attribute to the `IconButton` before merging.

src/components/subscriptions/details/Drawer.js
sboleyn: Thank you for sharing this! I will implement.

PR #517, opened 2023-02-22
CORE-1882 Fix duration display for cancelled analyses

psarando:
This PR fixes the display bug in the Analyses Listing's "Duration" column, where analyses cancelled by the user will show a duration of "about 53 years".


psarando:
I'll go ahead and merge this now, but reviews still welcome 🙇

PR #518, opened 2023-03-01
add paid flag, fix user lookup

sboleyn:


src/components/subscriptions/details/Drawer.js
```diff
@@ -269,12 +244,23 @@ function UsagesDetails(props) {
...
+                                {formatFileSize(item.usage)}
+                            </Typography>
+                        );
+                    } else {
+                        return (
+                            <Typography>
+                                {item.usage} {item.resource_type.unit}
+                            </Typography>
+                        );
+                    }
```
psarando, line 263: Each element returned in this `map` array should have a `key`, including the `Typography` for CPUs display.

So I would nest the `resourceInBytes` check inside the `Typography` tag:
```suggestion
                    return (
                        <Typography key={index}>
                            {resourceInBytes
                                ? formatFileSize(item.usage)
                                : `${item.usage} ${item.resource_type.unit}`}
                        </Typography>
                    );
```

psarando (approved):
LGTM 👍

src/components/subscriptions/details/Drawer.js
sboleyn: Yes this looks much better. I'll implement this before merging.

PR #522, opened 2023-03-14
Update available add-on

sboleyn:
- Initial commit for the update available add-on functionality to the Subscription UI. 

- In EditQuotas, I changed formatting of submission so that someone could hypothetically update a quota to 6.5 GiB or any other decimal.


src/components/subscriptions/addons/formatters.js
```diff
@@ -1,39 +1,84 @@
...
+        defaultPaid,
+        resourceType,
+    } = values;
 
-    const { id, unit, ...rest } = resourceType;
+    const resourceObj = resourceTypes.filter(
+        (resource) => resourceType === resource.unit
+    );
+
+    const { id, unit, name } = resourceObj[0];
```
psarando, line 53:
```suggestion
    const resourceObj = resourceTypes.find(
        (resource) => resourceType === resource.unit
    );

    const { id, unit, name } = resourceObj;
```

psarando (approved):
LGTM 👍

PR #523, opened 2023-03-15
CORE-1876 Replace `express-ws` with `ws`

psarando:
HMR in `next.js` v12 uses web sockets now, which were being consumed or blocked by `express-ws`.
See HenningM/express-ws#162 for example.

This replaces `express-ws` with the `ws` package, combining a few examples of possible solutions found online.

The `express-ws` project appears dead at this time, and it used `ws` under the hood anyway.


psarando:
I'll go ahead and merge this now so we can start testing notifications in QA (I already tested these changes in local dev), but reviews are still welcome 🙇

PR #525, opened 2023-03-17
Admins can delete available add-ons

sboleyn:
Includes deleting one add-on, multiple add-ons, and a select all feature.


src/components/subscriptions/addons/AddOnsListing.js
```diff
@@ -43,27 +54,101 @@ function AddOnsListing(props) {
...
+            onSuccess: () => {
+                announce({
+                    text: t("addonDelete"),
+                });
+                setSelected([]);
+                queryClient.invalidateQueries(AVAILABLE_ADDONS_QUERY_KEY);
+            },
+            onError: (error) => {
+                showErrorAnnouncer(t("deleteAddonError"), error);
+                setSelected([]);
```
psarando, line 73: I'm not sure about clearing the user's selection on an error, because it may just be a temporary network or service error. So the user might want to try the delete again with the same selection (and a lot of times our error messages tell the user to please try again).

src/components/subscriptions/addons/AddOnsListing.js
```diff
@@ -43,27 +54,101 @@ function AddOnsListing(props) {
...
+
+        if (start < 0) {
+            return;
+        }
+
+        const rangeIds = [];
+        for (let i = start; i <= end; i++) {
+            rangeIds.push(addonsData?.addons[i].uuid);
+        }
+        // Toggle the selection based on the last tool clicked.
```
psarando, line 120: Copy-paste typo 😅 
```suggestion
        // Toggle the selection based on the last addon clicked.
```

psarando (approved):
LGTM 👍 

Though I do wonder about clearing the user's selection on any deletion error 🤔

src/components/subscriptions/addons/AddOnsListing.js
sboleyn: That's a good point. I'll remove line 73.

src/components/subscriptions/addons/AddOnsListing.js
sboleyn: Good catch!

PR #526, opened 2023-04-01
create add a subscription add-on functionality

sboleyn:
Added functionality to add an add-on to a subscription

- Modified code to allow users to add add-ons to their subscriptions
- Updated comments to refer to "add-ons" instead of "tools"
- Rearranged some folders, resulting in changes to several files

Apologies for the folder rearrangement and the increased number of changed files.


src/components/subscriptions/edit/EditSubAddons.js
```diff
@@ -0,0 +1,206 @@
+import React from "react";
+
+import { mapSubAddonsPropsToValues } from "./formatters";
+import ids from "../ids";
+
+import { useTranslation } from "react-i18next";
```
psarando, line 6: We should only import `useTranslation` from our `i18n` module, which in turn is importing it from `next-i18next`.

Importing `useTranslation` from `react-i18next` can actually cause conflicts with the translations setup by `next-i18next`. To help avoid this confusion, we can stick with importing from our own `i18n` namespace.
```suggestion
import { useTranslation } from "i18n";
```

Also, it looks like `useTranslation` is being imported from `react-i18next` in `src/components/subscriptions/edit/EditQuotas.js` and `src/components/subscriptions/edit/EditSubscription.js`, so those should be updated to import from `i18n` as well (sorry I didn't catch those sooner).

src/components/subscriptions/edit/AddSubAddon.js
```diff
@@ -0,0 +1,127 @@
...
+import { Field, Form, Formik } from "formik";
+import { Button, MenuItem, Typography } from "@material-ui/core";
+import DEDialog from "components/utils/DEDialog";
+import FormTextField from "components/forms/FormTextField";
+import { announce } from "components/announcer/CyVerseAnnouncer";
+import ErrorTypographyWithDialog from "components/error/ErrorTypographyWithDialog";
+
+import ids from "../ids";
+import buildID from "components/utils/DebugIDUtil";
+import { useTranslation } from "next-i18next";
```
psarando, line 13:
```suggestion
import { useTranslation } from "i18n";
```

psarando (changes requested):
I haven't finished reviewing everything yet, but I caught a couple changes that I thought I could point out now, so they can get fixed in this PR early.

src/components/subscriptions/edit/EditSubAddons.js
sboleyn: Thank you, I will address this

src/components/subscriptions/edit/AddSubAddon.js
sboleyn: I believe I was letting VSCode autofill this import for me, thank you for catching this.

src/components/subscriptions/edit/AddSubAddon.js
```diff
@@ -0,0 +1,127 @@
...
+                            select
+                            size="small"
+                            id={ids.SUB_ADDONS.UUID_FIELD}
+                            label={t("addonName")}
+                        >
+                            {availableAddons?.map((addon, index) => (
+                                <MenuItem
+                                    id={ids.SUB_ADDONS.AVAILABLE_ADDONS}
+                                    key={index}
+                                    value={addon.uuid}
```
psarando, line 106: A lot of React tutorials say it's fine to use a `map` index as the `key`, since it just has to be some unique value, but if you already have a UUID available, I think it's generally best to use that instead.

Probably not an issue here, but if the code ever needs to reorganize these items, then it may be less thrash to use the UUID for the `key`; or in the case of the app editor, where the user can reorder param items, it actually caused bugs to use an index rather than a UUID.

src/components/subscriptions/edit/EditSubAddons.js
```diff
@@ -0,0 +1,206 @@
...
+                                        </TableCell>
+                                        <TableCell>
+                                            <Field
+                                                name={`${name}.${index}.amount`}
+                                                component={FormNumberField}
+                                                disabled
+                                                fullWidth={false}
+                                                id={ids.SUB_ADDONS.NAME_FIELD}
+                                                required
+                                            />
```
psarando, line 185: I don't think this needs to be changed for this PR, and I'm not sure if there are plans in the future to allow admins to edit these values, but if not, then I think it makes these values a little harder to read displaying them in disabled form fields.

Formik would not be needed at all to display these values if we do not plan to allow admins to edit these values, and they can be displayed in a simpler listing, similar to `src/components/subscriptions/edit/Usages.js`.

psarando (approved):
LGTM 👍 

I just had a couple of fairly minor style suggestions, but I don't think they are required for this PR.

src/components/subscriptions/edit/AddSubAddon.js
sboleyn: Thank you for catching this, and that makes sense. I will implement this for the next PR.

src/components/subscriptions/edit/EditSubAddons.js
sboleyn: One of the requirements for this UI is that the admins can edit these values. That will be included in the next PR. However, I have noted that Formik fields wouldn't be needed if the display could be handled with a listing :).

psarando: I suspected that might be the case 👍 but I thought I'd mention it anyway, just in case 😅

PR #529, opened 2023-04-12
Enable data sharing

ianmcorvidae:
This is not done but I wanted to get the start of it up.

Outstanding stuff:

- [x] Show groups properly in the sharing dialog (rather than showing '@grouper-....')
- [ ] Show group name properly in the popup that happens when sharing/unsharing (rather than '@grouper-')
- [ ] More gracefully handle the "group hasn't been propagated to irods yet" case somehow

### DIFF
--- a/src/components/sharing/index.js
+++ b/src/components/sharing/index.js
@@ -7,8 +7,6 @@
 
 import React, { useEffect, useState } from "react";
 import buildID from "components/utils/DebugIDUtil";
-import { announce } from "components/announcer/CyVerseAnnouncer";
-import { INFO } from "components/announcer/AnnouncerConstants";
 import {
     Button,
     Dialog,
@@ -35,9 +33,7 @@ import {
     getUnsharingUpdates,
     getUserMap,
     getUserSet,
-    isGroup,
 } from "./util";
-import { TYPE } from "./constants";
 import { ADMIN_APPS_QUERY_KEY } from "serviceFacades/apps";
 import { getUserInfo } from "serviceFacades/users";
 import { USER_INFO_QUERY_KEY } from "serviceFacades/filesystem";
@@ -68,7 +64,6 @@ function Sharing(props) {
     const [permissions, setPermissions] = useState([]);
     const [originalUsers, setOriginalUsers] = useState({});
     const [userIdList, setUserIdList] = useState([]);
-    const [hasData, setHasData] = useState(false);
     const [userMap, setUserMap] = useState({});
     const [resourceTotal, setResourceTotal] = useState(0);
     const [errorDialogOpen, setErrorDialogOpen] = useState(false);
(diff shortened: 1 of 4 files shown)


ianmcorvidae:
Right now this PR basically works (as long as the group has been propagated into irods -- if not, it throws an error you have to lok in the network tools to diagnose), it's just... ugly.

src/components/sharing/index.js
```diff
@@ -220,7 +196,17 @@ function Sharing(props) {
...
+                        error: {
+                            response: {
+                                data: { error_code: code, reason: rest },
+                            },
+                        },
+                    });
+                } else if (failures.length > 0) {
                     setErrorDetails({
                         message: tSharing("sharingUnsuccessful"),
                         error: failures,
```
psarando, line 212: May as well fix the error reporting here too?
```suggestion
                        error: { response: { data: failures } },
```
Might be kind of ugly for now, but it is only for the error details dialog, and I think a JSON string that users can report to support / intercom is better than checking the logs? 🤷

src/components/sharing/index.js
```diff
@@ -220,7 +196,17 @@ function Sharing(props) {
...
+                if (failures.length === 1) {
+                    const { error_code: code, ...rest } = failures[0].error;
+                    setErrorDetails({
+                        message: tSharing("sharingUnsuccessful"),
+                        error: {
+                            response: {
+                                data: { error_code: code, reason: rest },
+                            },
+                        },
+                    });
```
psarando, line 208: 👍

src/components/sharing/util.js
```diff
@@ -85,8 +96,17 @@ export const getSharingFns = (type) => {
...
+                        id: user.id,
+                    };
+                    if (user.source_id) {
+                        subject.source_id = user.source_id;
+                    } else {
+                        subject = irodsNameToSubject(user.id);
+                    }
                     return {
-                        user: user.id,
+                        subject: subject,
```
psarando, line 109: FYI, this is some equivalent syntactic sugar:
```suggestion
                        subject,
```

psarando (approved):
LGTM 👍

src/components/sharing/index.js
ianmcorvidae: Hm, I think that wouldn't do anything as-is, it'd just produce a mostly-empty error dialog (or possibly an error, since it might try to use the array as an object? not sure). I guess I could dump everything into `response.data.reason`, though, at least. I honestly don't have any idea how often there's more than one failure at this stage anyway, but you're right it's at least worth showing _something_

src/components/sharing/util.js
ianmcorvidae: Good point. I think I was constructing it down here before and eventually rearranged it how this is, but didn't update it for the cleaner syntax.

src/components/sharing/index.js
psarando: Oops, that's right, I keep forgetting that it would have to be set inside `response.data.reason` or `response.data.message` 👍 

The only time I think there would be more than 1 failure is if sharing with multiple users, and more than 1 of those users had a sharing error.

ianmcorvidae:
I think I'm going to merge this for now and probably open another PR for other changes (I'll probably make the change to dump the error JSON into the reason field directly). Analysis sharing works great, and data sharing is good except a.) if the group isn't propagated to irods, it'll throw an error, and b.) the notification says `@grouper-.....`, but that's from terrain and not here anyway.

PR #530, opened 2023-04-12
Addons stories

sboleyn:
Add empty listing handling for add-ons listing and created stories for:

- Available add-ons listing
- Create/edit/delete subscription add-on
- Small update to the subscription listing story


psarando (approved):
LGTM :shipit: 👍 

Also keep in mind that some of these story tests can be imported and used as Jest tests too, which can ensure that the components will at least do an initial render without errors.
For examples, see `src/__tests__/tools.js`.
But this is not required for this PR.

PR #531, opened 2023-04-13
Display info

sboleyn:
Add subscription details, "Pro until 2023-04-13",  to the user menu


src/components/layout/UserMenu.js
```diff
@@ -61,13 +63,21 @@ export default function UserMenu(props) {
...
     const { t } = useTranslation(["common"]);
     const theme = useTheme();
     const classes = useStyles();
+    const formattedEndDate = formatDateObject(
+        new Date(subscription?.effective_end_date),
+        dateConstants.DATE_FORMAT
+    );
+    const subscriptionInfo = `${subscription?.plan?.name} ${t(
+        "until"
+    ).toLowerCase()} ${formattedEndDate}`;
```
psarando, line 80: I think this would be better formatted by the i18n tools, otherwise there's potential conflict in other languages for the proper ordering or word choice in this message.

If this i18n message was updated to take params like this:
```json
    "until": "{{planName}} until {{endDate}}",
```
then this could become something like this:
```suggestion
    const subscriptionInfo = t("until", {
        planName: subscription?.plan?.name,
        endDate: formattedEndDate
    });
```
Of course, naming this message key as just `util` would no longer make sense, so it should be renamed and maybe moved into the `subscriptions.json` messages.

psarando (changes requested):

src/components/layout/UserMenu.js
sboleyn: Thank you! Yeah I wasn't sure how to handle this one. I like your solution.

public/static/locales/en/common.json
```diff
@@ -100,4 +101,4 @@
     "view": "View",
     "viewAllNotifications": "View All Notifications & Activities",
     "viewOutput": "View output"
-}
+}
```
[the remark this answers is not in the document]
sboleyn: Thanks for catching this.

src/components/layout/UserMenu.js
psarando: No problem!

Something else I thought of this morning: Maybe this message shouldn't get moved into the `subscriptions.json` messages, because this `UserMenu` is displayed on every page, so those messages would then have to be required and downloaded for every page. So maybe it's best to leave it in `common.json` after all.

psarando (approved):
LGTM :shipit: 👍 🎉

src/components/layout/UserMenu.js
sboleyn: I'm just seeing this (I'm not sure why I don't get notifications of these comments). I will make a note and circle back on this.

psarando: It looks like you already did it though 😄 
https://github.com/cyverse-de/sonora/pull/531/commits/660a3db847fc982ee8c8cef80313c4a3b3b592dc

PR #532, opened 2023-04-14
CORE-1893 Update data nav to use paths from data info service

psarando:
Updated `components/data/toolbar/Navigation` to use the trash and home paths returned by the data info service, rather than relying solely on config values.

Added some hook functions to `components/data/utils` to assist with looking up the home and trash paths, but falling back to config values if the `bootstrap` or `roots` endpoints have not yet populated these paths (for logged out users, for example).

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -45,7 +45,7 @@ analysis:
 
 irods:
     home_path: /iplant/home
-    trash_path: /iplant/trash/home/de-irods
+    trash_path: /iplant/trash/home
     community_path: /iplant/home/shared
     web_dav:
         anon_uri: https://data.cyverse.org/dav-anon  
--- a/src/components/data/toolbar/Navigation.js
+++ b/src/components/data/toolbar/Navigation.js
@@ -10,12 +10,20 @@ import React, { useEffect, useState } from "react";
 
 import styles from "../styles";
 import ids from "../ids";
+import {
+    useBaseHomePath,
+    useBaseTrashPath,
+    useUserHomePath,
+    useTrashPath,
+} from "../utils";
+
 import constants from "../../../constants";
 import {
     getFilesystemRoots,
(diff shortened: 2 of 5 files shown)


config/default.yaml
```diff
@@ -45,7 +45,7 @@ analysis:
 
 irods:
     home_path: /iplant/home
-    trash_path: /iplant/trash/home/de-irods
+    trash_path: /iplant/trash/home
```
psarando, line 48: I also updated the default config to match the new value from the data-info service.

src/components/data/toolbar/Navigation.js
```diff
@@ -343,37 +353,63 @@ function Navigation(props) {
...
     const preProcessData = (respData) => {
         if (respData) {
-            const respRoots = respData.roots;
-            const home = userProfile
-                ? respRoots.find((root) => root.label === userProfile.id)
-                : {
-                      label: t("home"),
-                      path: `${irodsHomePath}/${constants.ANONYMOUS_USER}`,
-                  };
-            home.icon = <HomeIcon fontSize="small" />;
```
psarando, line 353: Before I settled on creating the new hook functions for the base paths, I started refactoring this code, but I left some of this refactoring in this PR, since it didn't seem prudent to directly modify and use the objects returned by the service.

So now this function creates new objects with only the values this component needs from the service.

I also didn't see the point in constructing this anonymous user's home path here if `getFilesystemRoots` is already constructing it and returning it for logged out users.

psarando (commented):
I also labeled this PR with `bug` because the data nav toolbar buttons and menu weren't updating with Trash actions when the data-info service returned `/iplant/trash/home` as the base trash path, but the UI's configs were still using `/iplant/trash/home/de-irods`.

slr71 (approved):
I didn't catch any issues. 👍

psarando:
Thanks for the review! 👍

PR #533, opened 2023-04-26
CORE-1894 disable intercom unless valid subscription

sboleyn:
This should be a valid solution unless the server relies on the intercom.enabled for things other than the application. Please let me know your thoughts.

### DIFF
--- a/src/components/layout/AppBar.js
+++ b/src/components/layout/AppBar.js
@@ -104,7 +104,6 @@ function DEAppBar(props) {
         children,
         activeView,
         setAppBarRef,
-        updatePlanAndSubscriptionState,
         intercomUnreadCount,
         clientConfig,
         showErrorAnnouncer,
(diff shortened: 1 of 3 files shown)


sboleyn:
fullWidth
mobile

src/components/layout/AppBar.js
```diff
@@ -165,10 +165,17 @@ function DEAppBar(props) {
 
     useEffect(() => {
         if (clientConfig) {
+            // Enable Intercom for non-basic subscriptions
+            if (userSubscription) {
+                let isBasic =
+                    userSubscription.result?.plan?.name?.toLowerCase() ===
+                    "basic";
+                clientConfig.intercom.enabled = !isBasic;
+            }
```
psarando, line 174: It doesn't seem like Intercom will work for non-basic users if we disable the value in the config file by default, since `pages/_app.js` does the initial setup of Intercom using only the config value read from the file, as returned by the server, and not by the value set in the client-side configs:
https://github.com/cyverse-de/sonora/blob/0bbdb1bb9c8084f793cf145a734cfe2d06d65004/src/pages/_app.js#L235-L245

Also, maybe intercom should be disabled if the user subscription status couldn't be loaded for some reason?

Finally, I don't think it's a good idea to directly mutate the `clientConfig` state like this, so maybe disabling intercom for non-basic plan names should be done in [pages/_app.js where the client config is setup from the config file values](https://github.com/cyverse-de/sonora/blob/0bbdb1bb9c8084f793cf145a734cfe2d06d65004/src/pages/_app.js#L140) instead, and also only if the plan name has also loaded?

So maybe something like this:
```javascript
            // Disable Intercom for non-basic subscriptions
            const planName = userSubscription.result?.plan?.name;
            const isBasic = planName?.toLowerCase() === "basic";
            intercom.enabled = publicRuntimeConfig.INTERCOM_ENABLED && planName && !isBasic;
```

psarando (changes requested):
I have tried to test this out locally, but I think I may have found a bug for non-basic users.

src/components/layout/AppBar.js
sboleyn: Shucks, I missed this earlier. I used a state updater function to pass the subscription information back to the page. Please let me know your thoughts.

psarando (approved):
LGTM 👍

PR #534, opened 2023-05-03
move SimpleExpansionPanel to common utilities

sboleyn:
Lots of components that are not under components/tools/ are using that SimpleExpansionPanel. It should probably be moved into somewhere more common like components/utils/.

### DIFF
--- a/src/components/utils/SimpleExpansionPanel.js
+++ b/src/components/utils/SimpleExpansionPanel.js
@@ -1,6 +1,6 @@
 /**
  *
- * A MUI accordinon that allows for displaying and
+ * A MUI accordion that allows for displaying and
  * hiding additional content when clicked.
  *
  */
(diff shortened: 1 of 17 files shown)


slr71 (approved):
I didn't notice any problems. 👍

src/components/utils/SimpleExpansionPanel.js
```diff
@@ -1,7 +1,13 @@
+/**
+ *
+ * A MUI accordinon that allows for displaying and
```
psarando, line 3:
```suggestion
 * A MUI accordion that allows for displaying and
```

psarando (approved):
LGTM 👍 ✨

src/components/utils/SimpleExpansionPanel.js
sboleyn: Good catch!

PR #539, opened 2023-06-17
Improve some stories for better Chromatic snapshots

psarando:
* Increased the chromatic delay for some stories so they finish loading before Chromatic takes the snapshot.
* Added static dates to the Timestamp Form Field and admin Instant Launch List stories.
* Added `disableSnapshot` to the VICE Loading story, since the animation usually changes each snapshot.


psarando:
I'll merge this now that the Chromatic stories look good, but reviews still welcome 👍

PR #542, opened 2023-07-17
CORE-1913 Add running count to Analyses nav icon

psarando:
This PR will add a running analyses count to the Analyses nav icon in the App Bar's Drawer items.

It also updates the AppBar to use the analyses stats query, which becomes invalidated when notifications arrive for running, canceled, or completed analyses, so that the running analyses count automatically updates based on these notifications.

Dashboard - 1 Running Analysis

Analyses Listing - 2 Running Analyses

---

With the App Bar Drawer expanded to display the icon labels:
App Bar Drawer expanded


slr71 (approved):
Looks good to me! :shipit:

psarando:
Thanks for the review! 👍

PR #546, opened 2023-08-17
CORE-1950 Pro user cannot see Intercom widget

sboleyn:
Remove admin endpoint when requesting user subscription details


psarando (approved):
LGTM 🎉 :shipit: 👍

PR #547, opened 2023-10-02
CORE 1961

sboleyn:
Add configuration setting that can be used to disable data search in the DE
Remove data search features from the DE when data search is disabled
-Remove data filter
-Remove data tab and data tab panel
-Disable data search queries
-Set default tab to 'apps' when clicking enter to search on globalsearchfield

### DIFF
--- a/src/components/search/GlobalSearchField.js
+++ b/src/components/search/GlobalSearchField.js
@@ -362,6 +362,8 @@ function GlobalSearchField(props) {
     const [bootstrapQueryEnabled, setBootstrapQueryEnabled] = useState(false);
     const [userHomeDir, setUserHomeDir] = useState(null);
 
+    const [defaultTab, setDefaultTab] = useState(null);
+
     const theme = useTheme();
     const isMobile = useMediaQuery(theme.breakpoints.down("xs"));
 
@@ -433,6 +435,14 @@ function GlobalSearchField(props) {
     useEffect(() => {
         setFilter(selectedFilter);
     }, [selectedFilter]);
+
+    useEffect(() => {
+        setDefaultTab(
+            config?.elasticEnabled
+                ? SEARCH_RESULTS_TABS.data
+                : SEARCH_RESULTS_TABS.apps
+        );
+    }, [config]);
 
     const { isFetching: searchingAnalyses, error: analysesSearchError } =
         useAnalysesSearch(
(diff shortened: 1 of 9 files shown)


sboleyn:
Screen Shot 2023-10-03 at 2 10 15 PM
Screen Shot 2023-10-03 at 2 10 44 PM
Screen Shot 2023-10-03 at 2 11 04 PM

src/components/search/GlobalSearchField.js
```diff
@@ -429,6 +436,14 @@ function GlobalSearchField(props) {
         setFilter(selectedFilter);
     }, [selectedFilter]);
 
+    useEffect(() => {
+        setDefaultTab(
+            config?.elasticEnabled
+                ? SEARCH_RESULTS_TABS.data
+                : SEARCH_RESULTS_TABS.apps
+        );
+    }, [config]);
```
psarando, line 445: I don't think this effect and the `defaultTab` state is required.
It should be enough to replace the state and this effect with a `const`:
```jsx
const defaultTab = config?.elasticEnabled
    ? SEARCH_RESULTS_TABS.data
    : SEARCH_RESULTS_TABS.apps;
```
Since this component is calling the `useConfig` hook, then it should re-render with any `config` updates, which should also update this `defaultTab` const.

For reference, see https://react.dev/learn/you-might-not-need-an-effect

src/pages/search.js
```diff
@@ -7,12 +7,22 @@ import GlobalSearchField from "components/search/GlobalSearchField";
...
         router?.query;
-    let tab = selectedTab || SEARCH_RESULTS_TABS.data;
+    useEffect(() => {
+        setDefaultTab(
+            config?.elasticEnabled
+                ? SEARCH_RESULTS_TABS.data
+                : SEARCH_RESULTS_TABS.apps
+        );
+    }, [config]);
+    let tab = selectedTab || defaultTab;
```
psarando, line 25: This `defaultTab` state and effect could also be replaced with a `const`:
```suggestion
    const defaultTab = config?.elasticEnabled
        ? SEARCH_RESULTS_TABS.data
        : SEARCH_RESULTS_TABS.apps;
    let tab = selectedTab || defaultTab;
```

psarando (changes requested):
Overall this LGTM, but we should see if this will still work by using `defaultTab` consts rather than state and useEffects.

src/components/search/GlobalSearchField.js
sboleyn: Thank you for this reference. I was debating whether this should be an effect and this guide will be helpful.

psarando (approved):
LGTM! 👍

PR #549, opened 2023-10-14
CORE-1952 Display limit check errors from instant launch actions

psarando:
This PR will update the `InstantLaunchButtonWrapper` to check for error codes related to VICE permission or limit errors, and to display the same related dialog as displayed from the `AppName` in the apps listing.

Since `POST /analyses` does not return `limitChecks` info in errors, I had to refactor the `instantlyLaunch` service facade to parse `limitChecks` info from the `/quicklaunches/{id}/app-info` call it already makes, and to reject the analysis submission if the `canRun` flag is set to `false`.


src/serviceFacades/instantlaunches.js
```diff
@@ -358,14 +356,29 @@ export const instantlyLaunch = ({ instantLaunch, resource, output_dir }) => {
     // Contains the Promises that resolve to the data needed to perform a job submission.
     const promiseList = [
         savedLaunchPromise,
-        getAppInfo({ launchId: savedLaunchId }).catch((e) => console.log(e)),
+        getAppInfo({ launchId: savedLaunchId }),
```
psarando, line 359: Not sure why the errors from these promises were getting caught and ignored, but I removed those catches so that errors from those endpoints can be properly displayed in the UI (rather than causing undefined `ql` or `app` exceptions below).

src/components/vice/RunError.js
```diff
@@ -1,15 +1,23 @@
...
 
+const SupportLink = (props) => (
+    <Link
+        variant="body1"
+        component="button"
+        onClick={intercomShow}
+        {...props}
+    />
+);
+
```
slr71, line 20: 👍

slr71 (approved):
Looks good to me! 👍

psarando:
Thanks for the review! 👍

PR #554, opened 2023-11-03
MUI V5 Migration

sboleyn:
The majority of changes are relating to the [MUI 5 migration](https://mui.com/material-ui/migration/migration-v4/) and cover Breaking changes parts 1 and 2 of the migration documentation. A fix to a few random console errors were also wrapped into these changes.

PR #557, opened 2023-11-22
CORE-1936 MUI v5 fixes

psarando:
This PR will add some fixes for changes introduced by MUI v5 added in #554.

* Updated dialog action buttons to `contained` variants, since the `Button` component now uses the `primary` color by default, and most of our dialog primary action buttons now looked the same as "cancel" buttons. The `contained` variant should also help the primary action to stand out more.
* Updated data selection drawer `zIndex` to `modal`.
  * The data selection drawer `zIndex` was less than the `zIndex` for dialogs in MUI v5, so it was hidden behind dialogs like the data move or copy metadata dialogs.

This PR also includes some fixes not related to the MUI v5 upgrade, but are for bugs that I came across during testing; such as a `startingPath` warning in the `PathListViewer`, and a `client-side exception has occurred` error that can display after launching analyses and navigating away and then back to the dashboard.

See visual changes in [Chromatic](https://www.chromatic.com/build?appId=5ebad5c6b991d60022b1e72e&number=2653).

### DIFF
--- a/src/components/analyses/AnalysisCommentDialog.js
+++ b/src/components/analyses/AnalysisCommentDialog.js
@@ -105,7 +105,7 @@ function AnalysisCommentDialog(props) {
                                 </Button>
                                 <Button
                                     id={buildID(baseId, UtilIds.DIALOG.CONFIRM)}
-                                    color="primary"
+                                    variant="contained"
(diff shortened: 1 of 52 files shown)


psarando:

PR #561, opened 2024-01-03
CORE-1971: Add the include-defaults flag to the endpoint for 'getToolDetails'.

ianmcorvidae:
defaults to include-defaults=false since most places that should include them aren't via this endpoint.

I'm not sure if there's more I'll need to do in this service, but it depends on cyverse-de/apps#266 and the corresponding common-swagger-api and terrain PRs anyway.


slr71 (approved):
Looks good to me! 👍

PR #563, opened 2024-01-27
CORE-1967 Initial Local Contexts label display

psarando:
This PR will add the initial Local Contexts label display to Data Listings and Metadata Templates.

When the metadata for a folder contains a `LocalContexts` attribute, and a `rightsURI` sub-attribute with a valid Local Contexts project URI as its value, then the Local Contexts labels for that project will be displayed in the data listing, as well as in metadata templates (such as the DataCite 4.2 template) that have a `LocalContexts` attribute field.

Each label is also an icon button that opens a dialog which displays the label or notice's descriptive text, along with a larger image of the label, and a link to the Local Contexts project page.

This PR also adds a custom `LocalContexts` attribute field for metadata templates (such as the DataCite 4.2 template) which displays only a Local Contexts Hub project URI field for the user, and will auto-populate DataCite metadata and child AVUs using the project ID parsed from the URI and the Local Contexts Hub API response (which is also used to display the project's labels below the field). 

---

DataCite Metadata Template

---

Data Listing

---

It's also possible for Local Contexts Hub projects to have sub-projects.

This can also be supported in the DE if the user adds the sub-project's URI to the associated data in the main project's folder:

Data Listing folder contents

---

Finally, this is an example Dialog displaying a project's TK Label and associated text:
TK Label Dialog

---

### DIFF
--- a/src/components/metadata/templates/LocalContextsField.js
+++ b/src/components/metadata/templates/LocalContextsField.js
@@ -87,6 +87,8 @@ const LocalContextsField = ({
         return schemeURI;
     });
 
+    const { touched, errors } = form;
+    const fieldError = getFormError(field.name, touched, errors);
     const projectID = parseProjectID(projectHubURI);
 
     const { data: project, isFetching } = useQuery({
@@ -95,7 +97,7 @@ const LocalContextsField = ({
             getLocalContextsProject({
                 projectID,
             }),
-        enabled: URL.canParse(projectHubURI),
+        enabled: projectHubURI && !fieldError,
         onSuccess: (project) => {
             let newValue = avu.value || "";
 
@@ -206,9 +208,7 @@ const LocalContextsField = ({
         }
     };
 
-    const { touched, errors } = form;
-    const errorMsg =
-        getFormError(field.name, touched, errors) || projectHubError;
+    const errorMsg = fieldError || projectHubError;
 
     return (
         <>
--- a/src/components/metadata/templates/LocalContextsField.js
+++ b/src/components/metadata/templates/LocalContextsField.js
@@ -1,0 +1,241 @@
+/**
+ * This custom metadata template field displays only a Local Contexts Hub
+ * project URL field, and auto-populates DataCite metadata and child AVUs using
+ * the project ID parsed from the URL and the Local Contexts Hub API response.
+ *
+ * @author psarando
+ */
+import React from "react";
+
+import { useQuery } from "react-query";
+
+import { useTranslation } from "i18n";
+
+import LocalContextsLabelDisplay from "../LocalContextsLabelDisplay";
+
+import ErrorTypographyWithDialog from "components/error/ErrorTypographyWithDialog";
+import getFormError from "components/forms/getFormError";
+import {
+    LocalContextsAttrs,
+    parseProjectID,
+} from "components/models/metadata/LocalContexts";
+
+import {
+    LOCAL_CONTEXTS_QUERY_KEY,
+    getLocalContextsProject,
+} from "serviceFacades/metadata";
+
+import { Skeleton, TextField } from "@mui/material";
+
+const findAVU = (avus, attr) => avus?.find((avu) => avu.attr === attr);
+
+const LocalContextsField = ({
+    avu,
+    onUpdate,
+    helperText,
+    form: { setFieldValue, ...form },
+    field: { value, onChange, ...field },
+    ...props
+}) => {
+    const { t } = useTranslation("localcontexts");
+
+    const [rightsURIAVU, setRightsURIAVU] = React.useState(
+        () =>
+            findAVU(avu.avus, LocalContextsAttrs.RIGHTS_URI) || {
+                attr: LocalContextsAttrs.RIGHTS_URI,
+                value: "",
+                unit: "",
+            }
+    );
+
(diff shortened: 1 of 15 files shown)


src/components/metadata/templates/LocalContextsField.js
```diff
@@ -0,0 +1,241 @@
...
+                    ...missingLabels.map((label) => ({
+                        attr: LocalContextsAttrs.RIGHTS_ID,
+                        value: label,
+                        unit: "",
+                    })),
+                ];
+
+                onUpdate({ ...avu, value: newValue, avus: newAVUs });
+            }
+        },
```
psarando, line 168: This logic in the Local Contexts Hub project query's `onSuccess` function synchronizes the labels defined in the project with the DE metadata. It also ensures the other child AVUs required by DataCite v4.2 are included.

The top-level AVU's value is based on suggestions from [DataCite's docs](https://support.datacite.org/docs/local-contexts-notices-and-labels).

The required AVU and child AVUs should be like the following:
```json
{
      "attr": "LocalContexts",
      "value": "The \"January Test\" project has Labels and/or Notices applied through the Local Contexts Hub. For more information, refer to the project page: https://sandbox.localcontextshub.org/projects/adcbca87-081a-4e2c-b6af-bff12fe3b7b0",
      "unit": "",
      "avus": [
        {
          "attr": "rightsIdentifier",
          "value": "Traditional Knowledge (TK) Notice",
          "unit": ""
        },
        {
          "attr": "rightsIdentifier",
          "value": "Biocultural (BC) Notice",
          "unit": ""
        },
        {
          "attr": "rightsIdentifier",
          "value": "Attribution Incomplete Notice",
          "unit": ""
        },
        {
          "attr": "rightsURI",
          "value": "https://sandbox.localcontextshub.org/projects/adcbca87-081a-4e2c-b6af-bff12fe3b7b0/",
          "unit": ""
        },
        {
          "attr": "schemeURI",
          "value": "https://localcontexts.org",
          "unit": ""
        },
        {
          "attr": "rightsIdentifierScheme",
          "value": "Local Contexts",
          "unit": ""
        }
      ]
    }
```

src/components/metadata/templates/LocalContextsField.js
```diff
@@ -0,0 +1,241 @@
...
+            newAVUs = [
+                ...newAVUs,
+                newRightsURIAVU,
+                schemeURIAVU,
+                rightsIDSchemeURIAVU,
+            ];
+
+            onUpdate({ ...avu, avus: newAVUs });
+        }
+    };
```
psarando, line 209: This function is called whenever the user updates the Local Contexts Hub project URI in the form field, and if the URI does not match the `rightsURI` child AVU's value, then it is updated, and the other child AVUs required by DataCite v4.2 are also included.

src/server/api/metadata.js
```diff
@@ -119,5 +119,15 @@ export default function metadataRouter() {
         })
     );
 
+    logger.info("adding the GET /api/unified-astronomy-thesaurus handler");
```
ianmcorvidae, line 122: This log message isn't terribly important but isn't accurate and should probably get updated

ianmcorvidae (approved):
I saw one small thing but as far as I understood the rest, looks fine to me. The only other thought I had, which it's possible is already in place since I don't think it'd be in this PR anyway, is that given the LocalContexts project URI acts as a required field, I don't know if the template should have a note saying something like "if LocalContexts doesn't apply, use <a different template, I guess just regular DataCite>" just to avoid confusion if someone selects the LocalContexts one in error.

slr71 (approved):
I didn't catch anything. :shipi

psarando:
> The only other thought I had, which it's possible is already in place since I don't think it'd be in this PR anyway, is that given the LocalContexts project URI acts as a required field, I don't know if the template should have a note saying something like "if LocalContexts doesn't apply, use <a different template, I guess just regular DataCite>" just to avoid confusion if someone selects the LocalContexts one in error.

So there is a little bit of trickery happening with this custom field in the DataCite template, because the top-level `LocalContexts` attribute is actually not marked as required, so it is not included by default and the user is not required to add it in their DataCite metadata:
custom LocalContexts attribute template default

If the user does add the `LocalContexts` attribute in the template, then it displays the field as required, but it's based on the fact that the `rightsURI` sub-attribute is required in DataCite metadata, so the custom field is borrowing the validations of that `rightsURI` sub-attribute.

Here's how it currently looks in prod without this custom component, when a user initially adds the attribute in the template:
prod LocalContexts template attribute

The template still defines this attribute with all these sub-attributes, it's just the UI is displaying the simplified custom component for the `LocalContexts` attribute; and this custom component will be displayed in any metadata template that has a `LocalContexts` attribute, not just the DataCite template.

This also means all the required sub-attributes will still be included in a CSV download of the template for bulk metadata uploads.

We can also add another metadata template that only has the `LocalContexts` attribute, for users that only need labels and notices to display for their data, but because the custom component will and all the sub-attributes required by the DataCite template, the user can still use that template later if they need a DOI for their data, and those required attributes will already be there.

I was also thinking the template's description for this attribute could be updated once these changes go live in prod, so it will make sense in both the custom component in the UI, but also in the CSV guide download.

psarando:
Thanks for the reviews 👍

PR #567, opened 2024-03-12
CORE-1967 Allow anonymous access in /filesystem/metadata endpoints

psarando:
This PR removes the `/secured` prefix for `/filesystem/metadata` endpoints (updated by cyverse-de/terrain#281).

This is required in order for Local Contexts Labels and Notices to be viewed by logged out users for public data under Community Data folders.


psarando:
I'm going to merge this now, but reviews are still welcome 👍

PR #570, opened 2024-04-08
CORE-1183: async path list creation

ianmcorvidae:
This is super rudimentary and probably needs a bit more but I _think_ this is all that's needed to get it to basically not crash? anyway putting it up for now

### DIFF
--- a/src/components/data/toolbar/DataDotMenu.js
+++ b/src/components/data/toolbar/DataDotMenu.js
@@ -458,7 +458,9 @@ function DataDotMenu(props) {
                 onClose={() => setPathListDlgOpen(false)}
                 onCreatePathList={(id, path) => {
                     setPathListDlgOpen(false);
-                    routeToFile(id, path);
+                    if (id && path) {
+                        routeToFile(id, path);
+                    }
                 }}
                 onCancel={() => setPathListDlgOpen(false)}
                 startingPath={path}


src/components/data/toolbar/DataDotMenu.js
```diff
@@ -458,7 +458,9 @@ function DataDotMenu(props) {
                 onClose={() => setPathListDlgOpen(false)}
                 onCreatePathList={(id, path) => {
                     setPathListDlgOpen(false);
-                    routeToFile(id, path);
+                    if (id && path) {
```
psarando, line 461: Also, at the very least, the following line should be updated to check for undefined `path` and `id` as well:
https://github.com/cyverse-de/sonora/blob/d714dd6f09347280a868e925a24819426e25a08a/src/components/data/PathListAutomation.js#L103
Such as `onCreatePathList(data?.file?.id, data?.file?.path);`

We'll probably also want to display an alert that the request is processing, and the user will get a notification when it's complete. Like we do when moving files/folders, for example:
https://github.com/cyverse-de/sonora/blob/d714dd6f09347280a868e925a24819426e25a08a/src/components/data/MoveDialog.js#L48-L57

ianmcorvidae: This is good and I really appreciate the pointers to how to change some of these things since I haven't been working with the UI that much :sweat_smile: 

I think I've fixed both of these things

psarando (approved):
LGTM 👍

PR #572, opened 2024-06-04
CORE-1998 Remove User Survey from Dashboard

psarando:
This PR will remove the User Survey from the Dashboard, which is no longer needed.


psarando:

PR #573, opened 2024-06-05
CORE-1999: fix the matching file count when no matching files are found

slr71:
Prior to this change search results when there were no matches showed the wrong result count:

image

After this change, the results display the correct count:

image


ianmcorvidae:
Looks reasonable to me!

psarando (approved):
LGTM :shipit: 👍

slr71:
Thanks for the reviews!

PR #582, opened 2024-07-09
CORE-1889 Tapis v3 Migration

psarando:
This PR will rename `Agave` to `Tapis` just about everywhere, except there is still an `AppsTypeFilter` for `Agave`, only for the `Analyses` listing, so users can still list their old `Agave` jobs, even if they can't relaunch them anymore.

This PR also includes fixes for listing enum param default values in relaunch forms, and for refreshing the job history in the Analysis Landing page.

### DIFF
--- a/src/components/analyses/toolbar/Toolbar.js
+++ b/src/components/analyses/toolbar/Toolbar.js
@@ -14,7 +14,7 @@ import { allowAnalysesCancel } from "../utils";
 
 import { useConfig } from "contexts/config";
 
-import AppsTypeFilter from "components/apps/AppsTypeFilter";
+import AppsTypeFilter, { getFilters } from "components/apps/AppsTypeFilter";
 
 import buildID from "components/utils/DebugIDUtil";
 
@@ -171,6 +171,17 @@ function AnalysesToolbar(props) {
         hasSelection && allowAnalysesCancel(selectedAnalyses, username, config);
     const { isSmDown, isMdDown } = useBreakpoints();
 
+    const typeFilter = (
+        <AppsTypeFilter
+            baseId={analysesNavId}
+            filter={appTypeFilter}
+            classes={classes}
+            handleFilterChange={handleAppTypeFilterChange}
+            options={getFilters(true)}
+            disabled={false}
+        />
+    );
+
     return (
         <>
             <Toolbar variant="dense" id={analysesNavId} style={{ padding: 0 }}>
@@ -182,13 +193,7 @@ function AnalysesToolbar(props) {
                             classes={classes}
                             handleFilterChange={handleOwnershipFilterChange}
                         />
-                        <AppsTypeFilter
-                            baseId={analysesNavId}
-                            filter={appTypeFilter}
-                            classes={classes}
-                            handleFilterChange={handleAppTypeFilterChange}
-                            disabled={false}
-                        />
+                        {typeFilter}
                     </>
                 )}
                 {viewFiltered && (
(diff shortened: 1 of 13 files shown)


src/components/analyses/toolbar/Toolbar.js
```diff
@@ -182,13 +193,7 @@ function AnalysesToolbar(props) {
...
                             handleFilterChange={handleOwnershipFilterChange}
                         />
-                        <AppsTypeFilter
-                            baseId={analysesNavId}
-                            filter={appTypeFilter}
-                            classes={classes}
-                            handleFilterChange={handleAppTypeFilterChange}
-                            disabled={false}
-                        />
+                        {typeFilter}
```
slr71, line 196: ✨

src/components/analyses/toolbar/Toolbar.js
```diff
@@ -300,13 +305,7 @@ function AnalysesToolbar(props) {
...
                     />
                     <br />
-                    <AppsTypeFilter
-                        baseId={analysesNavId}
-                        filter={appTypeFilter}
-                        classes={classes}
-                        handleFilterChange={handleAppTypeFilterChange}
-                        disabled={false}
-                    />
+                    {typeFilter}
```
slr71, line 308: ✨

src/serviceFacades/apps.js
```diff
@@ -32,10 +31,7 @@ const ADMIN_APP_DETAILS_QUERY_KEY = "fetchAppDetailsForAdmin";
 const ADMIN_APP_AVU_QUERY_KEY = "fetchAppAVUs";
 
 const getAppTypeFilter = (appTypeFilter) => {
-    const typeFilter =
-        appTypeFilter && appTypeFilter !== appType.all
-            ? "&app-type=" + appTypeFilter
-            : "";
+    const typeFilter = appTypeFilter ? "&app-type=" + appTypeFilter : "";
```
slr71, line 34: :sparkles:

src/serviceFacades/apps.js
```diff
@@ -362,11 +358,7 @@ function getAppsForAdmin({
         params["search"] = searchTerm;
     }
 
-    if (appTypeFilter && appTypeFilter !== appType.all) {
-        params["app-type"] = appTypeFilter;
-    } else {
-        params["app-type"] = "";
-    }
+    params["app-type"] = appTypeFilter || "";
```
slr71, line 361: :sparkles:

slr71 (approved):
Looks good to me! 👍

psarando:
Note that this PR also updates the Analyses Listing filter:
Analyses Listing Filter - Tapis v3

src/serviceFacades/apps.js
```diff
@@ -32,10 +31,7 @@ const ADMIN_APP_DETAILS_QUERY_KEY = "fetchAppDetailsForAdmin";
 const ADMIN_APP_AVU_QUERY_KEY = "fetchAppAVUs";
 
 const getAppTypeFilter = (appTypeFilter) => {
-    const typeFilter =
-        appTypeFilter && appTypeFilter !== appType.all
```
ianmcorvidae, line 36: This is just me checking -- I guess this change to checking only if `appTypeFilter` is truthy works because `appType.all` isn't actually defined in `AppType.js`, so it's undefined or similar anyway?

(just want to make sure since it otherwise appears this is changing how this code works with that particular value)

ianmcorvidae (approved):
One minor confusion on my part which is probably fine, otherwise looks reasonable to me!

src/serviceFacades/apps.js
psarando: Yep! I discovered `appType.all` was removed in #143, so I cleaned this up since it's been `undefined` since that change.

psarando:
Thanks for the reviews! 👍

PR #583, opened 2024-07-30
Small update to use the _first_ @ rather than the last for username truncation

ianmcorvidae:
There's one other spot we're using the suffix, but it already removes all `@` signs (it splits on `@` and then takes the first returned component). I guess it won't do exactly what it should if for some reason the username suffix differs from the configured value, but I'm just not that worried about that. This change is (I think) only relevant to display, but still seemed better to be consistent.


slr71 (approved):
LGTM

PR #591, opened 2024-09-23
Set up preference for periodic email notifications & pass to app submissions

ianmcorvidae:
I think this should be ready to go -- adds the two preferences and passes them down to the apps service, assuming I haven't done anything dumb at least.

### DIFF
--- a/src/components/apps/launch/index.js
+++ b/src/components/apps/launch/index.js
@@ -81,6 +81,8 @@ const Launch = ({
 
     const preferences = bootstrapInfo?.preferences;
     const notify = preferences?.enableAnalysisEmailNotification || false;
+    const notifyPeriodic = preferences?.enablePeriodicEmailNotification || true;
+    const periodicPeriod = preferences?.periodicNotificationPeriod || 14400;
 
     const defaultMaxCPUCores = config?.tools?.private.max_cpu_limit;
     const defaultMaxMemory = config?.tools?.private.max_memory_limit;
@@ -140,6 +142,8 @@ const Launch = ({
         <AppLaunchWizard
             baseId={baseId}
             notify={notify}
+            notifyPeriodic={notifyPeriodic}
+            periodicPeriod={periodicPeriod}
             defaultOutputDir={defaultOutputDir}
             defaultMaxCPUCores={defaultMaxCPUCores}
             defaultMaxMemory={defaultMaxMemory}
--- a/src/components/preferences/General.js
+++ b/src/components/preferences/General.js
@@ -191,6 +191,24 @@ function General(props) {
                         label={t("urlImportEmailLbl")}
                     />
                 </Grid>
+                <Grid item>
+                    <Field
+                        id={buildID(
+                            baseId,
+                            ids.PERIODIC_EMAIL_NOTIFICATION_SWITCH
+                        )}
+                        component={FormSwitch}
+                        name={
+                            prefConstants.keys
+                                .ENABLE_PERIODIC_EMAIL_NOTIFICATION
+                        }
+                        color="primary"
+                        inputProps={{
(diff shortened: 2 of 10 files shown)


slr71 (approved):
I didn't catch any issues. 👍

ianmcorvidae:
Cool, thanks for the review!

src/components/apps/launch/index.js
```diff
@@ -81,6 +81,8 @@ const Launch = ({
 
     const preferences = bootstrapInfo?.preferences;
     const notify = preferences?.enableAnalysisEmailNotification || false;
+    const notifyPeriodic = preferences?.enablePeriodicEmailNotification || true;
```
psarando, line 84: This won't work if the user sets this preference to `false`, since the `||` here will set `notifyPeriodic` to `true` anyway.

You could coerce it into a boolean like this:
```javascript
const notifyPeriodic = !!preferences?.enablePeriodicEmailNotification;
```

src/components/preferences/General.js
```diff
@@ -191,6 +193,42 @@ function General(props) {
...
+                        }}
+                        label={t("periodicEmailLbl")}
+                    />
+                </Grid>
+                <Grid item xs={6}>
+                    <Field
+                        id={buildID(baseId, ids.PERIODIC_NOTIFICATION_PERIOD)}
+                        name={prefConstants.keys.PERIODIC_NOTIFICATION_PERIOD}
+                        label={t("periodicPeriodLbl")}
+                        component={FormSelectField}
```
psarando, line 219: I've been preferring `FormTextField` with the `select` prop, for a more consistent UX (such as using an outlined form field):
```javascript
                        component={FormTextField}
                        select
```

src/components/preferences/General.js
```diff
@@ -191,6 +193,42 @@ function General(props) {
...
+                    >
+                        <MenuItem key={1} value={3600}>
+                            1 hour
+                        </MenuItem>
+                        <MenuItem key={2} value={14400}>
+                            4 hours
+                        </MenuItem>
+                        <MenuItem key={3} value={86400}>
+                            24 hours
+                        </MenuItem>
```
psarando, line 229: These `key` props aren't needed unless these components are nested inside an array `[]`.

src/components/preferences/messages.js
```diff
@@ -3,6 +3,9 @@ export default {
     messages: {
         analysesViewShortcutLbl: "Open analyses view",
         analysisStatusEmailLbl: "Email me when my analysis status changes",
+        periodicEmailLbl: "Email me periodically when analyses are running",
+        periodicPeriodLbl:
+            "How frequently to notify about currently running jobs",
```
psarando, line 8: This is probably an obsolete file (I think all our messages should be in `public/static/locales/en/`), but we can verify that later.

psarando (commented):
Sorry I didn't get a chance to review this sooner.
I had a couple of minor style comments, but I think I also found 1 bug.

src/components/apps/launch/index.js
ianmcorvidae: Oh, good catch, yeah. I'll make & push that change.

src/components/preferences/General.js
ianmcorvidae: Ah, interesting. I saw that was in use one place and the select in a different one, and went with FormSelectField because it seemed like it should be more usual, but I'm good to change it as well, so I'll do that.

src/components/preferences/General.js
ianmcorvidae: Ah, right. I copied from somewhere else that was creating an array from another array, rather than a hardcoded list like this.

PR #592, opened 2024-09-27
Make periodic notifications work properly with instant launches

ianmcorvidae:
This is the one place that calls to the `instantlyLaunch` helper function, so it should be the only place that needs to get preferences via bootstrap info to pass to it. This should also mean that anyone who's set up email notifications or not will have them behave correctly with respect to email (in the case where their preference doesn't match the saved launch).

### DIFF
--- a/src/components/apps/launch/index.js
+++ b/src/components/apps/launch/index.js
@@ -80,7 +80,7 @@ const Launch = ({
     );
 
     const preferences = bootstrapInfo?.preferences;
-    const notify = preferences?.enableAnalysisEmailNotification || false;
+    const notify = !!preferences?.enableAnalysisEmailNotification;
     const notifyPeriodic = !!preferences?.enablePeriodicEmailNotification;
     const periodicPeriod = preferences?.periodicNotificationPeriod || 14400;
 
--- a/src/components/instantlaunches/InstantLaunchButtonWrapper.js
+++ b/src/components/instantlaunches/InstantLaunchButtonWrapper.js
@@ -60,54 +60,46 @@ function InstantLaunchButtonWrapper(props) {
         }
     }, [ilUrl]);
 
-    const { mutate: launch } = useMutation(
-        (args) => {
-            args.preferences = bootstrapInfo?.preferences;
-            return instantlyLaunch(args);
-        },
-        {
-            onSuccess: (listing) => {
-                if (listing.analyses.length > 0) {
-                    const analysis = listing.analyses[0];
-                    if (analysis.interactive_urls?.length > 0) {
-                        setIlUrl(
-                            `${
-                                constants.VICE_LOADING_PAGE
-                            }/${encodeURIComponent(
-                                analysis.interactive_urls[0]
-                            )}`
-                        );
-                    } else {
-                        setOpen(false);
-                    }
+    const { mutate: launch } = useMutation(instantlyLaunch, {
+        onSuccess: (listing) => {
+            if (listing.analyses.length > 0) {
+                const analysis = listing.analyses[0];
+                if (analysis.interactive_urls?.length > 0) {
+                    setIlUrl(
+                        `${constants.VICE_LOADING_PAGE}/${encodeURIComponent(
+                            analysis.interactive_urls[0]
+                        )}`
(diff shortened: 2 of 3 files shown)


ianmcorvidae:
Two things I'm not sure of here:

* should I be doing this within the `useMutation`, or should it go where the mutation is being called from?
* should I encapsulate the "grab the relevant preferences out of preferences" code somewhere in order to reuse it between here and the regular app launch? (if so, where would be good?)

src/serviceFacades/instantlaunches.js
```diff
@@ -436,6 +441,19 @@ export const instantlyLaunch = ({ instantLaunch, resource, output_dir }) => {
...
+            submission.notify = notify;
+            const notifyPeriodic =
+                !!preferences?.enablePeriodicEmailNotification;
+            submission.notify_periodic = notifyPeriodic;
+            const periodicPeriod =
+                preferences?.periodicNotificationPeriod || 14400;
+            submission.periodic_period = periodicPeriod;
+
+            return submission;
+        })
```
psarando, line 456: This works fine as-is, but here's another more compact way to modify this submission JSON:
```suggestion
        .then((submission) => ({
            ...submission,
            notify: !!preferences?.enableAnalysisEmailNotification,
            notify_periodic: !!preferences?.enablePeriodicEmailNotification,
            periodic_period: preferences?.periodicNotificationPeriod || 14400,
        }))
```

psarando (changes requested):
> should I encapsulate the "grab the relevant preferences out of preferences" code somewhere in order to reuse it between here and the regular app launch?

I think calling `useBootstrapInfo` from `InstantLaunchButtonWrapper` and passing the preferences into `instantlyLaunch` is fine, unless I'm not understanding the question?

src/serviceFacades/instantlaunches.js
ianmcorvidae: This is basically the code I was wondering if I should encapsulate somewhere else (since there is similar `const periodicPeriod = preferences?.periodicNotificationPeriod || 14400;` type code in the app launch code from the last PR) , but condensing it this way is better than doing that, I think.

src/components/instantlaunches/InstantLaunchButtonWrapper.js
```diff
@@ -58,46 +60,54 @@ function InstantLaunchButtonWrapper(props) {
...
-                    setIlUrl(
-                        `${constants.VICE_LOADING_PAGE}/${encodeURIComponent(
-                            analysis.interactive_urls[0]
-                        )}`
-                    );
+    const { mutate: launch } = useMutation(
+        (args) => {
+            args.preferences = bootstrapInfo?.preferences;
+            return instantlyLaunch(args);
+        },
```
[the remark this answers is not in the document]
ianmcorvidae: Yeah, that's probably better. I had initially not thought of the option of doing it that way for some reason, then realized when I was leaving my comment asking for feedback :sweat_smile:

ianmcorvidae:
Thanks for the review, good thoughts. Getting more comfortable with this UI stuff here, slowly, heh. I think I've addressed your comments, @psarando -- did one a little differently (easier to just do `bootstrapInfo.?preferences` right when calling the function, since it's only one place anyway) but I think it should be good.

src/components/apps/launch/index.js
```diff
@@ -80,7 +80,7 @@ const Launch = ({
     );
 
     const preferences = bootstrapInfo?.preferences;
-    const notify = preferences?.enableAnalysisEmailNotification || false;
```
ianmcorvidae, line 83: Updated this here to match the suggestion for the other spot that does this. Cleaner, I think.

src/serviceFacades/instantlaunches.js
psarando: I see now. That's a good instinct, but this is probably fine for now. If we need to grab these preferences with these defaults in a 3rd place, then it will definitely be time to refactor 😃

psarando (approved):

psarando (approved):
LGTM 👍 🎉

PR #593, opened 2024-09-30
CORE-2011 Update setting max CPUs in app launch form

psarando:
This PR will update setting the default selected max CPUs in the app launch form, and adds a `tools.default_selected_max_cpus` config to allow us to adjust the default selected value.

This default value will only be selected if it's less than the actual max set by the tool, and only if no `default_max_cpu_cores` is returned from the API (as is the case in job relaunches where a value was requested in the original submission).


ianmcorvidae (approved):
Looks good to me, I think! I think 4 is a good starting point value as well. (I guess that value is... the default configuration value for the default configuration used when there's no default send by the API? :dizzy_face: )

psarando:
Thanks for the review! 👍

PR #594, opened 2024-10-01
Use special text for periodic updates, and try more ways of finding the analysis ID for linking

ianmcorvidae:
cyverse-de/timelord#16 is what sets the analysis ID into the notification


psarando (approved):
I didn't catch any issues 👍

PR #595, opened 2024-10-02
Amend analysis page icons

ianmcorvidae:
Upon design conversations, "Share with Support" doesn't need prime placement on this page, because there's already a button for it in the top bar of every page. However, extend & terminate are more primary actions that should appear on large screens.

when big:
image

(with menu open)
image

when smaller, with menu open:
image

Thoughts/questions:

* Should the terminate icon (or the whole button, maybe) be changed to the red that's used for the small action buttons on the analysis overview
* Should "extend time limit" be distinguished at all by color, or any of the other (otherwise unchanged) buttons, especially if we change "terminate".

### DIFF
--- a/src/components/analyses/landing/AnalysisSubmissionLanding.js
+++ b/src/components/analyses/landing/AnalysisSubmissionLanding.js
@@ -431,7 +431,9 @@ export default function AnalysisSubmissionLanding(props) {
                                                 onClick={() => {
                                                     handleTerminateSelected();
                                                 }}
-                                                startIcon={<CancelIcon />}
+                                                startIcon={
+                                                    <CancelIcon color="error" />
+                                                }
                                             >
                                                 {!isSmDown && (
                                                     <>{t("terminate")}</>
--- a/src/components/analyses/landing/AnalysisSubmissionLanding.js
+++ b/src/components/analyses/landing/AnalysisSubmissionLanding.js
@@ -73,8 +73,9 @@ import {
     useTheme,
 } from "@mui/material";
 import {
-    ContactSupport,
+    Cancel as CancelIcon,
     ExpandMore,
+    HourglassEmptyRounded as HourGlassIcon,
     Refresh,
     Launch,
 } from "@mui/icons-material";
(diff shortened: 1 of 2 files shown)


psarando:
I haven't reviewed these code changes yet, but I wanted to address these comments and questions first:

> Upon design conversations, "Share with Support" doesn't need prime placement on this page, because there's already a button for it in the top bar of every page. However, extend & terminate are more primary actions that should appear on large screens.
> 

The "Request Help" button in the analysis landing page is different from the `?` icon in the app toolbar on every page, which is a "Chat with Support" button that will open Intercom chat, and it's only displayed for subscribers.

The "Request Help" button opens a dialog with troubleshooting steps (which depend on the current job status), and a confirmation button that will share that analysis, and its inputs and outputs, with the `siuser`:

---
Share with Support dialog
---

That said, I have no problem with still moving the "Request Help" button into the menu, but we should probably verify that with whoever else was involved in the design discussions.

> Thoughts/questions:
> 
> * Should the terminate icon (or the whole button, maybe) be changed to the red that's used for the small action buttons on the analysis overview
> 

I think the terminate icon should be red to match the icon in the analysis listing, and on the dashboard tiles.

> * Should "extend time limit" be distinguished at all by color, or any of the other (otherwise unchanged) buttons, especially if we change "terminate".

I don't think the time limit icon needs a different color. I see the red terminate icon as an exception.

ianmcorvidae:
That's a good point with the different contents. I asked Tyson/Nirav/Tina/etc. and the conclusion is that it seems good to put it in the dot menu. With these sorts of intercom support features moving more toward paid-only, having it a bit more hidden is okay as well, so users can be directed to it when needed/desired, from what Nirav said.

I'll change that terminate icon to the red, I agree there. I didn't think of it until I was posting the screenshots here, heh.

ianmcorvidae:
image
image

updated screenshots

src/components/analyses/landing/AnalysisSubmissionLanding.js
```diff
@@ -392,28 +393,54 @@ export default function AnalysisSubmissionLanding(props) {
...
+                                                color="primary"
+                                                onClick={() => {
+                                                    setConfirmExtendTimeLimitDlgOpen(
+                                                        true
+                                                    );
+                                                }}
+                                                startIcon={<HourGlassIcon />}
+                                            >
+                                                {!isSmDown && (
+                                                    <>{t("extendTime")}</>
```
psarando, line 415: This Fragment looks unnecessary, since `{!isSmDown && t("extendTime")}` should work the same.

I see it's already used like this in other places in this file, so it's fine to leave it like this for now just for consistency in this file.

psarando (approved):
LGTM 👍

src/components/analyses/landing/AnalysisSubmissionLanding.js
ianmcorvidae: Ha, yup, I just copied the others in the file assuming it was necessary. I'll leave it for now, but I'll keep it in mind as I make more UI changes (which it looks like I will be).

ianmcorvidae:
Thanks for the review! Been nice learning some of this UI stuff a bit better here.

PR #596, opened 2024-10-08
Instant Launch badges (WIP)

ianmcorvidae:
This (so far) adds a single new page at `/instantlaunches/:id`, which immediately launches the provided instant launch.

Continuing work:
* allow passing an input parameter as a query parameter, to be integrated into the submission
* something beyond an empty page (possibly just a redirect to the analysis landing after a job is launched)
* added info in the popup dialog that tells you what you're launching

### DIFF
--- a/src/components/instantlaunches/launch/index.js
+++ b/src/components/instantlaunches/launch/index.js
@@ -3,11 +3,89 @@
  *
  * The component that launches a provided instant launch, immediately.
  */
-import React from "react";
+import React, { useState } from "react";
+
+import { useQuery } from "react-query";
+
+import {
+    GET_INSTANT_LAUNCH_FULL_KEY,
+    getFullInstantLaunch,
+} from "serviceFacades/instantlaunches";
+import {
+    getResourceUsageSummary,
+    RESOURCE_USAGE_QUERY_KEY,
+} from "serviceFacades/dashboard";
+import { getUserQuota } from "common/resourceUsage";
+import { useConfig } from "contexts/config";
+import { useUserProfile } from "contexts/userProfile";
+import globalConstants from "../../../constants";
+import { useTranslation } from "i18n";
+import isQueryLoading from "components/utils/isQueryLoading";
+import InstantLaunchButtonWrapper from "components/instantlaunches/InstantLaunchButtonWrapper";
+import DELink from "components/utils/DELink";
+import withErrorAnnouncer from "components/error/withErrorAnnouncer";
 
 const InstantLaunchStandalone = (props) => {
-    const { id: instant_launch_id } = props;
-    return <p>itsa {instant_launch_id}</p>;
+    const { id: instant_launch_id, showErrorAnnouncer } = props;
+    const [config] = useConfig();
+    const [userProfile] = useUserProfile();
+    const [computeLimitExceeded, setComputeLimitExceeded] = useState(
+        !!config?.subscriptions?.enforce
+    );
+    const { t } = useTranslation(["instantlaunches", "common"]);
+
+    const { data, status, error } = useQuery(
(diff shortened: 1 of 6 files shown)


src/components/instantlaunches/launch/index.js
```diff
@@ -0,0 +1,81 @@
...
+
+    if (isLoading) {
+        return <p>... loading ...</p>;
+    } else if (error) {
+        return (
+            <p>
+                error{" "}
+                {error?.response?.status === 404 ? "not found" : "unsure why"}
+            </p>
+        );
```
ianmcorvidae, line 69: I should comment that I definitely want to change this before merging anything :laughing:

src/pages/instantlaunch/[id]/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+export default function InstantLaunch() {
+    const router = useRouter();
+    const { id } = router.query;
+
+    return <InstantLaunchStandalone id={id} />;
+}
+
+export async function getServerSideProps({ locale }) {
+    const title = i18n?.t("instantLaunches") || "";
```
Comment (psarando) [inline src/pages/instantlaunch/[id]/index.js:24]: As mentioned in slack, I'm not sure if this is an issue outside of dev mode (though it could still be an issue for the very first user to hit a non-root-page when the nextjs server first starts up), but I think the fall back string for this title can be the English string `"Instant Launches"` in this case, rather than `""`.

src/components/instantlaunches/launch/index.js
psarando: These could be replaced with `<Typography variant="caption"><Skeleton /></Typography>` and `ErrorTypography`, and of course translated text using `i18n` messages.

See `components/apps/launch/AppInfo` for one example.

In a future PR, maybe the loading message/skeleton could also be replaced with `components/vice/loading/LoadingAnimation` 🤔

src/pages/instantlaunch/[id]/index.js
Comment (ianmcorvidae) [reply src/pages/instantlaunch/[id]/index.js:24]: Yeah, that's a good point. I realize I should probably add a string for this page title specifically anyway, since the plural "Instant Launches" doesn't make that much sense for a single-instant-launch page anyway. I think you're right that this is basically only a dev-mode problem so I think I'll just remove the `?.` and fallback here to avoid duplicating the string.

src/components/instantlaunches/launch/index.js
ianmcorvidae: Cool, I'll look at that example for sure. This stuff responds quickly normally so the loading stuff is usually really fast, but may as well get it looking nice in some fashion. Errors definitely need to be prettier, since I have a feeling 404s might not be super uncommon for this situation, heh.

ianmcorvidae:
It's still a little rudimentary but it's not linked from anywhere yet and basically works so I think I'll merge this PR for now, and open more as I get more things done. Any further reviews/thoughts always welcome.

src/components/instantlaunches/launch/index.js
```diff
@@ -0,0 +1,93 @@
...
+                <ErrorTypography
+                    errorMessage={error.message}
+                    onDetailsClick={() => setErrorDialogOpen(true)}
+                />
+                <DEErrorDialog
+                    open={errorDialogOpen}
+                    errorObject={error}
+                    handleClose={() => setErrorDialogOpen(false)}
+                />
+            </>
```
psarando, line 80: This looks basic enough that this could be replaced with an `ErrorTypographyWithDialog`. 

See `components/apps/savedLaunch/SavedLaunchListing` for an example (I guess `AppInfo` wasn't the best example for `ErrorTypography`, which seems like it could also use an `ErrorTypographyWithDialog` instead).

psarando (approved):
LGTM 👍

PR #597, opened 2024-10-10
Support passing a resource path to the instant launch page

ianmcorvidae:
This was easier than expected. Grabs a stat if a path's passed in, shows an error if it gets one, passes along the data to the instant launch. Nice.

### DIFF
--- a/src/components/instantlaunches/launch/index.js
+++ b/src/components/instantlaunches/launch/index.js
@@ -15,6 +15,7 @@ import {
     getResourceUsageSummary,
     RESOURCE_USAGE_QUERY_KEY,
 } from "serviceFacades/dashboard";
+import { useDataDetails } from "serviceFacades/filesystem";
 import { getUserQuota } from "common/resourceUsage";
 import { useConfig } from "contexts/config";
 import { useUserProfile } from "contexts/userProfile";
@@ -29,13 +30,19 @@ import ErrorTypography from "components/error/ErrorTypography";
 import DEErrorDialog from "components/error/DEErrorDialog";
 
 const InstantLaunchStandalone = (props) => {
-    const { id: instant_launch_id, showErrorAnnouncer } = props;
+    const {
+        id: instant_launch_id,
+        resource: resource_path,
+        showErrorAnnouncer,
+    } = props;
     const [config] = useConfig();
     const [userProfile] = useUserProfile();
     const [computeLimitExceeded, setComputeLimitExceeded] = useState(
         !!config?.subscriptions?.enforce
     );
     const [errorDialogOpen, setErrorDialogOpen] = useState(false);
+    const [resource, setResource] = useState(!!resource_path ? null : {});
+
     const { t } = useTranslation(["instantlaunches", "common"]);
 
     const { data, status, error } = useQuery(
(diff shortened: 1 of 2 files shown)


src/components/instantlaunches/launch/index.js
```diff
@@ -29,13 +30,19 @@ import ErrorTypography from "components/error/ErrorTypography";
...
+        resource: resource_path,
+        showErrorAnnouncer,
+    } = props;
     const [config] = useConfig();
     const [userProfile] = useUserProfile();
     const [computeLimitExceeded, setComputeLimitExceeded] = useState(
         !!config?.subscriptions?.enforce
     );
     const [errorDialogOpen, setErrorDialogOpen] = useState(false);
+    const [resource, setResource] = useState(!!resource_path ? null : {});
```
ianmcorvidae, line 44: I'm unsure this is strictly necessary, could possibly just pass in `{}`. But I think it's more likely to error (as it should) if when a path is passed in it sets this to null.

psarando, line 44: Since the `InstantLaunchButtonWrapper` is already using `resource = {}` in its props, this could just be `useState(null)`.

src/components/instantlaunches/launch/index.js
```diff
@@ -61,20 +68,34 @@ const InstantLaunchStandalone = (props) => {
...
+                    errorMessage={err.message}
                     onDetailsClick={() => setErrorDialogOpen(true)}
                 />
                 <DEErrorDialog
                     open={errorDialogOpen}
-                    errorObject={error}
+                    errorObject={err}
                     handleClose={() => setErrorDialogOpen(false)}
                 />
             </>
```
psarando, line 101: I would replace this with an `ErrorTypographyWithDialog`. Also, the pattern we've been using with the error dialogs is to use a more general error message in the pop up, such as `t("instantLaunchError")`, and then the user can click the error dialog button to view more details to see if it was due to fetching resource info or trying to run the instant launch.

psarando (approved):
LGTM 👍 though I did have a couple of suggestions.

src/components/instantlaunches/launch/index.js
ianmcorvidae: I think that doesn't work, from what I could find? The destructuring assignment defaults do that with `undefined` but not with `null` according to MDN ( https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#default_value )

image

(and I figured that sticking `undefined` in react state would probably explode somehow, lol)

src/components/instantlaunches/launch/index.js
ianmcorvidae: Yeah, that's probably good. I admittedly saw the `errorHandler` prop to that and didn't want to figure that out just then, so did this, but given that's passed through to `DEErrorDialog` anyway now that I look, no reason to avoid it. I'll change to the wrapper & the more generic message (whether that one or a new one).

src/components/instantlaunches/launch/index.js
ianmcorvidae: Looking deeper, though, we're just checking the truthiness of `resource` later on, so null would work fine. I'll change it.

ianmcorvidae:
Thanks for the review! I made your recommended changes, both seemed good (after me waffling a little bit :sweat_smile: )

src/components/instantlaunches/launch/index.js
psarando: Oops, you're right about the destructuring, but there is no problem using `undefined` with `useState`. We use `useState()` all over the place.

I think in Typescript, depending on how your types are defined, it might complain if you use `undefined` with `useState` instead of an empty string or something else, but we haven't incorporated any Typescript in our project yet 😅

PR #599, opened 2024-10-11
Display information about the app and any files/folders in the instant launch dialog

ianmcorvidae:
This is a fairly simple display for now, just a bit of body-styled text that shows the app name and any file/folder that's passed in.

I'm figuring we can update this display later, but the bulk of this PR is getting everything wired up so we have access to this information in the popup.

### DIFF
--- a/src/components/instantlaunches/index.js
+++ b/src/components/instantlaunches/index.js
@@ -63,8 +63,12 @@ export const InstantLaunchSubmissionDialog = ({
             <DialogContent>
                 {appInfo && (
                     <Typography variant="body1">
-                        Launching app: {appInfo.name}
-                        {resource ? " with data " + resource.path : ""}
+                        {resource
+                            ? t("launchingAppWithData", {
+                                  appName: appInfo.name,
+                                  path: resource.path,
+                              })
+                            : t("launchingApp", { appName: appInfo.name })}
                     </Typography>
                 )}
                 <div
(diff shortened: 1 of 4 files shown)


src/components/instantlaunches/index.js
```diff
@@ -57,6 +61,12 @@ export const InstantLaunchSubmissionDialog = ({ open }) => {
             </DialogTitle>
 
             <DialogContent>
+                {appInfo && (
+                    <Typography variant="body1">
+                        Launching app: {appInfo.name}
+                        {resource ? " with data " + resource.path : ""}
```
psarando, line 67: These should be i18n messages, and it might be easiest to use 2 messages, something like this:
```suggestion
                        {resource
                            ? t("launchingAppWithData", {
                                  appName: appInfo.name,
                                  path: resource.path,
                              })
                            : t("launchingApp", { appName: appInfo.name })}
```

psarando (changes requested):

src/components/instantlaunches/index.js
ianmcorvidae: Heh, yeah. I even commented yesterday that I should do this and then lost track of it. Pushing up a fix now, sorry about that

psarando (approved):
LGTM 👍

ianmcorvidae:
Cool, thanks for the review/rereview!

PR #603, opened 2024-10-14
When autolaunching an instant launch, redirect to the analysis landing page

ianmcorvidae:
At least for now, only the dedicated launch page uses the `autolaunch` prop, so I don't think we need to add a new prop at this time.

### DIFF
--- a/src/components/instantlaunches/InstantLaunchButtonWrapper.js
+++ b/src/components/instantlaunches/InstantLaunchButtonWrapper.js
@@ -12,6 +12,10 @@ import React, { useEffect, useCallback } from "react";
 
 import { useMutation } from "react-query";
 
+import { useRouter } from "next/router";
+
+import NavigationConstants from "common/NavigationConstants";
+
 import { useDefaultOutputDir } from "components/data/utils";
 import withErrorAnnouncer from "components/error/withErrorAnnouncer";
 import { getHost } from "components/utils/getHost";
@@ -52,6 +56,7 @@ function InstantLaunchButtonWrapper(props) {
     const [runErrorDetails, setRunErrorDetails] = React.useState(null);
     const [ilUrl, setIlUrl] = React.useState();
 
+    const router = useRouter();
     const { t } = useTranslation("launch");
 
     React.useEffect(() => {
@@ -74,6 +79,12 @@ function InstantLaunchButtonWrapper(props) {
                     );
                 } else {
                     setOpen(false);
+                }
+                if (autolaunch) {
+                    // go to the analysis landing
+                    router.push(
+                        `/${NavigationConstants.ANALYSES}/${analysis?.id}`
+                    );
                 }
             } else {
                 setOpen(false);
--- a/src/components/instantlaunches/InstantLaunchButtonWrapper.js
+++ b/src/components/instantlaunches/InstantLaunchButtonWrapper.js
@@ -55,22 +55,30 @@ function InstantLaunchButtonWrapper(props) {
         React.useState(false);
     const [runErrorDetails, setRunErrorDetails] = React.useState(null);
     const [ilUrl, setIlUrl] = React.useState();
+    const [landingUrl, setLandingUrl] = React.useState();
 
     const router = useRouter();
     const { t } = useTranslation("launch");
 
     React.useEffect(() => {
-        if (ilUrl) {
+        if (ilUrl && landingUrl) {
             window.open(`${getHost()}${ilUrl}`);
             setIlUrl(null);
             setOpen(false);
+            if (autolaunch) {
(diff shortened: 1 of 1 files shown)


src/components/instantlaunches/InstantLaunchButtonWrapper.js
```diff
@@ -75,6 +80,12 @@ function InstantLaunchButtonWrapper(props) {
                 } else {
                     setOpen(false);
                 }
+                if (autolaunch) {
+                    // go to the analysis landing
+                    router.push(
+                        `/${NavigationConstants.ANALYSES}/${analysis?.id}`
+                    );
```
ianmcorvidae, line 87: The main question I have here is if I should use `router.replace` instead, so that the instant launch autolaunch page doesn't end up in the browser history (i.e., so that hitting "back" will go back to wherever someone clicked the badge from, not back to the launch page). Feeling like the answer is maybe yes - we do this for example to prevent hitting back and going back to the file create page.

ianmcorvidae: I convinced myself I should do this, heh, so I'm gonna go ahead and do that. No sense keeping any page that does something automatically in the browser history.

src/components/instantlaunches/InstantLaunchButtonWrapper.js
```diff
@@ -75,6 +80,12 @@ function InstantLaunchButtonWrapper(props) {
                 } else {
                     setOpen(false);
                 }
+                if (autolaunch) {
+                    // go to the analysis landing, not keeping this page in browser history
+                    router.replace(
+                        `/${NavigationConstants.ANALYSES}/${analysis?.id}`
+                    );
+                }
```
psarando, line 88: If this is going to navigate away from this page immediately in the launch success callback, then I see no need to call that `setIlUrl` above, which is intended to open a new window in a `useEffect` after that `setIlUrl` state changes in this success callback. I would think the page would navigate away before that `useEffect` could open that URL.

So I would move this up and make it the first check in the if-else block:
```javascript
                if (autolaunch) {
                    // go to the analysis landing, not keeping this page in browser history
                    router.replace(
                        `/${NavigationConstants.ANALYSES}/${analysis?.id}`
                    );
                } else if (analysis.interactive_urls?.length > 0) {
```

On the other hand, if the VICE URL should still open in a new window on successful launch, then maybe this `router.replace` call should be moved into that `useEffect` after the `ilUrl` is opened in a new window.

psarando (changes requested):

src/components/instantlaunches/InstantLaunchButtonWrapper.js
ianmcorvidae: Ah, yeah, both things should still happen, good point. This redirect is there because, unlike any other place you'd open an instant launch (dashboard, sidebar, instant launch listing, data listing), there's not really any content on the autolaunch page, so without this it opens the VICE URL in a new tab but just shows a blank DE page. I figured taking users to the page where they could stop the analysis or such would be the most useful place to send them, especially since they may have come from a link outside the DE/not already have a DE tab open for notifications, termination, etc.

ianmcorvidae: I moved the `router.replace` into the `useEffect`, hopefully correctly (worked when I tested it, at least)

src/components/instantlaunches/InstantLaunchButtonWrapper.js
```diff
@@ -51,21 +55,30 @@ function InstantLaunchButtonWrapper(props) {
...
     const [runErrorDetails, setRunErrorDetails] = React.useState(null);
     const [ilUrl, setIlUrl] = React.useState();
+    const [landingUrl, setLandingUrl] = React.useState();
 
+    const router = useRouter();
     const { t } = useTranslation("launch");
 
     React.useEffect(() => {
-        if (ilUrl) {
+        if (ilUrl && landingUrl) {
```
ianmcorvidae, line 64: Because the landingUrl getting set depends on even less than the ilUrl, I think it's safe to have this be a straight `&&` rather than also guarding on the value of `autolaunch`, but it could be changed to `ilUrl && (!autolaunch || landingUrl)` if we'd rather be more careful, I guess.

psarando: LGTM. If fact, this could have remained just `if (ilUrl)` and below could check `if (autolaunch && landingUrl)`.

psarando (approved):
LGTM 👍

ianmcorvidae:
Cool. Thanks for the review & catching my oversight!

PR #605, opened 2024-11-02
CORE-1955 App Version Ordering Form

psarando:
This PR will add an App Version Ordering Form that uses the endpoint added by cyverse-de/terrain#290.

App Version Ordering Form

---

A new menu item will be added to the apps listing row dot-menu for apps the user can edit:
Apps Listing Row Dot-Menu


psarando:
I think this should be ready for review now, although I might hold off on merging it until the QA db is back up, so I can test it against some local services and the QA db.

psarando:
I'm going to go ahead and merge this now, thanks for the review 👍

PR #606, opened 2024-11-04
Instant launch embed code & user-created instant launches

ianmcorvidae:
This PR:

* Wires up the non-administrative "create instant launch"
* Switches the admin stuff over to the proper admin endpoints
* Adds logic to create the right URL for embedding with a direct instant launch link, rather than a saved launch/wizard link
* Adds a toggle between instant and saved launch
* Adds a button to create an instant launch for users


src/components/apps/savedLaunch/SavedLaunchListing.js
```diff
@@ -29,23 +33,33 @@ import {
...
+    addInstantLaunch,
+} from "serviceFacades/instantlaunches";
+
 import { useConfig } from "contexts/config";
 import { useUserProfile } from "contexts/userProfile";
 
 import buildID from "components/utils/DebugIDUtil";
 import CopyTextArea from "components/copy/CopyTextArea";
 
+import Button from "@mui/material/Button";
```
ianmcorvidae, line 48: I don't know if it would be good to change these all into a single import (same with the icons). I just copied the code style within this file, though.

src/components/apps/savedLaunch/SavedLaunchListing.js
```diff
@@ -218,50 +252,68 @@ function ListSavedLaunches(props) {
...
+    }, [selected, selectedIL, useInstantLaunch, getShareUrl]);
+
+    const embedCodeClickHandler = () => {
+        setAnchorEl(null);
         setEmbedDialogOpen(true);
     };
 
     const shareClickHandler = () => {
         setAnchorEl(null);
-        setSavedLaunchUrl(getShareUrl(selected.id));
```
ianmcorvidae, line 235: `getShareUrl` doesn't actually take an argument, it looks at `selected` itself. So I removed that.

src/components/apps/savedLaunch/SavedLaunchListing.js
```diff
@@ -381,6 +433,49 @@ function ListSavedLaunches(props) {
...
+                            <Button
+                                variant="contained"
+                                startIcon={<Add />}
+                                onClick={(event) => {
+                                    event.stopPropagation();
+                                    event.preventDefault();
+                                    createInstantLaunch(selected.id);
+                                }}
+                            >
+                                {t("common:create")}
```
ianmcorvidae, line 469: It occurs to me after marking this ready that this text probably should specify that you're creating an instant launch, specifically. I copied it from the admin page where it's pretty clearly for instant launches only. I'll think about what to use for that.

slr71 (approved):
I didn't catch any issues. 👍

src/components/apps/savedLaunch/SavedLaunchListing.js
```diff
@@ -218,50 +252,68 @@ function ListSavedLaunches(props) {
...
+        const url = `${host}${getSavedLaunchPath(
+            systemId,
+            appId,
+            selected?.id
+        )}`;
+        return url;
+    }, [selected, systemId, appId]);
+
+    useEffect(() => {
+        var shareUrl = "";
```
psarando, line 275: This can be a [let](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let):
```suggestion
        let shareUrl = "";
```

src/components/apps/savedLaunch/SavedLaunchListing.js
```diff
@@ -381,6 +433,49 @@ function ListSavedLaunches(props) {
...
+                                    </Typography>
+                                }
+                            />
+                        ) : (
+                            <Button
+                                variant="contained"
+                                startIcon={<Add />}
+                                onClick={(event) => {
+                                    event.stopPropagation();
+                                    event.preventDefault();
```
psarando, line 465: I don't think these `event` calls should be required here. Usually this is only needed if the Button is on top/inside of something else that is intercepting clicks, like maybe an `AccordionSummary` or `MenuItem`  🤔

src/components/apps/utils.js
```diff
@@ -51,6 +51,21 @@ export const getAppLaunchPath = (systemId, appId, versionId) =>
...
     `/${NavigationConstants.APPS}/${systemId}/${appId}/launch?saved-launch-id=${launchId}`;
 
+/**
+ * Builds a path to the direct instant launch page.
+ *
+ * @param {string} ilId The Instant Launch ID
+ * @param {string} resourcePath A file or folder path to pass to the instant launch (optional)
+ */
+export const getInstantLaunchPath = (ilId, resourcePath = "") => {
+    var uri = `/${NavigationConstants.INSTANT_LAUNCH}/${ilId}`;
```
psarando, line 61: This can also be a `const`:
```suggestion
    const uri = `/${NavigationConstants.INSTANT_LAUNCH}/${ilId}`;
```

src/components/apps/utils.js
```diff
@@ -51,6 +51,21 @@ export const getAppLaunchPath = (systemId, appId, versionId) =>
...
+    if (resourcePath !== "") {
```
psarando, line 62: It might be safer to just check truthiness?
```suggestion
    if (resourcePath) {
```

ianmcorvidae:
Putting in some screenshots here for good measure to show the current state with my current pushes:

image
image
image

(getting back to Paul's feedback now)

src/components/apps/utils.js
ianmcorvidae: Hm, that's probably fair. I can't imagine it somehow gets passed something that's a non-string and I guess it'd figure out how to interpolate it (not that it'd be a meaningful value anyway) if it did.

psarando: I think it's happened at least once before where the UI sent `null` in a param like this to the backend 😖

src/components/apps/savedLaunch/SavedLaunchListing.js
```diff
@@ -198,7 +215,24 @@ function ListSavedLaunches(props) {
...
 
+    // It would be better to have a query that is limited to the selected app only
+    const {
+        data: allILs,
+        error: errorILs,
+        isFetching: fetchingILs,
+    } = useQuery(ALL_INSTANT_LAUNCHES_KEY, listFullInstantLaunches);
+
     const savedLaunchClickHandler = (event, savedLaunch) => {
+        var instantLaunch = allILs.instant_launches.find(
```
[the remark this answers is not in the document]
ianmcorvidae: Yeah, may as well be safe with the `?.` -- the saved launches dialog won't render this button until the `allILs` query has returned (since it's using `isQueryLoading([isFetching, fetchingILs])`) but still better than weird errors with undefined values.

src/components/apps/savedLaunch/SavedLaunchListing.js
ianmcorvidae: of course, `let`! I knew there was a better option than `var` but I knew this couldn't be a `const`. Shoulda googled it, heh.

src/components/apps/savedLaunch/SavedLaunchListing.js
ianmcorvidae: Hm, that's fair, though I kind of feel like keeping it here because then it's more insurance in case this at some point _does_ end up within something that might care, but if there's a reason to remove it I'm happy to do that as well.

src/components/apps/utils.js
ianmcorvidae: Oh, right, I'm not actually modifying this string later, just returning a value with a suffix. Good catch!

src/components/apps/utils.js
ianmcorvidae: Yeah. In this case it's just generating a URL, I guess, so it'd just end up looking weird and presumably fail to load the path when trying to launch from it. Which isn't great, but not "shoving bad data into the database" bad or anything, heh.

src/components/apps/savedLaunch/SavedLaunchListing.js
ianmcorvidae: Cool. Wasn't sure if we had any sort of code style we cared about, but always very glad to just match existing code :grin:

psarando (approved):
LGTM 👍

psarando:
Oh, it looks the service facade updates broke a couple of our stories: https://www.chromatic.com/build?appId=5ebad5c6b991d60022b1e72e&number=2796

ianmcorvidae:
Ah, yeah, I see. The stories don't have the instant launch information so they're throwing an error when loading the saved launches. I'll look into adding the instant launch endpoint stuff.

ianmcorvidae:
Alright. I added some simple dummy data to get the stories back to rendering as before.

ianmcorvidae:
I'll go ahead and merge this now I think. QA is down but I think it's ready to go. I'll test as I'm able to here.

PR #610, opened 2024-12-04
Make the standalone instant launch page show errors immediately

ianmcorvidae:
This is better than putting it behind an `ErrorTypographyWithDialog` because there's not really much reason to just show a text error that tells you almost nothing with a detail button, on a page with nothing else. I also made it so if you close the error dialog it just redirects you to the dashboard since there's not much reason to stay on the page at that point.

### DIFF
--- a/src/components/instantlaunches/launch/index.js
+++ b/src/components/instantlaunches/launch/index.js
@@ -3,11 +3,9 @@
  *
  * The component that launches a provided instant launch, immediately.
  */
-import React, { useState, useEffect } from "react";
+import React, { useState } from "react";
 
 import { useQuery } from "react-query";
-
-import { useRouter } from "next/router";
 
 import {
     GET_INSTANT_LAUNCH_FULL_KEY,
@@ -28,7 +26,7 @@ import InstantLaunchButtonWrapper from "components/instantlaunches/InstantLaunch
 import withErrorAnnouncer from "components/error/withErrorAnnouncer";
 import LoadingAnimation from "components/vice/loading/LoadingAnimation";
 
-import DEErrorDialog from "components/error/DEErrorDialog";
+import WrappedErrorHandler from "components/error/WrappedErrorHandler";
 
 const InstantLaunchStandalone = (props) => {
     const {
@@ -42,10 +40,7 @@ const InstantLaunchStandalone = (props) => {
         !!config?.subscriptions?.enforce
     );
     const [resource, setResource] = useState(null);
-    const [errDialogOpen, setErrDialogOpen] = useState(false);
-    const [dialogError, setDialogError] = useState(null);
 
-    const router = useRouter();
     const { t } = useTranslation(["instantlaunches", "common"]);
 
     const { data, status, error } = useQuery(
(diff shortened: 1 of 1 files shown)


src/components/instantlaunches/launch/index.js
```diff
@@ -81,14 +86,24 @@ const InstantLaunchStandalone = (props) => {
...
-            <ErrorTypographyWithDialog
-                errorMessage={t("instantLaunchError")}
-                errorObject={err}
+            <DEErrorDialog
+                open={errDialogOpen}
+                errorObject={dialogError}
+                handleClose={() => {
+                    setErrDialogOpen(false);
+                    router.push("/");
+                }}
```
ianmcorvidae, line 106: My question here is basically: Should I have the `DEErrorDialog` (closed) always get rendered in and just let the `useEffect` open it, or is it still worth having this `else if` to only render in the dialog when we actually have an error?

ianmcorvidae: (er, I guess by "always get rendered in" I mean after the loading animation is finished, probably still shouldn't render it on the load screen)

psarando: > My question here is basically: Should I have the `DEErrorDialog` (closed) always get rendered in and just let the `useEffect` open it, or is it still worth having this `else if` to only render in the dialog when we actually have an error?

In this case [I don't think the `useEffect` is needed at all](https://react.dev/learn/you-might-not-need-an-effect), since rendering the dialog is already conditional on the `isLoading` and error states.
```jsx
            <DEErrorDialog
                open={true}
                errorObject={error || resourceError}
                handleClose={() => {
                    router.push("/");
                }}
            />
```

With that said, I think it might be simpler to use the `src/components/error/WrappedErrorHandler`, for example:
https://github.com/cyverse-de/sonora/blob/b26b43437d8dddd1397d7132e4627e1d07901457/src/components/apps/launch/index.js#L138

psarando (changes requested):

ianmcorvidae:
Heh, yeah, this is a lot cleaner with WrappedErrorHandler.

src/components/instantlaunches/launch/index.js
```diff
@@ -84,13 +84,7 @@ const InstantLaunchStandalone = (props) => {
...
         return <LoadingAnimation />;
     } else if (error || resourceError) {
-        const err = error || resourceError;
-        return (
-            <ErrorTypographyWithDialog
-                errorMessage={t("instantLaunchError")}
-                errorObject={err}
-            />
-        );
+        return <WrappedErrorHandler errorObject={error || resourceError} />;
```
psarando, line 87: This should also have a `baseId`, though I don't see one already setup for this button wrapper 🤔 
I guess `ids.BASE` should work for now 🤷

ianmcorvidae: Yeah, I haven't been sure what to do for the debug IDs. I figure we don't really use them anymore really anyway, but I'll add... something, at least.

psarando (approved):
LGTM 🎉  :shipit: 👍

PR #611, opened 2024-12-07
Use `await import` for `lottie-web` in LoadingAnimation

psarando:
Builds on my laptop are failing when trying to build SSR pages that use the `LoadingAnimation`, such as `/instantlaunch/[id]`.

They are not failing in GitHub workflow skaffold builds, or on other dev machines, for now, but maybe this will prevent future build errors.

### DIFF
--- a/src/components/vice/loading/LoadingAnimation.js
+++ b/src/components/vice/loading/LoadingAnimation.js
@@ -4,7 +4,6 @@
 import React, { useEffect, useRef } from "react";
 
 import { makeStyles } from "tss-react/mui";
-import lottie from "lottie-web";
 import animation from "./vice_loading.json";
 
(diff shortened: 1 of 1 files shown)


ianmcorvidae (approved):
Seems good to me. It already was building fine (for me), but lottie-web is pretty big anyway so it's good to keep out of the bundles. As long as the loading animation is still displaying fine, this seems like a pretty clear win.

psarando:
> Seems good to me... As long as the loading animation is still displaying fine, this seems like a pretty clear win.

Oops, helps if I actually test these changes 😊 

Turns out I used the wrong kind of dynamic/async import.
It's working now with `await import`, so I'll go ahead and merge this.

Thanks for the review! 👍

PR #614, opened 2025-01-13
Login table

ianmcorvidae:
This adds a table of recent logins to the dashboard, using the apps/terrain endpoints added elsewhere.

This appears to the right of the analysis stats, which have been cut to half-width, on larger screens, and after the analysis stats on small screens. IP addresses redacted in case it matters, but looks like this:

image

I'm not sure how much I like the second column for the full date, but I thought it was more readable with both the 'ago' and the full date, and with the full dates aligned with each other in the table. Open to suggestions on how to do that a bit better, but I think this basically works in any case.


src/components/dashboard/dashboardItem/LoginTable.js
```diff
@@ -0,0 +1,84 @@
...
+        queryKey: [LOGINS_QUERY_KEY],
+        queryFn: () => logins({ limit: 5 }),
+    });
+
+    if (status === "error") {
+        return (
+            <div style={{ padding: theme.spacing(1) }}>
+                <ErrorTypographyWithDialog
+                    errorObject={error}
+                    errorMessage={""}
```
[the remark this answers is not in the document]
ianmcorvidae: (alright, handled that, sorry about that)

src/components/dashboard/dashboardItem/LoginTable.js
```diff
@@ -0,0 +1,84 @@
...
+                                {i18nCommon("timestamp", {
+                                    timestamp: getFormattedDistance(
+                                        row["login_time"] / 1000
+                                    ),
+                                })}
+                            </TableCell>
+                            <TableCell>
+                                ({formatDate(row["login_time"])})
+                            </TableCell>
+                            <TableCell>{row["ip_address"]}</TableCell>
```
psarando, line 77: Just a reminder that accessing JSON objects like arrays work fine (and is required if the field has characters like `-`), but you can also access these fields as `row.ip_address`, `row.login_time`, and `data?.logins` (best to guard with `?` on `data` in case it didn't load yet, although the previous `error` and `loading` checks should guard against that as well).

src/components/dashboard/dashboardItem/LoginTable.js
```diff
@@ -0,0 +1,84 @@
...
+                <TableHead>
+                    <TableRow>
+                        <TableCell>{t("loginTime")}</TableCell>
+                        <TableCell></TableCell>
+                        <TableCell>{t("ipAddress")}</TableCell>
+                    </TableRow>
+                </TableHead>
+                <TableBody>
+                    {data["logins"].map((row) => (
+                        <TableRow>
```
psarando, line 66: The browser dev console should be reporting a warning that these array elements each require a unique key. Since these rows aren't going to be dynamically rearranging, I think an `index` key should be fine here (unless the `row` has a `uuid`, which would be even better):
```suggestion
                    {data["logins"].map((row, index) => (
                        <TableRow key={index}>
```

src/components/dashboard/dashboardItem/LoginTable.js
```diff
@@ -0,0 +1,84 @@
...
+                            <TableCell>
+                                {i18nCommon("timestamp", {
```
psarando, line 68: This works fine, so no need to change it, but just FYI that the i18n library also supports passing multiple namespaces to `useTranslation`, and the first one will be used by default in calls to `t`.

So if the `useTranslation` call was updated to `const { t } = useTranslation(["dashboard", "common"]);`, then the previous calls like `t("loginTime")` can remain the same, but this one can become:
```suggestion
                                {t("common:timestamp", {
```

src/components/dashboard/dashboardItem/LoginTable.js
```diff
@@ -0,0 +1,84 @@
...
+                                    timestamp: getFormattedDistance(
+                                        row["login_time"] / 1000
+                                    ),
+                                })}
+                            </TableCell>
+                            <TableCell>
+                                ({formatDate(row["login_time"])})
+                            </TableCell>
```
psarando, line 76: Personally I think it makes more sense to swap these columns, so something like: `2025-01-21 09:10:11    (about 1 day ago)`

psarando (changes requested):
It looks like the login table rows need unique keys.

Otherwise I just had some minor suggestions, but there's also bonus points for adding `mockAxios` calls to the dashboard stories 😉

src/components/dashboard/dashboardItem/LoginTable.js
ianmcorvidae: Ah! I had done the useTranslation that way at first, but hadn't realized I needed to prefix the ones on the non-first, so it wasn't working (obviously). I might change it just since that does seem cleaner.

src/components/dashboard/dashboardItem/LoginTable.js
ianmcorvidae: Ah right, of course. We do have a UUID but currently aren't passing it up from the backend. I'll probably just use an index as you say; it's true these shouldn't change after initial rendering.

src/components/dashboard/dashboardItem/LoginTable.js
ianmcorvidae: I suppose with that I could also remove the second-column thing, because the dates will be the same width regardless. so that definitely seems cleaner. Trying too hard to copy other sites' stuff I think, heh

ianmcorvidae: image

here's how I've got it now, and I think this looks pretty good -- added the italics as well, to visually distinguish the parenthetical a little bit more

src/components/dashboard/dashboardItem/LoginTable.js
ianmcorvidae: I've kept it as array-like syntax for now for the individual row fields, but I did change it over to the `data?.logins` way, since that did seem like a good point. I figure that the array-like syntax lets it be consistent when there's hyphens, but that's probably not really very meaningful, so if anyone feels strongly about it I can change it to dot notation.

psarando (approved):
LGTM 👍 🎉

ianmcorvidae:
Excellent, thanks for the re-review! I'll merge now

PR #615, opened 2025-01-21
Add automatic SSO redirect for authenticated users

suemkwon:
When a user has an active Keycloak session,  they will now be automatically redirected to the dashboard instead of seeing the landing/login page.

- Add checkSso middleware to root route
- Add redirect to dashboard for authenticated users
- Maintain fallback to landing page for unauthenticated users

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -101,6 +101,20 @@ app.prepare()
         logger.info("configuring keycloak");
         server.use(keycloakClient.middleware());
 
+        server.get("/", keycloakClient.checkSso(), async (req, res, next) => {
+            try {
+                const userProfile = authn.getUserProfile(req);
+
+                if (userProfile) {
+                    return res.redirect(`/${NavigationConstants.DASHBOARD}`);
+                }
+
+                return nextHandler(req, res);
+            } catch (error) {
+                return next(error);
+            }
+        });
+
         logger.info("adding the /login handler");
         server.get("/login", keycloakClient.protect(), (req, res) => {
             res.redirect("/");


slr71:
I can explain the purpose a little bit. Users have mentioned in the past that it annoys them that they have to click the login link in order to log in even when they already have an SSO session. To replicate the behavior that users are complaining about:

- Log into the user portal.
- Open a new browser tab and go to the DE.
- You're still not logged into the DE.
- Click the login link (either the link on the dashboard or the user avatar)

With this code change, the behavior is a little bit different:

- Log into the user portal.
- Open a new browser tab and go to the DE.
- You're automatically logged into the DE.

The code seems to work as intended, but it's taking a while to actually log the user in. I wonder if there's a way to fix that.

src/server/index.js
```diff
@@ -101,6 +101,20 @@ app.prepare()
         logger.info("configuring keycloak");
         server.use(keycloakClient.middleware());
 
+        server.get("/", keycloakClient.checkSso(), async (req, res, next) => {
+            try {
+                const userProfile = authn.getUserProfile(req);
+
+                if (userProfile) {
+                    return res.redirect(`/${NavigationConstants.DASHBOARD}`);
```
psarando, line 109: This works fine as-is, but users don't normally navigate to the `/dashboard` page anywhere else within the DE, since our next config has it aliased to `/`.

I noticed if we redirect here to `/login` instead of `/dashboard`, then this should log the user into the DE as well, and the handler below will redirect the user back to `/`.

psarando (approved):
Thanks! I was able to reproduce this issue.

LGTM, in that case 👍 

I'm not sure why it seems like it takes a while to log-in with this change. In my testing in prod while reproducing this issue, I logged in to user.cyverse.org, then directly navigated to de.cyverse.org/dashboard or de.cyverse.org/login, and the dashboard seemed to load for me in about one second (and I was logged into the DE).

src/server/index.js
psarando: Actually, never mind this 😊 

I just tested this out in dev mode, and Firefox complained about a redirect loop, so it seems like redirecting to `/dashboard` is our best option.

:shipit: 👍

psarando:
Another issue I've seen, that's probably related to this, is that when I log into the DE and close the tab for the day, and then reopen that tab the next day, it still displays as if I'm still logged in with a keycloak session, but I'm really not. So when it tries to retrieve something from an authenticated endpoint (such as `bags` content), the page displays a "forbidden" error.

I'm not sure why the root `/` page doesn't detect the keycloak session correctly in these cases, but eventually we should probably find a solution for both issues; but this workaround will work for the logged-in case for now.

PR #616, opened 2025-01-22
Replace min/max CPU fields with single CPU cores field

suemkwon:
- Changed resource requirements form to use a single CPU cores field instead of separate min/max fields
- Updated form validation to remove min/max CPU validation logic
- Updated review display to show single CPU cores value
- Added new translation key for "CPU Cores" label
- Removed unused code related to min/max CPU validation

These changes simplify the CPU resource selection process and reduce user confusion around CPU requirements.
Screenshot 2025-01-22 at 3 25 21 PM


src/components/apps/launch/ResourceRequirements.js
```diff
@@ -127,8 +127,8 @@ const StepResourceRequirementsForm = ({
                 <Grid item xs={12}>
                     <FastField
                         id={buildID(baseId, ids.RESOURCE_REQUESTS.TOOL_CPU)}
-                        name={`requirements.${index}.min_cpu_cores`}
-                        label={t("minCPUCores")}
+                        name={`requirements.${index}.cpu_cores`}
```
slr71, line 130: The api is still expecting `min_cpu_cores` here, so this field name should stay the same as before. We may change the APIs eventually, but this can be revisited then if we do.

src/components/apps/launch/ResourceRequirements.js
```diff
@@ -372,11 +339,11 @@ const StepResourceRequirementsReview = ({
                         <Table>
                             <TableBody>
                                 <ResourceRequirementsReviewRow
-                                    label={t("minCPUCores")}
-                                    value={min_cpu_cores}
+                                    label={t("cpuCores")}
+                                    value={cpu_cores}
```
slr71, line 343: I could be wrong, but I think that this line might have to be left unchanged as well.

slr71 (changes requested):
The changes look good to me for the most part. There was at least one API incompatibility (and possibly two) in this change. Those will have to be fixed before this can be merged.

psarando (changes requested):
I recommend replacing the `min_cpu_cores` form field with the `max_cpu_cores` form field, since I think the `max_cpu_cores` field is more essential is controlling how many CPUs per hour a user will consume for either VICE or Condor analyses.

Also, I don't see any Chromatic builds associated with this PR, so you probably still need to [set up a Chromatic token for your `sonora` fork](https://github.com/cyverse-de/sonora/blob/main/guidelines.md#chromatic-setup).

src/components/apps/launch/ResourceRequirements.js
```diff
@@ -320,30 +298,19 @@ const StepResourceRequirementsReview = ({
...
+    const { step_number, min_memory_limit, min_disk_space, cpu_cores } =
+        stepRequirements;
 
-    const hasRequest = !!(
-        min_cpu_cores ||
-        min_memory_limit ||
-        min_disk_space ||
-        max_cpu_cores
-    );
+    const hasRequest = !!(min_memory_limit || min_disk_space || cpu_cores);
```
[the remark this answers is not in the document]
slr71: Assuming that we want to set `min_cpu_cores` and `max_cpu_cores` explicitly if the user selects any value for the CPU Cores field, I think this might have to be left as it was before.

psarando:
> Also, I don't see any Chromatic builds associated with this PR, so you probably still need to [set up a Chromatic token for your `sonora` fork](https://github.com/cyverse-de/sonora/blob/main/guidelines.md#chromatic-setup).

I'm seeing Chromatic builds for this PR now, so I guess I was wrong about this (I switched to a new "merge experience" view, but also there were no visual storybook changes, so I the Chromatic builds were hidden from my view by default).

src/components/apps/launch/ResourceRequirements.js
psarando: > My concern with setting only the `max_cpu_cores` field is that if the default number of CPUs for the app is lower than the requested maximum then there's a chance that the analysis will be scheduled on a node that doesn't have enough CPUs to run the analysis.
> 

That's a good point, I didn't think of that.

> Another option would be to set both the `min_cpu_cores` and `max_cpu_cores` fields to the same value, which may be preferable simply because it's more explicit. Now that I think about it, being explicit is probably the best option simply because it avoids confusion, so my recommendation would be to set both the `min_cpu_cores` and `max_cpu_cores` fields.

This does sound like the best option. In that case, I still recommend using the `max_cpu_cores` field in the UI form, and removing the `min_cpu_cores` form field from displaying in the UI. This way the logic in the `initAppLaunchValues` formatter does not have to be updated to set a default selected value, so that the max a tool allows will not be used by the API by default:
https://github.com/cyverse-de/sonora/blob/886c95d805801a8c989f29e67db2cf2732d68634/src/components/apps/launch/formatters.js#L44-L64

The submission formatter will also need to be updated so that each requirement sets its `min_cpu_cores` value with the `max_cpu_cores` form value (maybe with a `requirements: requirements.map( ... )` function):
https://github.com/cyverse-de/sonora/blob/886c95d805801a8c989f29e67db2cf2732d68634/src/components/apps/launch/formatters.js#L185-L215

So we would still remove the `selectMins` and `selectMaxes` typography labels, but the `max_cpu_cores` field can now be labeled with the new `cpuCores` label.

src/components/apps/launch/ResourceRequirements.js
```diff
@@ -320,30 +298,19 @@ const StepResourceRequirementsReview = ({
...
     return (
         (showAll || hasRequest) && (
             <Accordion
                 defaultExpanded={
                     !!(
-                        stepRequirementErrors?.min_cpu_cores ||
                         stepRequirementErrors?.min_memory_limit ||
                         stepRequirementErrors?.min_disk_space ||
-                        stepRequirementErrors?.max_cpu_cores
+                        stepRequirementErrors?.cpu_cores
```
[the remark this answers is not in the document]
psarando: This controls if the resource requirements are displayed in the review step, but only if those form fields have any values selected. If we are only displaying the `max_cpu_cores` form field, and hiding `min_cpu_cores` in the form, then the `stepRequirementErrors?.min_cpu_cores ||` line can be removed here.

src/components/apps/launch/ResourceRequirements.js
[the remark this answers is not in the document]
psarando: This logic is only for displaying values in the review step that have been set in the previous step of the form. If we are only displaying the `max_cpu_cores` field, then the check for `min_cpu_cores` does not need to be included in `hasRequest` here.

src/components/apps/launch/ResourceRequirements.js
psarando: Note that the `value` here comes from the JSON returned by the `initAppLaunchValues` formatter. So we can rename this label with the new `cpuCores` label, but in order to change the form field name, both the `initAppLaunchValues` formatter and the submission formatter would need to be updated to convert to and from the API JSON and the JSON used by the form fields.

Since both `max_cpu_cores` and `min_cpu_cores` are still needed to populate the CPU cores selection list values, I think it might be less work to just use `max_cpu_cores` as the form field name (though the display label can be updated):

https://github.com/cyverse-de/sonora/blob/886c95d805801a8c989f29e67db2cf2732d68634/src/components/apps/launch/ResourceRequirements.js#L103-L107

src/components/apps/launch/validate.js
```diff
@@ -238,18 +230,7 @@ const validate = (t, hasParams) => (values) => {
...
-                const err = validateNotAbove(
-                    req.min_cpu_cores,
-                    req.max_cpu_cores,
-                    t
-                );
-                if (err) {
-                    reqErrors[i] = { min_cpu_cores: err };
-                }
-            }
-        });
```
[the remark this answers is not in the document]
psarando: I overlooked this in my last review, but if we're now hiding `min_cpu_cores` in the form, then this error won't display anywhere anymore, so `reqErrors[i] = { min_cpu_cores: err }` here should be updated to `reqErrors[i] = { max_cpu_cores: err }`.

src/components/apps/launch/ResourceRequirements.js
```diff
@@ -127,8 +127,8 @@ const StepResourceRequirementsForm = ({
                 <Grid item xs={12}>
                     <FastField
                         id={buildID(baseId, ids.RESOURCE_REQUESTS.TOOL_CPU)}
-                        name={`requirements.${index}.min_cpu_cores`}
-                        label={t("minCPUCores")}
+                        name={`requirements.${index}.max_cpu_cores`}
+                        label={t("cpuCores")}
```
psarando, line 131: I would also remove the `selectMins` grid item just above this, since these are no longer all minimum requirements.

psarando (approved):
LGTM 👍 

I also tested the DE Word Count story, using the Chromatic build from the [Storybook Publish](https://5ebad5c6b991d60022b1e72e-emsgvywrvp.chromatic.com/) link under the `checks` section of this PR, and it appears to be submitting the correct requirements for both min and max CPUs 🎉

slr71 (approved):
Looks good to me as well! 👍 :shipit:

PR #617, opened 2025-01-30
Update file viewer to download binary files instead of displaying them

suemkwon:
The file viewer was attempting to display binary files in the text editor.

Modified the file viewer to automatically trigger downloads for non-text files instead of trying to display them. This is implemented by checking if the content-type starts with "text/" - if not, the file is handled by VIEWER_TYPE.DOCUMENT which opens in a new tab/window for download.

Testing: 
- Verified .txt .py .csv .pdf .jpg .png etc. still open in the file editor as expected
- Verified binary files (.xlsx, .docx, etc.) now trigger download instead of displaying in editor

### DIFF
--- a/src/components/data/viewers/FileViewer.js
+++ b/src/components/data/viewers/FileViewer.js
@@ -271,18 +271,22 @@ export default function FileViewer(props) {
                         setViewerType(VIEWER_TYPE.PATH_LIST);
                         break;
                     } else {
-                        if (!createFileType) {
-                            setReadChunkKey([
-                                READ_CHUNK_QUERY_KEY,
-                                {
-                                    path,
-                                    chunkSize:
-                                        viewerConstants.DEFAULT_PAGE_SIZE,
-                                },
-                            ]);
-                            setReadChunkQueryEnabled(true);
+                        if (manifest["content-type"].startsWith("text/")) {
+                            if (!createFileType) {
+                                setReadChunkKey([
+                                    READ_CHUNK_QUERY_KEY,
+                                    {
+                                        path,
+                                        chunkSize:
+                                            viewerConstants.DEFAULT_PAGE_SIZE,
+                                    },
+                                ]);
+                                setReadChunkQueryEnabled(true);
+                            }
+                            setViewerType(VIEWER_TYPE.PLAIN);
+                        } else {
+                            setViewerType(VIEWER_TYPE.DOCUMENT);
                         }
-                        setViewerType(VIEWER_TYPE.PLAIN);
                         break;
                     }
             }
--- a/src/components/data/viewers/FileViewer.js
+++ b/src/components/data/viewers/FileViewer.js
@@ -275,7 +275,6 @@ export default function FileViewer(props) {
                         const viewableApplicationTypes = [
                             mimeTypes.JSON,
                             mimeTypes.XML,
-                            mimeTypes.XHTML_XML,
                         ];
 
                         if (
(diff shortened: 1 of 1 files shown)


psarando (approved):
LGTM 👍 🎉

ianmcorvidae:
seems good to me, but one edge case to check: JSON. sometimes that comes through as `application/json` I think, and we wouldn't want to kill the editor for that, probably

johnworth (approved):
LGTM :shipit:

psarando:
> seems good to me, but one edge case to check: JSON. sometimes that comes through as `application/json` I think, and we wouldn't want to kill the editor for that, probably

Good catch 👍 
We should check for that, and add special handling if it's not already in there.

ianmcorvidae:
Checking one list I found, the other type of concern I found was `application/xml` and related ones like `application/xhtml+xml`

ianmcorvidae:
looks like https://github.com/apache/tika/blob/main/tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml is the source of the possible type detections by filename at least

suemkwon:
Since the PR was approved but there are concerns about JSON/XML files, I'd like to suggest this modification to handle those edge cases:

Screenshot 2025-01-30 at 2 34 15 PM

This would allow JSON and XML files to still open in the editor while maintaining the download behavior for true binary files. Should I make this change before merging?

ianmcorvidae:
I think that's a decent thing to do for now to get the other benefits of this in place. Maybe something in the Tika APIs could be used to detect & tell the UI if a given type should be editable or not, so that might be something to look into if we want to do it more properly

psarando (changes requested):
> Checking one list I found, the other type of concern I found was `application/xml` and related ones like `application/xhtml+xml`

According to the comments in `src/components/models/MimeTypes.js`, it seems `application/xhtml+xml` is intended to display natively in the browser, like PDFs.
https://github.com/cyverse-de/sonora/blob/4ed0ff66055e302a4284bae64b755c26b66c0562/src/components/models/MimeTypes.js#L13-L18

@suemkwon, the suggestion in your screenshot looks pretty close, but I would use the `mimeTypes` model (which has `JSON` and `XML` constants) and the `mimeType` variable already available in this function.

psarando:
> According to https://developer.mozilla.org/en-US/docs/Web/HTTP/MIME_types#textjavascript it should be `text/json` now, but it seems our services are returning `application/json` 😢

Oops, looks like I confused `application/json` with `application/javascript` for a minute there. So `application/json` is still the standard.

src/components/data/viewers/FileViewer.js
```diff
@@ -271,7 +271,17 @@ export default function FileViewer(props) {
                         setViewerType(VIEWER_TYPE.PATH_LIST);
                         break;
                     } else {
-                        if (manifest["content-type"].startsWith("text/")) {
+                        // Special handling for text-based formats that don't use text/ prefix
+                        const viewableApplicationTypes = [
+                            mimeTypes.JSON,
+                            mimeTypes.XML,
+                            mimeTypes.XHTML_XML,
```
psarando, line 278: The `application/xhtml+xml` content type is another type of HTML, so it should be viewed natively in the browser, like plain HTML and PDFs. So we should remove `mimeTypes.XHTML_XML` from this check.

suemkwon: Ah, good catch by @psarando ! The XHTML_XML type should be handled like HTML and PDFs using VIEWER_TYPE.DOCUMENT, not VIEWER_TYPE.PLAIN.

psarando (approved):
LGTM :shipit: 👍 

If users need more editable `application/` types in the future, we can add them to this list, or update our services as @ianmcorvidae suggested earlier. 🎉

PR #618, opened 2025-02-05
Enable public data downloads by using new unauthentication-allowed Terrain endpoints

ianmcorvidae:
Well, this was easy and seems to all work as expected. We weren't preventing logged-out users from pushing the buttons already, so nothing needed to be removed as far as that.

### DIFF
--- a/src/server/api/downloads.js
+++ b/src/server/api/downloads.js
@@ -79,7 +79,7 @@ const metadataTemplateCSVhandler = async (req, res) => {
 
 const doDownloadFromTerrain = (userID, accessToken, apiURL) => {
     logger.info(`TERRAIN ${userID} GET ${apiURL.href}`);
-    let requestOptions = {
+    const requestOptions = {
         method: "GET",
         url: apiURL,
         headers: {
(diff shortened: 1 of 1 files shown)


src/server/api/downloads.js
```diff
@@ -90,15 +79,17 @@ const metadataTemplateCSVhandler = async (req, res) => {
 
 const doDownloadFromTerrain = (userID, accessToken, apiURL) => {
     logger.info(`TERRAIN ${userID} GET ${apiURL.href}`);
-    const requestOptions = {
+    let requestOptions = {
```
psarando, line 82: Although this can remain a `let`, it can also stay as a `const`, since `requestOptions` isn't getting assigned to any other value (even though its contents are getting updated).

psarando (approved):
LGTM 👍

src/server/api/downloads.js
ianmcorvidae: Ah, right. Soft immutability, heh. Too many different programming languages in my brain, clearly :D

ianmcorvidae:
Thanks for the review!

PR #619, opened 2025-02-13
Use the user subscription from the usage summary, rather than making two calls

ianmcorvidae:
(that both fetch the same information)


psarando (changes requested):
Nicely done 🎉 

I think I caught a bug, though: It looks like there's a few places in the AppBar now that need to replace `userSubscription?.result` with just `userSubscription`, since the `/qms/user/plan` endpoint nested the subscription info in a `{result: ...}` key.

psarando (approved):
LGTM :shipit: 👍

PR #620, opened 2025-02-19
CORE-1985 Upgrade to React v18, next.js v14, and Storybook v8

psarando:
This PR will upgrade the UI to `react` v18, `next` v14, and `storybook` v8.

This also required updating the build containers and workflows to `node` v20+ (v23 appears to work on my laptop), and updating many other dependent packages, such as updating to `jest` v29.

The `eslintConfig` was also updated to use `eslint-config-next`, and with the react v18 upgrade, these required fixing some errors and warnings not reported before the upgrades. Some cases required some fairly extensive refactoring, such as fixing React Hook in class component errors.


ianmcorvidae (approved):
Lots of boilerplate-y changes, but looks good to me! Probably silences a bunch of errors and makes tracebacks more useful too.

psarando:
> Lots of boilerplate-y changes, but looks good to me! Probably silences a bunch of errors and makes tracebacks more useful too.

Thanks for the review! 👍 

Yes, lots of boilerplate, and I didn't expect anyone to review all these changes too closely, but it's probably a good reference for anyone that has done UI work recently, to see what's new with some of the new linter and react rules; such as no more anonymous default exports, and no more react hooks in class components.

I'll merge this now, which will also resolve some vulnerabilities in the old dependencies; then I'll follow up with some more commits to update a few more deps (to fix more vulnerabilities).

PR #621, opened 2025-02-26
CORE-2033 Rename Agave to Tapis in remaining components, stories, and tests

psarando:
This PR also fixes the `Reset HPC Token` action in the user settings page.


psarando:
Thanks for the review! 👍

PR #622, opened 2025-02-26
CORE-1985 Storybook v8 Fixes

psarando:


psarando:
This is just some minor fixes for the Storybook upgrade from #620, and everything looked good in Chromatic, so I'm going to go ahead and merge this now.

PR #623, opened 2025-03-07
CORE-2017: multi-year and future subscriptions

slr71:
This change adds support for the following fields to the subscription edit form.

* start date - allows the subscription start date to be specified
* end date - allows the subscription end date to be specified
* periods - indicates the number of subscription periods, which determines the size of the allocation for consumable resources.


psarando:
I don't see a Chromatic build associated with this PR. Is it enabled in your fork?

src/components/subscriptions/edit/formatters.js
```diff
@@ -31,13 +34,21 @@ export function mapPropsToValues(subscription) {
...
         paid,
+        periods,
+        start_date: start_date
+            ? formatDateObject(new Date(start_date), dateConstants.ISO_8601)
+            : undefined,
+        end_date: end_date
+            ? formatDateObject(new Date(end_date), dateConstants.ISO_8601)
+            : undefined,
     };
+    console.log(submission);
```
psarando, line 51: Looks like a debug log statement that can be removed now.

src/components/subscriptions/edit/formatters.js
```diff
@@ -31,13 +34,21 @@ export function mapPropsToValues(subscription) {
...
         username,
         plan_name,
         paid,
+        periods,
+        start_date: start_date
+            ? formatDateObject(new Date(start_date), dateConstants.ISO_8601)
+            : undefined,
+        end_date: end_date
+            ? formatDateObject(new Date(end_date), dateConstants.ISO_8601)
+            : undefined,
```
psarando, line 49: 👍

slr71:
Hmm, I thought it was. 🤔

src/components/subscriptions/edit/formatters.js
```diff
@@ -31,13 +34,21 @@ export function mapPropsToValues(subscription) {
 }
 
 export function formatSubscriptions(values) {
-    const { username, paid, plan_name } = values;
+    const { username, paid, periods, plan_name, start_date, end_date } = values;
```
psarando, line 37: Should these corresponding fields also be added to `mapPropsToValues` above? So these form fields get pre-populated from the API when editing subscriptions with existing start and end dates?

You can experiment with this in Storybook by adding the expected fields with mock values in the `subscription` object in `stories/subscriptions/SubscriptionMocks`.

slr71:
Ah, never mind. I had actions disabled on my fork to prevent the build actions from running.

slr71:
I just pushed a change to force a chromatic build.

psarando (changes requested):
Ah, yeah, I sometimes disable only the Chromatic action, so commits to my fork don't create unnecessary builds (using up some of our monthly snapshot quota), until I'm ready to open a PR.

This looks good so far, but I think I caught a bug for editing subscriptions with existing start and end dates.

src/components/subscriptions/edit/formatters.js
slr71: Hmm, that's a good question. Editing a subscription with the way that it's implemented currently simply creates a new subscription for the user and marks the old one as expired, so I'm not sure that filling in the start and end date would be helpful, although it might be good to change the wording from "Edit" to something else. 🤔 Do you have any suggestions for wording?

psarando: That's right. I was a little mislead by the `Edit` title of the dialog while writing my review, but I think it still makes sense from the UI's perspective that the admin is editing the user's subscription 🤔 

So populating the form with the existing start and end dates doesn't makes sense in that case.

Also, it's best to populate the form fields with something, even if it's just an empty string `""`. Although it doesn't seem to be an issue with these timestamp fields, I tried out this branch in Storybook, and when editing the `Subscription Periods` field, React reported this warning in the dev console: `Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://reactjs.org/link/controlled-components`.

So the `periods` field should be populated in `mapPropsToValues`, maybe with a `1` or at least `""`.

slr71: Sounds good. I can populate it with a `1`; that's the default value used by the services anyway.

psarando (approved):
LGTM 👍 🎉

PR #625, opened 2025-03-11
CORE-2034 Limit concurrent API calls from listings

psarando:
This PR will add the following:

* Added the p-limit package (v3 since v6 is not compatible with Jest).
* Added a `queriesConcurrencyLimit` config, with a default value of 8, which matches the default number of dashboard items that make similar API calls.
* Updated data listing metadata and LocalContexts project API calls with `pLimit` wrappers.
* Updated analyses listing history queries with `pLimit` wrappers.
    * This limits concurrent API calls not only to our analyses endpoints, but also to Tapis endpoints.
* Fixed metadata endpoint calls in data listing stories.


psarando:

PR #626, opened 2025-03-21
CORE-2034 Update data listing with new metadata search endpoint

psarando:
This PR will update the data listing so that each `ResourceNameCell` in the listing no longer queries the API for that item's metadata.

Instead, the top-level folder will use the new `POST /filesystem/metadata/search` endpoint (added in cyverse-de/terrain#305) to search for `LocalContexts` metadata for itself, and each item in its listing, all in one API call.


psarando:

PR #627, opened 2025-03-28
CORE-1963 Update Metadata Template to save directly to API

psarando:
This PR will update the Metadata Template form's `Save` button so that changes are saved directly to the API.

Also, on a metadata save success, the query cache for that data item's metadata is invalidated, so that the new metadata are reloaded from the API in these forms.


psarando:

PR #628, opened 2025-04-01
Global Alerts

ianmcorvidae:
I think this is ready for review, finally.

I'm using our included "warning" colors for the bar rather than a harsh red, now. Markdown is supported for formatting the header text, which allows adding things like links if desired as well.


src/components/alerts/admin/AlertsAdmin.js
```diff
@@ -0,0 +1,74 @@
...
+                        required={true}
+                    />
+                    <Field
+                        component={FormTextField}
+                        name="startDate"
+                        required={false}
+                    />
+                </Form>
+            )}
+        </Formik>
```
psarando, line 48: 👍 
Also consider using the https://github.com/cyverse-de/sonora/blob/main/src/components/forms/FormTimestampField.js for the start and end date fields (see https://github.com/cyverse-de/sonora/blob/main/src/components/subscriptions/edit/EditSubscription.js for example).

ianmcorvidae: Ah, good! I had failed to find that before, but looks like it'll do well for that. Thanks!

ianmcorvidae:
image

This is what the header looks like, in storybook.

slr71 (approved):
I didn't catch any issues.

psarando (changes requested):
Looks good overall 👍 

I did catch some issues with i18n messages, plus I had a few other suggestions.

src/components/layout/AdminDrawerItems.js
```diff
@@ -165,6 +165,21 @@ export default function AdminDrawerItems(props) {
...
+                    onClick={() =>
+                        router.push("/" + NavigationConstants.ADMIN + "/alerts")
+                    }
+                >
+                    <ListItemIcon>
+                        <LabelImportantIcon className={classes.icon} />
+                    </ListItemIcon>
+                    {open && <ListItemText>{"alerts"}</ListItemText>}
+                </ListItem>
+            </Tooltip>
```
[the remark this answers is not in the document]
ianmcorvidae: Ah, thanks. Forgot to go back to the layout and i18n it after I (mostly, heh) did it on the component itself.

src/components/alerts/admin/AlertsAdmin.js
```diff
@@ -0,0 +1,270 @@
...
+
+import WrappedErrorHandler from "components/error/WrappedErrorHandler";
+
+import { useTranslation } from "i18n";
+import { Field, Form, Formik } from "formik";
+
+import {
+    allAlerts,
+    ALL_ALERTS_QUERY_KEY,
+    ACTIVE_ALERTS_QUERY_KEY,
```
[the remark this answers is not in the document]
ianmcorvidae: Ah, good catch. I should actually be invalidating that query when the lists of alerts change, so the person editing will see the correct stuff at the top of their own screen right away.

src/pages/admin/alerts.js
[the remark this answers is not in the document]
ianmcorvidae: I think I've got this set up properly, hopefully.

src/components/alerts/admin/AlertsAdmin.js
```diff
@@ -0,0 +1,270 @@
...
+        <Formik initialValues={initialValues} onSubmit={handleSubmit}>
+            {(props) => (
+                <Form className={classes.flexContainer}>
+                    <Field
+                        component={FormTextField}
+                        name="alertText"
+                        label="Alert Text"
+                        required={true}
+                        className={classes.flexItem}
+                    />
```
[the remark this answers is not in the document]
ianmcorvidae: Yeah, that's looking a lot cleaner. I'm pushing that up shortly.

src/components/alerts/admin/AlertsAdmin.js
```diff
@@ -0,0 +1,253 @@
...
+                        alignItems="flex-start"
+                        justifyContent="flex-start"
+                    >
+                        <Field
+                            component={FormTextField}
+                            name="alertText"
+                            label="Alert Text"
+                            required={true}
+                            sx={{ minWidth: "25ch", width: 0 }}
+                        />
```
psarando, line 64: This can wait for a future PR, but this text field is kind of small, especially if the message gets kind of long with markdown formatting and links.

It might be helpful to allow this text field to stretch full-width across the top of everything else, and a Markdown preview (maybe with the yellow background) might be useful too.

The existing alerts listing could also use Markdown rendering.

Also, there isn't enough room for all these fields to say side-by-side in mobile view. I'm wondering if a Grid might allow them to flow / wrap in small screens.

---
Long text and a URL:

Long text and a URL

---
Mobile view:

Mobile view

psarando (approved):
LGTM 👍 

I had a few more suggestions and nitpicks, but nothing blocking, so feel free to tweak what you want, but we can :shipit: for now and do more improvements in a future PR(s).

src/components/alerts/admin/AlertsAdmin.js
```diff
@@ -0,0 +1,253 @@
...
+                            variant="rectangular"
+                            animation="wave"
+                            height={150}
+                            width="100%"
+                        />
+                    ) : (
+                        <AddAlertForm t={t} handleSubmit={handleSubmit} />
+                    )}
+                    {isFetchingList || isMutating ? (
+                        <Skeleton
```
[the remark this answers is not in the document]
ianmcorvidae: Yeah, I think you're right. I'd initially done a thing where it masked different things depending on what was loading, but in the end it doesn't make much sense to show the form without the list (which is what usually happened) or vice versa, so it's all ended up using the same stuff. I'll fold it all into one isQueryLoading.

src/components/alerts/admin/AlertsAdmin.js
ianmcorvidae: Hm, yeah, good thoughts. I was definitely struggling with the formatting stuff and how to get it working nicely.

With the text box, I just had a thought and I'm not sure how hard it would be, but maybe it would work well to have that field be the sort of expanding text field that among other places is used for GitHub comments. Then we can show it fairly small while it's empty, but let it expand downward as users type stuff. I do think it could stand to be wider as well though; I went with the same width as the timestamps because initially it was stretching so wide it was making _those_ fields unusable. At the least I can probably make it a bit wider.

Adding the markdown rendering to the table display wouldn't be too difficult I think, that's a good idea. I'm less sure how to do it cleanly within the table but I'll think about it. I think you're right, just not sure how to do it, haha.

For a Grid, I guess what we'd do is change where it does lines depending on breakpoints. That's probably feasible, but I might leave it for this PR as you suggest, since the formatting stuff has been my kryptonite on this PR a bit, haha.

ianmcorvidae:
Side note: while poking around I remembered there's a group of things I forgot to add i18n to, so I'll be updating that. Mariah also had one more graphical update for the color of the alert in the header (a slightly less eye-searing orange rather than the "warning" yellow color), which I'll integrate.

src/components/alerts/admin/AlertsAdmin.js
ianmcorvidae: I realized we've already got FormMultilineTextField, so I put that in there along with doubling the min-width. It's still not good on mobile, but I think that's a problem for another time.

ianmcorvidae:
Okay. Maybe finally up to a final mergeable state!

src/components/alerts/admin/AlertsAdmin.js
psarando: The `FormMultilineTextField` is also just a convenience wrapper around a `FormTextField` with a `multiline` prop set:
https://github.com/cyverse-de/sonora/blob/c1e49799889a244cd2255423d083c666407a790b/src/components/forms/FormMultilineTextField.js#L7-L9
So this field could still use `component={FormTextField}` with a `multiline` prop added, but without the `minRows={3}` (or maybe `minRows={1}` would be required, I can't remember). Then it would look like a single-line text field and expand as required; but this `FormMultilineTextField` LGTM, too 👍

psarando: I didn't explicitly mention it, but I guess I was envisioning the form fields stacked in a column, rather than a single row. That would allow the Alert text field to expand full-width. Although I do like the start and end dates side-by-side (especially with their current layout), so they could be wrapped in an inner Stack with `direction="row"`. Then a Markdown preview could go between the Alert text field and the date fields.

But all this tweaking can be a for a future PR.

psarando (approved):
LGTM 🎉 👍

ianmcorvidae:
Thanks for the reviews and rereviews! This one took me a while, but merging now.

PR #632, opened 2025-05-30
CORE-2050 Disable Sharing for Users with Basic Subscriptions

psarando:
This PR will update the `Share` button for apps, analyses, and data to display a warning announcer, with a link to the subscription page, for users that only have unpaid `Basic` subscriptions, and no CPU or Data add-ons.

This also restores the data listing sharing buttons for unpaid `Basic` subscriptions, but will display the warning message.

This PR also refactors the queries for fetching a user's resource usage summary into a common `useResourceUsageSummary` hook, which also returns common flags and usage values.

--- 
Warning Announcer message

### DIFF
--- a/public/static/locales/en/common.json
+++ b/public/static/locales/en/common.json
@@ -86,6 +86,7 @@
     "select": "Select",
     "selectAllLabel": "Select All",
     "settings": "Settings",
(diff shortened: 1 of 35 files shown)


psarando:
Thanks for the review! 👍 

I'll go ahead and merge this now so it can get tested in QA before the release, but additional reviews will still be welcomed 🙇

PR #631, opened 2025-05-21
CORE-2053: fixed an issue with some API requests failing with a 503

slr71:

PR #633, opened 2025-06-04
CORE-2053 Also stream error responses in server/api/terrain

psarando:
This PR will update `server/api/terrain` to also stream error responses from `terrain`, since changes from #631 now stream OK responses.


slr71:
I'm going to merge this now so that we can get it in QA right away.

PR #635, opened 2025-06-05
CORE-2050 Fixes for the Resource Usage Summary query

psarando:
This PR will add the following fixes:

* Restore the `APP_LAUNCH_RESOURCE_USAGE_QUERY_KEY` removed by #632.
  * This will prevent the app launch forms from being masked (which will reset the form and lose all progress) when an input selector's data listing gets masked by the Resource Usage Summary query.
* Fix a bug that displays an error fetching the summary for logged out users.
  * The `config?.subscriptions?.enforce` flag might return `undefined` initially, but the query's `enabled` flag must be a `boolean`, so this query would run for logged out users and display an error.

### DIFF
--- a/src/common/useResourceUsageSummary.js
+++ b/src/common/useResourceUsageSummary.js
@@ -44,7 +44,7 @@ function useResourceUsageSummary(
     } = useQuery({
         queryKey: [queryKey, userProfile?.id],
         queryFn: getResourceUsageSummary,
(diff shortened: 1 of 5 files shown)


psarando:
I'll merge this now, but reviews are still welcome 🙇

slr71 (approved):
LGTM! 👍 :shipit:

psarando:
Thanks for the review! 👍

PR #636, opened 2025-06-06
Refactor Global Alerts out of AppBar and into top-level PageWrapper

psarando:
This PR will refactor rendering the active alerts into a `GlobalAlerts` component that is nested inside the top-level `PageWrapper`, just above each page's content.

Since rendering alert markdown is async, the AppBar's height is used for setting the top-level PageWrapper's `appBarHeight` prop before the alert messages are rendered, which caused the bottom of pages to be cut off.

Dashboard with 2 Alerts

App Listing with 2 Alerts


psarando:

PR #637, opened 2025-06-12
CORE-2056 Move Global Alerts from every page to only on Dashboard

psarando:
The Global Alerts now take the place of the static Data Storage Limits Announcement.

Global Alerts in Dashboard


slr71 (approved):
Looks good to me! :shipit:

PR #638, opened 2025-06-13
CORE-2057 Fix Admin EditTool form

psarando:
This PR will fix a bug in the Admin EditTool form, where an admin attempted to edit a tool and a null tool error in the service facade prevented the PATCH request from being submitted.

The `handleSubmit` function was setting the formatted values in a `toolSubmission` state, then immediately calling `updateCurrentTool` or `addNewTool` mutation queries, which were trying to access that state, but the state values were not available yet.

This PR also includes other minor fixes, such as fixing warnings and the tool edit success announcement.

### DIFF
--- a/src/components/forms/FormSelectField.js
+++ b/src/components/forms/FormSelectField.js
@@ -18,7 +18,7 @@ const FormSelectField = ({
     required,
     form: { touched, errors },
     children,
-    fullWidth,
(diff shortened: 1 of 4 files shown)


psarando:
I'm going to merge this right away, but reviews are still welcome 🙇

PR #671, opened 2026-04-04
CORE-2123 Fix for `useQuery` in pages for `react-query` v4 upgrade

psarando:
It seems the `onSuccess` callback for `useQuery` is working differently now in `react-query` v4, beyond just being deprecated.

This seems to have only affected the App Editor, for some reason, which was retaining the `undefined` app description in its form, and displaying blank form fields even after the app was fetched.

I also noticed that the `subscriptions` page was incorrectly setting the query loading and error values, so I cleaned up that query as well.

### DIFF
--- a/src/pages/admin/subscriptions.js
+++ b/src/pages/admin/subscriptions.js
@@ -54,39 +54,23 @@ export default function Subscriptions() {
     const onTabSelectionChange = (_event, selectedTab) => {
         setSelectedTab(selectedTab);
     };
-    const [availableAddons, setAvailableAddons] = useState(null);
-    const [getAddonsQueryEnabled, setGetAddonsQueryEnabled] = useState(false);
     const queryClient = useQueryClient();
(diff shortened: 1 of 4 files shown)


src/pages/admin/subscriptions.js
```diff
@@ -54,39 +54,23 @@ export default function Subscriptions() {
...
+        isFetching: isFetchingAvailableAddons,
+        error: errorFetchingAvailableAddons,
+    } = useQuery({
+        queryKey: [AVAILABLE_ADDONS_QUERY_KEY],
+        queryFn: getAvailableAddOns,
+        enabled: !availableAddonsCache,
+        staleTime: Infinity,
+        cacheTime: Infinity,
+    });
+    const availableAddons = availableAddonsCache || data;
```
slr71, line 73: 👍

slr71 (approved):
Looks good to me. The updates are really clean. ✨

psarando:
Thanks for the review! 👍

PR #673, opened 2026-04-09
CORE-2123 Replace react-query `status` usage with `isFetching`

psarando:
In react-query v4, `status` is always initially `loading`, even when the query is disabled and has no data.

This PR also includes a fix in `PathListViewer` for the use of the data viewer toolbar's `onSaving` prop
(making it consistent with the other TextViewers).

### DIFF
--- a/src/components/data/viewers/Toolbar.js
+++ b/src/components/data/viewers/Toolbar.js
@@ -9,6 +9,7 @@ import React, { useEffect, useState } from "react";
 
 import { useMutation, useQueryClient, useQuery } from "@tanstack/react-query";
 import { useTranslation } from "i18n";
+import constants from "../../../constants";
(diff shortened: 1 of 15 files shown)


src/components/data/viewers/Toolbar.js
```diff
@@ -163,7 +164,7 @@ function ViewerToolbar(props) {
         },
     });
 
-    const { mutate: saveTextAsFile, isLoading: fileSaveLoading } = useMutation(
+    const { mutate: saveTextAsFile, status: fileSaveStatus } = useMutation(
```
psarando, line 167: Turns out this is no longer used outside of this file (and it wasn't used correctly before), and `isLoading` for mutations is now deprecated. So I switched this to the non-deprecated `status`.

psarando:

PR #676, opened 2026-04-10
CORE-2123 Replace `onSuccess` and `onError` callbacks in pages

psarando:
Since the `onSuccess` callback for `useQuery` is working differently now in `react-query` v4, beyond just being deprecated, I finished replacing the `onSuccess` and `onError` callbacks in the remaining pages that still used them.


psarando:
I'll merge this now, but reviews are still welcome 🙇

PR #684, opened 2026-05-06
Only show GPUs selector during app launch when relevant

ianmcorvidae:
when the maximum is higher than 0 and the minimum and maximum aren't equal, specifically


psarando (approved):
LGTM 👍 

Though I'm not seeing a Chromatic build associated with this PR.
You may have to still set up your Chromatic token: https://github.com/cyverse-de/sonora/blob/main/tech-stack.md#chromatic-setup

PR #691, opened 2026-05-11
Add VICE Operators admin tab

johnworth:
## Summary
- Adds a new "Operators" tab to the VICE admin page for managing registered VICE operators (list, sorted by priority; add, edit). Backed by terrain's `/api/admin/vice/operators` endpoints.
- Wires the matching Express proxy routes (`GET`/`POST /admin/vice/operators`, `PATCH /admin/vice/operators/id/:id`) through to terrain.
- Includes service facade, components, translation strings, and IDs for the new tab.

## Test plan
- [ ] Start the dev server and load `/admin/vice`.
- [ ] Click the Operators tab; verify the list renders existing operators with no 404.
- [ ] Add a new operator and verify it appears in the list.
- [ ] Edit an operator's priority/URL/TLS-skip flag and verify the update persists.
- [ ] Confirm validation errors surface inline in the add/edit form.


public/static/locales/en/vice-admin.json
psarando, line 1: We have been trying to keep the keys in these files sorted alphabetically.

src/components/vice/admin/operators/OperatorEditor.js
psarando, line 1: It appears Claude rolled its own Formik here.

I think it would be more consistent with the rest of the codebase, and more maintainable, if this component used `formik`, `components/forms/FormTextField`, and `components/forms/FormSwitch` as well.

src/components/vice/admin/operators/OperatorEditor.js
```diff
@@ -0,0 +1,229 @@
+/**
+ * VICE Admin — Operators editor dialog.
+ *
+ * Add / edit form for a registered operator. Live validation: every keystroke
+ * is run through `validateOperator`, errors are translated and surfaced via
+ * MUI TextField helperText. The Save button is disabled until the draft is
+ * valid.
+ *
+ * @author your-name
```
psarando: *Your* name here 😆

src/serviceFacades/vice/admin.js
```diff
@@ -91,6 +91,48 @@ export const userExists = ({ username }) => {
...
+});
+
+// Translate a camelCase draft from the editor back to the snake_case shape
+// terrain accepts.
+const toTerrain = ({ name, url, priority, skipTlsVerify }) => ({
+    name,
+    url,
+    priority,
+    tls_skip_verify: skipTlsVerify,
+});
```
psarando: These are unnecessary if the form field uses the snake case `tls_skip_verify` for its name (which is not an issue in JS).

src/components/vice/admin/operators/index.js
```diff
@@ -0,0 +1,441 @@
+/**
+ * VICE Admin — Operators tab.
+ *
+ * Lists registered VICE operators sorted by priority (ascending — lower value
+ * means higher position in the list), and allows the admin to register new
+ * operators or edit existing ones. Backed by terrain's
+ * `/api/admin/vice/operators` endpoints (terrain PR #323).
+ *
+ * @author your-name
```
psarando: *Your* name here 😆

src/components/vice/admin/operators/index.js
```diff
@@ -0,0 +1,441 @@
...
+                    <WrappedErrorHandler errorObject={error} baseId={BASE_ID} />
+                ) : isLoading ? (
+                    <Stack spacing={1}>
+                        <Skeleton variant="rectangular" height={48} />
+                        <Skeleton variant="rectangular" height={48} />
+                        <Skeleton variant="rectangular" height={48} />
+                    </Stack>
+                ) : (
+                    <TableContainer>
+                        <Table id={id(ids.TABLE)} size="small">
```
psarando: We also have a `components/table/TableLoading` component, but it returns a `TableBody`. So it would have to render inside the `Table`, below the `TableHead`, in place of the `TableBody` below.

The `TableLoading` lets the actual column headers display from the `TableHead` while the body data loads.

psarando (changes requested):
This is looking pretty good so far. I bet some of the styles are excessive, but I like the way it looks. 👍 

There is still some excessive code I think we can cut down on.

src/components/vice/admin/operators/OperatorEditor.js
```diff
@@ -0,0 +1,164 @@
...
+    // the plain message strings Formik expects.
+    const validate = (values) => {
+        const errs = validateOperator(values, allOperators, initial?.id);
+        return Object.fromEntries(
+            Object.entries(errs).map(([field, { key, values: v }]) => [
+                field,
+                t(key, v),
+            ])
+        );
+    };
```
psarando: A pattern we used in other components is to pass `t` to the `validateOperator` function, and then it can return the `errs` object with translated string values.

psarando (commented):
Much better! 🎉 
I just had one more nit to pick, otherwise LGTM 👍

src/components/vice/admin/operators/OperatorEditor.js
johnworth: Good call — done in 6873ed07. `validateOperator` now takes `t` as its first argument and returns the translated `errs` strings directly, so the extra translation pass in `OperatorEditor` is gone.

psarando (approved):
LGTM 👍

PR #695, opened 2026-05-14
Expand sonora Justfile to mirror terrain's recipe surface

johnworth:
## Summary
- Adds `dev`, `run`, `lint`, `fmt-check`, `fmt-fix`, `check`, `storybook`, and `build-storybook` recipes so the common npm scripts are reachable through `just`, matching the recipe surface of terrain's Justfile.
- `check` runs `fmt-check lint test`, mirroring the npm `check` script.
- Normalizes `{{ build-context }}` spacing and drops a stray trailing tab on the `build` recipe.

## Test plan
- [x] `just --list` parses and shows all 16 recipes.
- [ ] Spot-check a recipe or two locally (e.g. `just lint`, `just fmt-check`).

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -8,17 +8,20 @@ build-flags := ""
 
 default: build
 
-test:
-    npm run test
-
 deps:
     npm install
 
 build: deps
-    npm run build	
+    npm run build
+
+dev:
+    npm run dev
+
+run:
+    npm start
 
 build-image:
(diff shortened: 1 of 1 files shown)


Justfile
```diff
@@ -8,17 +8,20 @@ build-flags := ""
...
-
 deps:
     npm install
 
 build: deps
-    npm run build	
+    npm run build
+
+dev:
+    npm run dev
```
slr71, line 18: I initially wondered why more recipes didn't depend on `deps`, but since `just` is different from `make`, it occurs to me that that might add more time to some recipes that are used frequently.

slr71 (approved):
Looks good to me! 👍

PR #696, opened 2026-05-18
Make it so min_gpus > 0 correctly excludes 0 from the dropdown

ianmcorvidae:
Just an AI-assisted oversight on my part.

### DIFF
--- a/src/components/apps/launch/ResourceRequirements.js
+++ b/src/components/apps/launch/ResourceRequirements.js
@@ -66,6 +66,12 @@ function buildLimitList(startValue, minValue, maxValue) {
     return limits;
 }
 
+function buildGpuLimitList(minValue, maxValue) {
+    const limits = buildLimitList(1, minValue, maxValue);
+
+    return minValue > 0 ? limits.filter((value) => value !== 0) : limits;
+}
+
 /**
  * Form fields for selecting a step's resource requirements.
  *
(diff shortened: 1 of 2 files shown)


src/components/apps/launch/ResourceRequirements.js
```diff
@@ -66,6 +66,12 @@ function buildLimitList(startValue, minValue, maxValue) {
     return limits;
 }
 
+function buildGpuLimitList(minValue, maxValue) {
+    const limits = buildLimitList(1, minValue, maxValue);
+
+    return minValue > 0 ? limits.filter((value) => value !== 0) : limits;
+}
```
psarando, line 73: The other resource selection lists always include a 0 so that the user can select it in order to clear the selection (so nothing is submitted for that setting in the request), which allows the backend to choose its own default minimum.

I guess for GPUs, if a minimum is set, it must be submitted with the job from the frontend, and the backend doesn't choose a default? (Sorry if this was already answered before).

psarando (approved):
LGTM 👍

src/components/apps/launch/ResourceRequirements.js
ianmcorvidae: Hm, that's an interesting question. In the case where a minimum > 0 is set, I'm not sure what default the backend would set, though I believe it would do one (for example, it knows to pick within the appropriate range when the selector is not shown due to minimum and maximum being equal). I'll check on that -- I think the proper behavior would probably be for it to choose the minimum acceptable value, though, which might be what this effectively accomplishes.

ianmcorvidae: Thinking about this more, I think it's fine/expected that users won't be able to fall back to the backend setting the value. It'll make sure that the value is within the min/max, but it's not quite like CPU/memory where 0 is a nonsense value (0 GPUs is possible and even common), so I don't really want 0 to show when it's marked as an invalid value for the tool. If it ends up confusing to people, we could maybe add in an explicit way to clear the field at some later date, but for now I think this is fine.

psarando: Sounds good to me. Selecting 0 to clear the selection might not be the best UX, but it was fine when 0 didn't mean anything for the other resources. So I agree it can be updated in a later PR if it's confusing.

PR #699, opened 2026-05-28
Make it possible for min/max GPUs and models to appear in the user-facing dialog

ianmcorvidae:
(not just the admin one)

This will also need a bit of config changes to set the `tools.private.max_gpu_limit`, I believe.


src/components/tools/edit/formatters.js
psarando: Looks like this `gpu_models` setting is already included outside of this `if (isAdmin)` check, so this one is can be removed here.

psarando (approved):
LGTM 👍 

I did catch one DRY issue that's not directly from the changes in this PR.

slr71 (approved):
Looks good to me!

src/components/tools/edit/formatters.js
ianmcorvidae: Somehow missed this review. I'll fix up the DRY issue on Monday, good catch there.

PR #701, opened 2026-06-03
Reload onto maintenance page when API returns a maintenance 503

johnworth:
## Summary

When the gateway enters maintenance mode it responds to API/XHR requests with a 503. Previously these surfaced as generic error UI inside the SPA instead of taking the user to the gateway's maintenance page. This branch detects the maintenance-mode 503 on the client and forces a hard navigation so the browser re-requests the page as a document and lands on the maintenance page.

## Changes

- **`src/common/getAxios.js`**: Add a browser-only response interceptor on the shared axios instance. On a 503 carrying `x-maintenance-mode: true` (or `{ maintenance: true }` body), it sets a one-shot `window.__deMaintenanceReloading` guard and calls `window.location.reload()`, returning a never-resolving promise so no error UI flashes first. Guarded by `typeof window !== "undefined"` because the same instance is reused server-side to proxy to Terrain, where a 503 must propagate normally.
- **`src/components/uploads/api.js`**: Mirror the same behavior in `startUpload`, which uses `fetch` rather than the axios instance, so maintenance-mode 503s during uploads also redirect to the maintenance page.

## Notes

The `window.__deMaintenanceReloading` guard prevents many concurrent in-flight calls from each triggering a reload.

### DIFF
--- a/src/common/getAxios.js
+++ b/src/common/getAxios.js
@@ -4,6 +4,8 @@
  */
 import axios from "axios";
 
+import { isMaintenanceResponse, triggerMaintenanceReload } from "./maintenance";
+
 const axiosInstance = axios.create({
     timeout: 20000, //service call timeout
 });
@@ -16,18 +18,10 @@ if (typeof window !== "undefined") {
     axiosInstance.interceptors.response.use(
         (resp) => resp,
         (error) => {
-            const resp = error?.response;
-            const isMaintenance =
-                resp &&
-                resp.status === 503 &&
-                (resp.headers?.["x-maintenance-mode"] === "true" ||
-                    resp.data?.maintenance === true);
-
-            if (isMaintenance && !window.__deMaintenanceReloading) {
-                // Guard against many concurrent in-flight calls triggering repeat reloads.
-                window.__deMaintenanceReloading = true;
-                window.location.reload();
-                // Never resolve so no error UI flashes before the page reloads.
+            if (error?.response && isMaintenanceResponse(error.response)) {
+                triggerMaintenanceReload();
+                // Never resolve so no error UI flashes before the page reloads,
+                // even for concurrent calls that arrive after the reload starts.
                 return new Promise(() => {});
             }
             return Promise.reject(error);
--- a/src/common/getAxios.js
+++ b/src/common/getAxios.js
@@ -7,4 +7,32 @@ import axios from "axios";
 const axiosInstance = axios.create({
     timeout: 20000, //service call timeout
 });
+
+// Detect gateway maintenance mode on API/XHR responses and force a hard navigation
+// so the browser re-requests the page as a document and lands on the maintenance
+// page served by the gateway. Browser-only: this same instance is reused server-side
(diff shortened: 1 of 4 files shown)


slr71 (approved):
I didn't spot anything that looked concerning. Claude mentioned a few things in its review, but none of them seem like a serious problem to me. Here are the suggestions from Claude:

>   1. terrain.js — forwarding all upstream headers on the error path (medium).
>   if (e.response?.headers) {
>       res.set(e.response.headers);
>   }
>   This forwards every upstream header, including hop-by-hop ones (transfer-encoding, connection, content-length, content-encoding). Since the body is re-streamed chunk-by-chunk
>   through Express, a forwarded content-length/transfer-encoding/content-encoding from upstream could mismatch the re-emitted bytes and corrupt or hang the response. The success path
>   at line 41 already does the same res.set(apiResponse.headers), so this is consistent with existing behavior — but the error path is new territory and worth a sanity check that
>   real 503 responses don't carry a conflicting content-encoding/content-length. If you only need the maintenance signal to survive, forwarding just x-maintenance-mode (and maybe
>   content-type) would be safer and more surgical than the whole header set.

>   2. Potential reload loop risk (worth confirming).
>   The __deMaintenanceReloading guard is a window global, so it resets on every fresh page load. The design relies on the gateway returning the maintenance HTML page for the document
>   request after reload. If the gateway instead still serves the SPA shell for document requests during maintenance, the SPA reloads → makes an API call → gets 503 → reloads again,
>   in a tight loop. This is almost certainly fine given the gateway's intended behavior, but it's the one assumption the whole approach rests on — worth a one-line confirmation that
>   document requests get the static maintenance page during maintenance.

>   3. No test coverage.
>   isMaintenanceResponse and getHeader are pure functions and trivially unit-testable (503 + header, 503 + body, 503 with neither, non-503, fetch Headers vs plain object). Given this
>   is error-path code that's hard to exercise manually, a small unit test for maintenance.js would be valuable and low-cost.

>   4. PR description is out of date (minor).
>   The "Changes" section lists only getAxios.js and uploads/api.js, but the implementation extracted a new src/common/maintenance.js module and added the src/server/api/terrain.js
>   header-forwarding change. Updating the description would help reviewers.

>   5. Dangling never-resolving promise (minor / acceptable).
>   return new Promise(() => {}) intentionally leaves callers (and any react-query retries) hanging until the reload tears down the page. This is the intended trick and is fine, but
>   it does mean any finally/cleanup in callers won't run. Acceptable given the page is being destroyed.

For 1, this matches the success path behavior. It might be worth follow-up up later, but it's not something that needs to be addressed now.

I don't think that 2 is a real problem because the HTTPRoute update is the thing that causes maintenance responses to be returned.

For 3, test coverage could be useful to ensure that future changes are working as intended, but they're not strictly necessary.

For 4, the outdated PR description isn't a huge deal.

For 5, this is kind of intended, so I think it's fine.

psarando (commented):
LGTM 👍

PR #703, opened 2026-06-04
Add base_url, accepting_launches, and deactivated to VICE operators

johnworth:
## Summary

Surfaces terrain's newer VICE operator fields in the admin **Operators** tab (`src/components/vice/admin/operators/`).

- **List:** new **Status** column showing an *Accepting* / *Paused* chip plus a *Deactivated* chip, so both flags are always visible.
- **Add/Edit dialog:** supports an optional `base_url` field. The `accepting_launches` and `deactivated` switches are **edit-only**, since terrain's create (`POST`/`OperatorConfig`) schema does not accept them — new operators take server defaults.
- **Validation:** `base_url` is validated client-side (optional, must be http/https) via a shared `validateHttpUrl` helper reused by both URL fields, and is submitted as `null` when blank (the schema is `maybe NonBlankString`, so `""` would be rejected).

No changes were needed to the service facades or Express proxy routes — both pass the request body through generically.

## Related

- Backend support for `accepting_launches` and `deactivated`: cyverse-de/terrain#324
- `base_url` and the admin operator APIs landed earlier in cyverse-de/terrain#323

> **⚠️ Merge order:** cyverse-de/terrain#324 must merge first. The `accepting_launches` and `deactivated` toggles depend on the backend support it adds.

## API contract (terrain `src/terrain/routes/schemas/vice.clj`)

- `POST /admin/vice/operators` (`OperatorConfig`): adds optional `base_url`; does **not** accept `accepting_launches`/`deactivated`.
- `GET /admin/vice/operators` (`OperatorAdminSummary`): includes all three.
- `PATCH /admin/vice/operators/id/{id}` (`UpdateOperatorRequest`): all fields optional, including `accepting_launches`/`deactivated`.

## Testing

- `eslint` clean (zero warnings), `prettier` clean, full jest suite green (189 tests).
- Manually verified end-to-end against a running terrain/app-exposer: list status chips reflect API data; create dialog shows `base_url` but not the two toggles; edit dialog exposes all three; clearing `base_url` sends `null`.


psarando (approved):
LGTM 👍 

Though I noticed that the `stories/vice/admin/VICEAdmin.stories.js` is displaying a 404 error in the operators tab for the `/api/admin/vice/operators` endpoint.

I didn't think to check this in #691, but some mock data can be added to `stories/vice/admin/mocks.js` and used with `mockAxios` in this story, then we can also preview changes to these components in Chromatic.

psarando (approved):
LGTM 👍

src/components/vice/admin/operators/OperatorEditor.js
```diff
@@ -145,6 +163,24 @@ const OperatorEditor = ({
...
+                                    component={FormSwitch}
+                                    label={t("operatorAcceptingLaunches")}
+                                    color="success"
+                                />
+                                <Field
+                                    id={id(ids.DIALOG_DEACTIVATED)}
+                                    name="deactivated"
+                                    component={FormSwitch}
+                                    label={t("operatorDeactivated")}
+                                    color="warning"
```
psarando: This is really picky (so feel free to merge this PR as is), but the `Deactivated` chip in the listing is using `color="error"`, so I would use the same color here 😁

johnworth: Good catch — switched it to `color="error"` to match the listing chip in bbe0d9d4. Thanks!

psarando (approved):

PR #705, opened 2026-06-11
Make the publication requests be lazy-loaded when the tab is selected

ianmcorvidae:
(rather than loaded with the whole admin page)

Minor performance improvement and the apps and metadata changes will make it matter less, but it still seems wasteful to request it when a lot of the time the publication requests won't be relevant.

### DIFF
--- a/src/components/apps/admin/publicationRequests/RequestListing.js
+++ b/src/components/apps/admin/publicationRequests/RequestListing.js
@@ -135,7 +135,7 @@ function ToolsUsed(props) {
     );
 }
 function AppPublicationRequests(props) {
-    const { parentId } = props;
+    const { parentId, tabIsActive } = props;
     const { t } = useTranslation("apps");
 
     const [requests, setRequests] = React.useState();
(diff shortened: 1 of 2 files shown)


slr71 (approved):
I didn't spot any issues. 👍

src/components/apps/admin/publicationRequests/RequestListing.js
```diff
@@ -146,7 +146,7 @@ function AppPublicationRequests(props) {
     const { isFetching: isRequestsFetching } = useQuery({
         queryKey: [APP_PUBLICATION_REQUESTS_QUERY_KEY],
         queryFn: getAppPublicationRequests,
-        enabled: true,
+        enabled: !!tabIsActive,
```
psarando, line 149: Looks like the `tabIsActive` prop is already a boolean, so the `!!` is probably redundant here, but it doesn't hurt either (especially if `tabIsActive` changes later).

psarando (approved):
LGTM 👍

PR #713, opened 2026-07-21
Add mount data store toggle to VICE app launch Advanced Settings

ianmcorvidae:
Add a checkbox to the Advanced Settings step of the app launch wizard, visible only for interactive (VICE) apps. The checkbox is labelled "Mount data store in analysis" and defaults to checked (true), preserving existing behaviour. Helper text explains the difference between the direct-mount and copy-in/copy-out modes.

Wire mount_data_store into initAppLaunchValues (default true) and formatSubmission so it is included in the POST body to /analyses.

depends on https://github.com/cyverse-de/app-exposer/pull/160 https://github.com/cyverse-de/apps/pull/301 and https://github.com/cyverse-de/model/pull/29

### DIFF
--- a/public/static/locales/en/launch.json
+++ b/public/static/locales/en/launch.json
@@ -32,6 +32,8 @@
     "minDiskSpace": "Minimum Disk Space",
     "minCPUCores": "Minimum CPU Cores",
     "minMemory": "Minimum Memory",
+    "mountDataStore": "Mount data store in analysis",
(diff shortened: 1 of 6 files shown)


psarando (approved):
LGTM 👍 

Though I would also like to see the `stories/apps/launch/JupyterLabNoParams.js` story updated (or probably just the `stories/apps/launch/data/JupyterLabNoParamsApp.js` with the `interactive` type field) so that this checkbox can be previewed in Storybook.

psarando (approved):
Works as expected in Storybook now 👍 🎉

PR #714, opened 2026-07-23
CORE-2170 Add a custom request timeout for the `/api/uploads` endpoint

psarando:
This PR will add a custom request timeout for the `/api/uploads` endpoint, by disabling the `express` http server's global 5 minute request timeout, and adding express middleware body request timeouts: a longer timeout for the uploads endpoint (1 hour by default), and a reasonable timeout for all other routes (5 minutes by default).


slr71 (approved):
Looks good to me. 👍

PR #718, opened 2026-07-27
Fix client-side crash after moving or deleting a selected file

johnworth:
## Problem

Moving or deleting an item out of the folder being viewed refreshes the listing without changing `path`, so the effect that clears `selected` never fires and the removed item's ID lingers in the selection.

`getSelectedResources` resolved those stale IDs with `Array.find`, which returns `undefined`, and `formatSharedData` then destructured them during render:

    TypeError: Cannot destructure property 'label' of 'undefined'

That threw into the Next.js error boundary and unmounted the app, which looked like an unexpected logout even though the move had succeeded and the session was still valid.

A first pass filtered unresolvable IDs out of `getSelectedResources` and pruned `selected` in an effect. That was not enough — an effect runs after render, so the render in between still saw a `selected` containing an ID the refreshed listing no longer had. It also left `selected.length` and `getSelectedResources().length` free to disagree, which is the real hazard: `DataDotMenu` guards on `selected?.length === 1` but indexes `selectedResources[0]`, so the QA build traded the old crash for

    TypeError: can't access property "type", ef[0] is undefined

## Fix

- Derive `listedSelected` with `useMemo` during render in `Listing.js` and read it everywhere `selected` was read. The two lengths now agree by construction and no render can observe a stale ID.
- Drop the pruning effect — it was syncing state that is better derived.
- Guard the `selectedResources[0]` access in `DataDotMenu`, matching the optional chaining already used for `permission` two lines above.

## Tests

New `src/__tests__/dataDotMenu.js` covers a selection that outlives its listing, including the case where more IDs are selected than resolve to resources. Verified the test fails against the unguarded `DataDotMenu` and passes with the fix.

The shared `I18nProviderWrapper` loads locales over HTTP and suspends forever under jest, so the test stubs `useTranslation` and `useRouter` and asserts the component actually rendered — without those stubs the component never runs and the test proves nothing.

## Verification

- `npm run lint` — clean (zero warnings)
- `TZ=UTC npx jest` — 193 tests across 31 suites pass
- `npm run build` — succeeds
- Prettier check clean on all changed files

### DIFF
--- a/src/__tests__/dataDotMenu.js
+++ b/src/__tests__/dataDotMenu.js
@@ -1,0 +1,96 @@
+/**
+ * @jest-environment jsdom
+ */
+import React from "react";
+import preloadAll from "jest-next-dynamic";
+import renderer from "react-test-renderer";
+
+import DataDotMenu from "components/data/toolbar/DataDotMenu";
+import ResourceTypes from "components/models/ResourceTypes";
+
+import { EmotionCacheProvider } from "__mocks__/EmotionCacheProvider";
+import { RQWrapper } from "__mocks__/RQWrapper";
+import { ConfigProvider } from "contexts/config";
+
+import { ThemeProvider } from "@mui/material/styles";
+import theme from "components/theme/default";
+
+// The shared I18nProviderWrapper loads locales over HTTP and suspends forever
+// under jest, so anything rendered inside it never runs. Stub the hook instead
+// so this component actually renders and its logic is exercised.
+jest.mock("i18n", () => ({
+    useTranslation: () => ({ t: (key) => key }),
+    Trans: ({ children }) => children,
+    i18n: {},
+    RequiredNamespaces: [],
+}));
+
+// The component calls useRouter during render, which throws outside a Next app.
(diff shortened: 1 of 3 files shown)


src/__tests__/dataDotMenu.js
```diff
@@ -0,0 +1,96 @@
...
+    Trans: ({ children }) => children,
+    i18n: {},
+    RequiredNamespaces: [],
+}));
+
+// The component calls useRouter during render, which throws outside a Next app.
+jest.mock("next/router", () => ({
+    useRouter: () => ({ push: jest.fn() }),
+}));
+
```
psarando, line 32: This makes me wonder if any other tests require these mocks as well 🤔

psarando (approved):
LGTM 👍 🎉

src/__tests__/dataDotMenu.js
psarando: A quick `git grep` didn't turn up any another components that use `next/router` in any tests (assuming I did the search right).

PR #719, opened 2026-07-29
Point user-portal links at the configured portal URL

johnworth:
The logged-out banner kept the signup link in a useRef seeded with constants.DEFAULT_USER_PORTAL_URL, updated it from a useEffect, and then read the ref during render. Mutating a ref triggers no re-render, so the link stayed on https://user.cyverse.org until some unrelated re-render happened to occur -- and every refresh reverted it, since the SSR'd HTML always carried the default.

Read publicRuntimeConfig.USER_PORTAL_URL directly instead of going through the config context: the context is assembled in an effect in _app, so it is still null when the banner first renders and would leave the initial paint pointing at the default either way.

The register link on the sign-in error page and the user-portal link in the admin subscription drawer had the same problem in a simpler form -- they read the hardcoded constants and never consulted the config at all. That leaves constants.USER_PORTAL unused, so drop it.

### DIFF
--- a/src/components/error/ErrorHandler.js
+++ b/src/components/error/ErrorHandler.js
@@ -6,6 +6,7 @@
 import React, { useEffect } from "react";
 import { useRouter } from "next/router";
 import { useTranslation } from "i18n";
+import getConfig from "next/config";
 
 import NavigationConstants from "../../common/NavigationConstants";
 import GridLabelValue from "../utils/GridLabelValue";
@@ -59,6 +60,11 @@ function ErrorHandler(props) {
     const router = useRouter();
     const { classes } = useStyles();
     const errBaseId = buildID(baseId, ids.ERROR_HANDLER);
+
+    const { publicRuntimeConfig = {} } = getConfig() || {};
+    const registerURL = `${
+        publicRuntimeConfig.USER_PORTAL_URL || constants.DEFAULT_USER_PORTAL_URL
+    }/register`;
 
     useEffect(() => {
         trackIntercomEvent(IntercomEvents.ENCOUNTERED_ERROR, errorObject);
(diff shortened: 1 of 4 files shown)


psarando (approved):
LGTM 👍

src/components/error/ErrorHandler.js
```diff
@@ -60,6 +61,11 @@ function ErrorHandler(props) {
     const { classes } = useStyles();
     const errBaseId = buildID(baseId, ids.ERROR_HANDLER);
 
+    const { publicRuntimeConfig = {} } = getConfig() || {};
+    const registerURL = `${
+        publicRuntimeConfig.USER_PORTAL_URL || constants.DEFAULT_USER_PORTAL_URL
+    }/register`;
```
psarando, line 67: Can this use the `useConfig` context instead of the `publicRuntimeConfig`?

src/components/subscriptions/details/Drawer.js
```diff
@@ -292,7 +293,10 @@ function SubscriptionDrawer(props) {
 function SubscriptionHeader(props) {
     const { portalId, username } = props;
     const { classes } = useStyles();
-    const baseURL = constants.DEFAULT_USER_PORTAL_URL;
+    const { publicRuntimeConfig = {} } = getConfig() || {};
+    const baseURL =
+        publicRuntimeConfig.USER_PORTAL_URL ||
+        constants.DEFAULT_USER_PORTAL_URL;
```
psarando, line 299: Can this use the `useConfig` context instead of the `publicRuntimeConfig`?

psarando (changes requested):
Wait, I'll have to change my review, since this will cause conflicts in the next.js v16 upgrade, which gets rid of the `publicRuntimeConfig`.

It can be kept in the `Banner` for this PR, and I'll do some testing in the nextjs v16 branch for another solution, if the `useConfig` context won't work there.

PR #721, opened 2026-08-03
Show a selector for initial duration at job launch

ianmcorvidae:
(defaulting to pre-existing calculation method)

depends on https://github.com/cyverse-de/timelord/pull/31 https://github.com/cyverse-de/common-swagger-api/pull/100 https://github.com/cyverse-de/apps/pull/302

This is the one of this batch I'm least sure of, but I think it should be alright. The calculation of the dropdown options is a bit funky though, and hopefully the field shows at the correct times (or not).


src/__tests__/apps/launch.js
```diff
@@ -301,3 +305,102 @@ describe("buildGpuLimitList", () => {
         expect(buildGpuLimitList(3, 2)).toEqual([3]);
     });
 });
+
+// --- buildDurationLimitList unit tests ---
+
+describe("buildDurationLimitList", () => {
+    const SECONDS_PER_HOUR = 3600;
+    const SECONDS_PER_DAY = 86400;
```
psarando, line 313: I personally prefer the less-ambiguous raw calculations over the results (`60 * 60` and `60 * 60 * 24`).
I once read an article that gave a metric of how common the constant `84600` was across github 😅 
But not necessary to update this here.

src/components/apps/launch/formatters.js
```diff
@@ -294,4 +301,73 @@ const paramConfigsReducer = (configs, group) => {
     return configs;
 };
 
-export { formatSubmission, initAppLaunchValues, initGroupValues };
+const SECONDS_PER_MINUTE = 60;
+const SECONDS_PER_HOUR = 60 * SECONDS_PER_MINUTE;
+const SECONDS_PER_DAY = 24 * SECONDS_PER_HOUR;
```
psarando, line 306: 👍

src/components/apps/launch/formatters.js
```diff
@@ -234,6 +238,9 @@ const formatSubmission = (
         app_id,
         app_version_id,
         mount_data_store,
+        ...(initialTimeLimitSeconds && {
+            time_limit_seconds: initialTimeLimitSeconds,
+        }),
```
psarando: Why not name the form field `time_limit_seconds` instead of `initialTimeLimitSeconds`? Then it can just be `time_limit_seconds` here and in `initAppLaunchValues`?
Unless the endpoint won't accept an empty string for  `time_limit_seconds`?

src/__tests__/apps/launch.js
```diff
@@ -301,3 +305,102 @@ describe("buildGpuLimitList", () => {
...
+    const H = (n) => n * SECONDS_PER_HOUR;
+    const D = (n) => n * SECONDS_PER_DAY;
+
+    test("30 days max returns the full ladder without duplicates", () => {
```
psarando: Should this test be updated to `D(365)`, since the full ladder goes that high?

src/__tests__/apps/launch.js
```diff
@@ -301,3 +305,102 @@ describe("buildGpuLimitList", () => {
...
+            D(3),
+            D(4),
+            D(7),
+            D(14),
+            D(30),
+            D(45),
+        ]);
+    });
+
+    test("max that matches a ladder value is not duplicated", () => {
```
psarando: Seems like this is already covered by the test `7 days max` above?

src/components/apps/launch/formatters.js
```diff
@@ -294,4 +301,73 @@ const paramConfigsReducer = (configs, group) => {
...
+    if (!seconds) {
+        return "";
+    }
+    const days = Math.floor(seconds / SECONDS_PER_DAY);
+    seconds %= SECONDS_PER_DAY;
+    const hours = Math.floor(seconds / SECONDS_PER_HOUR);
+    seconds %= SECONDS_PER_HOUR;
+    const minutes = Math.floor(seconds / SECONDS_PER_MINUTE);
+    return formatDurationStr({ days, hours, minutes });
+};
```
psarando, line 365: Nice 👍 
Eventually, if anything else needs to use this function, it could move into `components/utils/DateFormatter`, but that's not necessary now.

src/components/apps/launch/AnalysisInfoForm.js
```diff
@@ -17,9 +18,64 @@ import InputSelector from "./InputSelector";
...
+                    {formatDuration(value)}
+                </MenuItem>
+            ))}
+        </FastField>
+    );
+}
+
+const AnalysisInfoForm = ({
+    formId,
+    appType,
```
psarando: Doesn't look like this `appType` prop is used anymore.

psarando (approved):
LGTM 👍 
I just had some questions and minor comments.

src/components/apps/launch/formatters.js
ianmcorvidae: Hm, that's a good question. We're wanting to deliberately send a null when one isn't selected, but it's not like a blank string is a valid number of seconds either, so it probably should be accepted as a null-y value too. I'll check on that.

ianmcorvidae: The model pointed out that `time_limit_seconds` already flows into the launch interface but with a different meaning, which is either the calculated default time limit for initial launches or the previous job's initial time limit on relaunches. So I think the destructuring to the new name is probably a little clearer upon another look.

psarando: I'm not sure what this means.

`initialTimeLimitSeconds` is populated by `time_limit_seconds` (sure it can come into the form from a default or a relaunch value), and then when the form is submitted, the submission's `time_limit_seconds` is then populated by the form's `initialTimeLimitSeconds`. It seems clearer to me to have the form just use the `time_limit_seconds` name (same as `mount_data_store` and other fields), especially since it can be changed by the user, then it's no longer the `initial` value.

src/components/apps/launch/AnalysisInfoForm.js
```diff
@@ -17,9 +18,59 @@ import InputSelector from "./InputSelector";
...
+import { MenuItem } from "@mui/material";
+
+/**
+ * Dropdown for selecting the initial duration of a VICE analysis.
+ * Stores the selected value (in seconds, or "" for "use default") as
+ * `initialTimeLimitSeconds` in Formik state.
+ */
+function InitialDurationField({ baseId, maxTimeLimitSeconds }) {
+    const { t } = useTranslation("launch");
+    const { values } = useFormikContext();
```
psarando, line 34: I keep forgetting about `useFormikContext` 👍

src/components/apps/launch/formatters.js
ianmcorvidae: Yeah, I think I might be splitting hairs unusefully. Or if I'm not, more things should be renamed than just this one form field. The use of the word `initial` at all is sort of generally confusing (because what it's meaning is initial relative to the job launch -- i.e., the value to use when timelord hasn't set any expiration date yet, as opposed to the real/eventual time limit which is that initial value plus whatever extensions the user requests, and doesn't refer to initial vs. final state of the form at all (but sort of sounds here like it does)), and it's not being used consistently, since I suppose if we wanted to enforce the word 'initial' it should be in the key used in the submission too.

I think you're right and I'll amend that a little later. Pondering if I should also update the database migration, but I guess maybe having "initial" makes more sense there since it's a place for persistent job information and not a transient submission (basically heading off the implicit question of "why is end_date - start_date not equal to the time_limit_seconds").

Sorry about this. LLM-assisted self-confusion on my part :grimacing:

psarando: No problem, and yeah, I was confusing `initial` as being specific to the form value.

I still find it easier to follow the code if the form fields match the endpoint fields. So even if the db and endpoint fields were renamed to `initial_time_limit_seconds`, the form field could match, and the formatted submission should still be able to guard against empty strings and 0 like this:
```suggestion
        ...(initial_time_limit_seconds && {
            initial_time_limit_seconds,
        }),
```

ianmcorvidae: I decided to just go with `time_limit_seconds` as the standard field name between form and endpoints, keeping `initial` in the DB where it's not a transient value so the prefix aids comprehension instead of hindering it. As you suggested here I kept the guard against empty strings and 0 though, since it was a bit cleaner to do it here and let the backend keep having nice clean integers.

